PR #2, opened 2019-07-23
Add os-autoinst-obs-auto-submit from original OBS project devel:openQA:tested/update

okurz:
Use an easier to handle VCS :)


okurz:
Sorry, the point was not to review the source itself because the script was already existing. The idea was only to move it to another repo.

PR #14, opened 2019-12-06
openqa-label-known-issues: Use "known issues" from progress.o.o for labelling as well

okurz:
This adds parsing of known issues from progress.opensuse.org and using
the found issues for searching the logfile with labelling incomplete
jobs. All issues which start their subject with 'known_issue:' and a
regex search term in double quotes '"' are used to match the log error
messages against and labelled accordingly. Right now no issues labeled
in this way are retriggered.


okurz:
merged locally

PR #18, opened 2020-02-07
Ignore incomplete jobs with reason 'quit'

Martchus:
In this case the worker has been restarted within a reasonable
time (systemd had not had to kill it). This usually happens during
deployment or when the worker is restarted/stopped for other
reasons. I suppose we can ignore these incompletes.

Note that more than 50 % of the incomplete jobs on o3 within the
last 24 hours have the reason 'quit'.

### DIFF
--- a/openqa-monitor-incompletes
+++ b/openqa-monitor-incompletes
@@ -3,7 +3,7 @@ host="${host:-"openqa.opensuse.org"}"
 ssh_host="${ssh_host:-"$host"}"
 scheme="${scheme:-"https"}"
 failed_since="${failed_since:-"(NOW() - interval '24 hour')"}"
-query="${query:-"select id,test from jobs where (result='incomplete' and t_finished >= $failed_since and id not in (select job_id from comments where job_id is not null));"}"
+query="${query:-"select id,test from jobs where (result='incomplete' and reason!='quit' and t_finished >= $failed_since and id not in (select job_id from comments where job_id is not null));"}"
 for i in $(ssh "$ssh_host" "cd /tmp; sudo -u geekotest psql --no-align --tuples-only --command=\"$query\" openqa"); do
     url="$scheme://$host/tests/${i%|*}"
     details="${i#*|}"


okurz:
ok, before we ignore let's make sure we are on the same page, just to verify:

* are jobs incompleting with result quit automatically retriggered?

Martchus:
I've just tried and yes, jobs are duplicated. However, the implementation is a little bit problematic:

```
[debug] [pid:28438] REST-API call: POST http://localhost:9526/api/v1/jobs/1219/status
^C[info] [pid:28438] Quit due to signal INT
[debug] [pid:28438] Stopping job 1219 from http://localhost:9526: 00001219-opensuse-Tumbleweed-DVD-x86_64-Build20200203-create_hdd_textmode@64bit - reason: quit
[debug] [pid:28438] REST-API call: POST http://localhost:9526/api/v1/jobs/1219/status
[info] [pid:28438] Trying to stop job gracefully by announcing it to command server via http://localhost:20013/qFN_BFK7eJCmMjiT/broadcast
[info] [pid:28438] Isotovideo exit status: 0
[info] [pid:28438] +++ worker notes +++
[info] [pid:28438] End time: 2020-02-07 12:54:19
[info] [pid:28438] Result: quit
[debug] [pid:28438] Job 1219 stopped as incomplete
[debug] [pid:28438] duplicating job 1219
[debug] [pid:28438] REST-API call: POST http://localhost:9526/api/v1/jobs/1219/duplicate?dup_type_auto=1
[warn] [pid:28438] Can't open /hdd/openqa-devel/openqa/pool/1/testresults/result-welcome.json for result upload - likely isotovideo could not be started or failed early. Error message: No such file or directory
[debug] [pid:28438] REST-API call: POST http://localhost:9526/api/v1/jobs/1219/status
[debug] [pid:28634] Uploading artefact isosize-1.txt
[debug] [pid:28438] REST-API call: POST http://localhost:9526/api/v1/jobs/1219/set_done?reason=quit&result=incomplete
[debug] [pid:28438] Job 1219 from http://localhost:9526 finished - reason: quit
[debug] [pid:28438] Informing http://localhost:9526 that we are going offline
```

That means the duplication is triggered from the worker side. I don't know why it has been implemented like this. I'd prefer having the job duplication triggered by the web UI when the worker marks the job as incomplete. So right now, when the duplication API call fails we can end up with jobs which the reason 'quit' and no duplicate. I've simulated it here by changing `"jobs/$job_id/duplicate"` to `"jobs/$job_id/duplicat"` in the worker code. If I'd changed the worker to make an API error out of it we'd be on the safe side without changing too much.

okurz:
ok, maybe that approach is better, I don't know. However I think ignoring the "quit" here is the right call.

PR #19, opened 2020-02-21
Not all pkgs have -test.changes so deal with it

asdil12:
Fixes #6

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -24,7 +24,7 @@ update_package() {
     sed -i -e 's,mode="buildtime" mode="disabled",mode="buildtime",' _service
     # ignore those updates
     rm -f _service:*-test.changes
-    cp .osc/*-test.changes .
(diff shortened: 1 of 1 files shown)


os-autoinst-obs-auto-submit
```diff
@@ -24,7 +24,9 @@ update_package() {
     sed -i -e 's,mode="buildtime" mode="disabled",mode="buildtime",' _service
     # ignore those updates
     rm -f _service:*-test.changes
-    cp .osc/*-test.changes .
+    if test -f .osc/*-test.changes; then
```
okurz: I don't think this can work as expected. Trying out locally `test -f *` or anything that returns multiple files fails

asdil12: fixed.

okurz:
thx. Let's see if this works :)

asdil12:
Looks like as if it got fixed: https://build.opensuse.org/request/show/778077

PR #32, opened 2020-08-04
Exclude incompletes with reason 'tests died'

Martchus:
These incompletes are produced when loading/compiling test modules did not work and that is most likely a test distribution issue.


okurz:
We will soon have to find a better way to distinguish by reason

PR #38, opened 2020-10-16
Print summary of jobs to be reviewed at the end

Martchus:


openqa-label-known-issues
```diff
@@ -156,11 +157,24 @@ investigate_issue() {
...
-        handle_unknown "$i" "$out"
+        handle_unknown "$i" "$out" "$reason"
     fi
     rm "$out"
 }
 
+print_summary() {
+    local to_review_count=${#to_review[@]}
+    if [[ $to_review_count -eq 0 ]]; then
+        echoerr -e "\n\e[1m\e[32mNo unknown incompletes to be reviewed.\e[0m"
```
okurz, line 168: I would prefer to use the common *nix style to not output anything if there was no problem.

okurz (commented):
Could you provide an example output in the git commit message please?

openqa-label-known-issues
Martchus: I prefer having an output similar to what most test frameworks provide (a green line which states everything's fine).

kalikiana: How about we use something well-defined like TAP?

Martchus: I prefer to have *something* better as soon as possible and would postpone and extra wishes for now. Also note that this is a shell script so we can't directly use a fancy library to do all kinds of formatting.

kalikiana: TAP is a specification, not a library. My point was that neither "nix style" nor "most test frameworks" is very meaningful and we are used to TAP and know what format to use as we extend it.

```
ok 1 - No unknown incompletes to be reviewed
```

```
not ok 1 - unknown incompletes to be reviewed
#  Job 12345
# Looks like you failed 1 test of 1
```

kalikiana:
btw this is related to [poo#73366](https://progress.opensuse.org/issues/73366)

openqa-label-known-issues
okurz: how about adding an output only on request, e.g. based on command line parameter. Think about the "happy path" which is that either no incompletes are there or all tickets could be automatically labeled. In this case I like that there is no output so I understand there is no action to be done. An the script is designed so that we could rerun it, see there is (still) no output and we know there is not more work needed. If I would need to read the text which says "No action needs to be done" then this would be more distracting.

okurz (changes requested):
We talked about this and it seems you only looked at output when calling the script with `bash -x` which I consider something quite different.

The overall design should be as in other cases of *nix to not output anything.

okurz (approved):

PR #40, opened 2020-10-21
Continue investigation if last lines before shutdown are unknown

Martchus:


okurz:
* please avoid contractions in git commit messages
* I don't understand why you would want that. Do we still ensure to exit with a non-zero exit code?

Martchus:
> please avoid contractions in git commit messages

I don't see any contradictions.

> I don't understand why you would want that.

I want the script to consider all incomplete jobs and not stop on a job where we can not print the addtional shutdown output. I also want to see the summary in that case. As far as I understand, printing the log lines near the shutdown is a purely optional feature of the review script to help the user but shouldn't change the overall behavior.

> Do we still ensure to exit with a non-zero exit code?

So far the script exits with a zero exit code if there are no jobs to review but *also* if there are unreviewed jobs. So unless an unexpected error occurs it basically always returns zero. I even tested this yesterday before and after my summary change and it didn't make a difference. Only if the script is interrupted in the middle due to an error it would return a non-zero exit code.

okurz:
> > please avoid contractions in git commit messages
> 
> I don't see any contradictions.

"contradictions" != "contractions" :)

openqa-label-known-issues
```diff
@@ -121,7 +121,7 @@ handle_unknown() {
     echoerr "$i : Unknown issue, to be reviewed -> $i/file/autoinst-log.txt"
     echoerr -e "Likely the error is within this log excerpt, last lines before shutdown:\n---"
     # Look for different termination points with likely context
-    (grep -A 12 'Backend process died, backend errors are reported below in the following lines' "$out" || grep -B 10 'sending magic and exit' "$out" || grep -B 5 'killing command server.*because test execution ended through exception' "$out" || grep -B 5 'EXIT 1' "$out" || grep -B 10 '\(Result: died\|isotovideo failed\)' "$out") | head -n -1 >&2
+    (grep -A 12 'Backend process died, backend errors are reported below in the following lines' "$out" || grep -B 10 'sending magic and exit' "$out" || grep -B 5 'killing command server.*because test execution ended through exception' "$out" || grep -B 5 'EXIT 1' "$out" || grep -B 10 '\(Result: died\|isotovideo failed\)' "$out" || echo '') | head -n -1 >&2
```
okurz, line 124: Better call `echoerr '(No log excerpt found)'`

Martchus: `echoerr` doesn't make sense because the output would not be considered by head and the `>&2` causes everything to be written to `stderr` anyways.

okurz: you are right, resolved

PR #41, opened 2020-10-29
Prevent submission of failed packages from devel:openQA:tested to Factory

Martchus:
See https://progress.opensuse.org/issues/68260

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -99,6 +99,8 @@ generate_dep_changelog() {
 osc="${osc:-"$prefix osc"}"
 
 TMPDIR=${TMPDIR:-$(mktemp -d)}
+trap 'rm -rf "$TMPDIR"' EXIT
+
 (cd "$TMPDIR"
 for package in os-autoinst openQA os-autoinst-distri-opensuse-deps; do
     $osc service wait "$src_project"/$package
@@ -129,5 +131,3 @@ for package in os-autoinst openQA os-autoinst-distri-opensuse-deps; do
     )
 done
 )
-
-rm -rf "$TMPDIR"
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -56,7 +56,7 @@ last_revision() {
     package=$2
     file=$package.changes
     service=obs_scm
-    if test "$project" != "openSUSE:Factory"; then
+    if [[ $project != $submit_target ]]; then
        file=_service:$service:$file
     fi
     local line
@@ -68,13 +68,13 @@ last_revision() {
 sync_changesrevision() {
     dir=$1
     package=$2
-    factory_rev=$(last_revision openSUSE:Factory "$package")
+    factory_rev=$(last_revision "$submit_target" "$package")
     path="$dir/$package"
     $prefix xmlstarlet ed -L -u "//param[@name='changesrevision']" -v "$factory_rev" "$path"/_servicedata
     if ! diff -u "$path"/.osc/_servicedata "$path"/_servicedata; then
        $osc up "$path"
-       $osc cat openSUSE:Factory "$package" "$package".changes > "$path/$package".changes
-       $osc ci -m 'sync changesrevision with openSUSE:Factory' "$path"
+       $osc cat "$submit_target" "$package" "$package".changes > "$path/$package".changes
+       $osc ci -m "sync changesrevision with $submit_target" "$path"
        $osc up -S "$path"
     fi
 }
@@ -82,7 +82,7 @@ sync_changesrevision() {
 generate_dep_changelog() {
     dir=$1
     package=$2
-    $osc cat openSUSE:Factory "$package" "$package".changes > "$package"-factory.changes
+    $osc cat "$submit_target" "$package" "$package".changes > "$package"-factory.changes
     {
         echo "-------------------------------------------------------------------"
         echo "$(LANG=c date -u) - Dominik Heidler <dheidler@suse.de>"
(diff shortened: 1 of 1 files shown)


os-autoinst-obs-auto-submit
```diff
@@ -128,6 +136,10 @@ for package in os-autoinst openQA os-autoinst-distri-opensuse-deps; do
         update_package $package
     )
 done
+if [[ $failed_packages ]]; then
+    echo "failed packages:$failed_packages"
+    false
+fi
 )
 
 rm -rf "$TMPDIR"
```
okurz, line 145: To do what I suggested above I guess this should be put into a signal handler to be called on "EXIT" so that it's also done in case of error

os-autoinst-obs-auto-submit
```diff
@@ -99,9 +99,17 @@ generate_dep_changelog() {
 osc="${osc:-"$prefix osc"}"
 
 TMPDIR=${TMPDIR:-$(mktemp -d)}
+failed_packages=
 (cd "$TMPDIR"
 for package in os-autoinst openQA os-autoinst-distri-opensuse-deps; do
+    echo "Considering package $package"
     $osc service wait "$src_project"/$package
+    failed_results=$($osc results --repo openSUSE_Factory --failed "$src_project" "$package")
```
okurz, line 107: I would start with something simpler to abort the complete run if *any* package within that project fails.

okurz (commented):
To answer your question from https://progress.opensuse.org/issues/68260#note-10 I think your approach is not necessarily wrong. We can consider two points for checking build results, in the source project as well as in the destination project. To be save why not do both :)

os-autoinst-obs-auto-submit
Martchus: I think this way is more consistent with the skipping logic which is already implemented. And in most cases it also make sense to still update openQA even if os-autoinst can not be built (and vice versa).

os-autoinst-obs-auto-submit
Martchus: Yes, I'll do that as an individual commit.

Martchus (commented):
Ok, then I'll keep this change (rephrasing the commit message) and also add a check for the final submission as a separate commit.

Martchus:
By the way, is `shellcheck` right here? I've tested the condition in my local shell and it worked (and shouldn't have worked if `\s` was interpreted as just `s`):

```
 In os-autoinst-obs-auto-submit line 108:
    if ! [[ $failed_results =~ $\s*^ ]]; then
                                ^-- SC1001: This \s will be a regular 's' in this context.

Makefile:16: recipe for target 'test-shellcheck' failed
```

Martchus:
I've pushed a new version but I suppose the waiting checks are a total race condition and I also don't know how to actually test it except for playing around with it in the interactive shell locally.

I moved the other change to a [different branch](https://github.com/Martchus/scripts/tree/prevent-submission-of-failed-packages-2) for now to do only one change at a time. I'll create a PR for it after this is done.

os-autoinst-obs-auto-submit
okurz: Yes, most cases. But I would go with "worst case" as design guideline

okurz:
> I also don't know how to actually test it except for playing around with it in the interactive shell locally.

You can experiment with a different target project for testing, e.g. submit to home:mkittler:test_poo68260

Martchus:
The submit target was actually hardcoded in several places preventing me from testing it like that. I've fixed that now. I've also created a custom source and ":tested" project (to be able to actually produce a new revision - otherwise it just says there's no new revision). But I'm still running into seemingly unrelated errors when testing it locally.

os-autoinst-obs-auto-submit
```diff
@@ -42,6 +44,27 @@ update_package() {
...
+        echo "Waiting for re-build of $package to be considered ($attempts)"
+        sleep "$osc_poll_interval"
+        attempts=$((attempts - 1))
+    done
+    while curl -s "$osc_query" | grep -e "$wip_states" ; do
+        echo "Waiting while $package is scheduled/built/published"
+        sleep "$osc_poll_interval"
+    done
+    if curl -s "$osc_query" | grep -e "$bad_states" ; then
+        echo "Built of $package has failed, skipping submission"
```
okurz, line 65: s/Built/Build/ or "Building package '$package' has failed, skipping submission"

os-autoinst-obs-auto-submit
```diff
@@ -42,6 +44,27 @@ update_package() {
...
+    local osc_query="https://api.opensuse.org/public/build/$dst_project/_result?repository=$submit_target&package=$package"
+    local wip_states='\(unknown\|blocked\|scheduled\|dispatching\|building\|signing\|finished\)'
+    local bad_states='\(failed\|unresolvable\|broken\)'
+    local attempts="$osc_build_start_poll_tries"
+    while ! curl -s  "$osc_query" | grep -e "$wip_states" ; do
+        if [[ $attempts -le 0 ]]; then
+            echo "warning: Re-build of $package has not been considered in time (or package has been re-built so fast that we've missed it)"
+            break
+        fi
+        echo "Waiting for re-build of $package to be considered ($attempts)"
```
okurz, line 56: the attempts within parentheses will just appear as a bare number. I would make that explicit and write something like "attempts left: $attempts"

os-autoinst-obs-auto-submit
```diff
@@ -42,6 +44,27 @@ update_package() {
...
+        sleep "$osc_poll_interval"
+        attempts=$((attempts - 1))
+    done
+    while curl -s "$osc_query" | grep -e "$wip_states" ; do
+        echo "Waiting while $package is scheduled/built/published"
```
okurz, line 61: Instead of `scheduled/built/published` I would use something like just "in-progress"

okurz:
still want to have it marked as "WIP"?

okurz (approved):

Martchus:
As mentioned, I couldn't really test it. But I also don't know what else to try so I remove the "WIP: " prefix.

PR #48, opened 2020-11-26
Fix investigation script

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/80418

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -33,7 +33,7 @@ trigger_jobs() {
     clone "$id" 'retry' "${@:2}"
 
     job_url="$host_url/tests/$id"
-    investigation=$(curl -s "$job_url"/investigation)
+    investigation=$(curl -s "$job_url"/investigation_ajax)
     last_good=$(echo "$investigation" | jq -r '.last_good')
 
     # for 2. current job/build + last good test (+ last good needles) ->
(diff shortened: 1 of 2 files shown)


openqa-investigate
```diff
@@ -33,8 +34,8 @@ trigger_jobs() {
     clone "$id" 'retry' "${@:2}"
 
     job_url="$host_url/tests/$id"
-    investigation=$(curl -s "$job_url"/investigation)
-    last_good=$(echo "$investigation" | jq -r '.last_good')
+    investigation=$(curl -s "$job_url"/investigation_ajax)
```
okurz, line 37: if we use that route from external then the name "investigation_ajax" does not sound that good. I think all API routes that are accessible from the outside should have a name that make sense when someone uses them because someone *will* use them, right? I think we can come up with something better in openQA and then we can change it again here.

okurz (approved):

openqa-investigate
Martchus: All routes the web application provide are accessible from the outside. This particular route is not part of the "API" and the structure of the JSON is determined by the way we display the data (e.g. the ID of the last good job is returned as `.last_good.text` and not `.last_good.id). We could create a decently designed API at some point.

okurz: sorry, my mention of "API" was ambiguous. I meant all web routes. So certainly we can define a "proper API endpoint" but my point is: Every web route should be designed assuming that people use it because if it's accessible then they will :)

PR #44, opened 2020-11-20
Add backlog-set-due-date script

ilausuch:
This script sets the duedate to 14 days from the script's
execution time for tickets that are not low priority and
have not been assigned to anyone

Add dry_run mode. It uses a test json file as first parameter
to perform the tests and thus have predictable results

https://progress.opensuse.org/issues/73468

PR #51, opened 2020-11-30
Enable backlog-set-due-date script

ilausuch:
In order to enable an action to the
backlog-set-due-date script introduced
in the PR #44 this PR modify the current
workflow for WIP-limits to include also
this script

progress.opensuse.org/issues/73468

### DIFF
--- a/.github/workflows/backlog.yml
+++ b/.github/workflows/backlog.yml
@@ -1,5 +1,5 @@
 ---
-name: Check SUSE QA Tools WIP-Limit
+name: Check SUSE QA Tools WIP-Limit and set due dates
 # yamllint disable-line rule:truthy
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


ilausuch:
4:00 PM runs at the same time than the other script for WIP-limits, should it be interesting run this at 4:30 PM? what do you think? In my opinion could be concurrent.

kalikiana:
> It runs at 4:00 PM that is the same time than the other script for WIP-limits, should it be interesting run this at 4:30 PM? what do you think? In my opinion could be concurrent.

I think it doesn't matter. It runs once a day. We probably won't hit API limits here.

kalikiana (approved):

.github/workflows/backlog_duedate.yml
```diff
@@ -0,0 +1,17 @@
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set SUSE QA Tools due dates
+        env:
+          redmine_api_key: ${{ secrets.REDMINE_API_KEY }}
+        run: |
+         sudo apt-get install curl jq
+         sh -ex backlog-set-due-date
```
okurz, line 17: just add https://github.com/os-autoinst/scripts/blob/master/.github/workflows/backlog.yml#L17

ilausuch: Done

okurz (approved):

PR #59, opened 2020-12-11
README: Explain "auto-review" further directly

okurz:
After I announcement "auto-review" to a broader audience pointing to
this README I would like to improve the explanation how auto-review
works and how it can be used including caveats to keep in mind.

Related progress issue: https://progress.opensuse.org/issues/80828


okurz:
merging right now so that people reading the announcement find the updated text. However please feel welcome to provide suggestions based on the diff and I will happily incorporate them in a followup.

PR #60, opened 2020-12-11
openqa-investigate: Allow to filter out by name, job group, parent group, no group

okurz:
Tested manually locally with current jobs that I could find on o3:

* 1502469 in "Development / Development Tumbleweed"
  -> "Job group 'Development Tumbleweed / Development' matches
  $exclude_group_regex 'Development.*/ ', skipping investigation"
* 1502404 outside any job group
  -> "Job w/o job group, $exclude_no_group is set, skipping
  investigation"

Related progress issue: https://progress.opensuse.org/issues/80830

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -8,6 +8,10 @@ scheme="${scheme:-"https"}"
 host_url="$scheme://$host"
 dry_run="${dry_run:-"0"}"
 prio_add="${prio_add:-"100"}"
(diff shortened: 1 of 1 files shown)


Martchus (commented):
I assume you've tested this. (Due to the lack of real tests here it seems I'll always have to ask that question for changes which are not completely trivial.)

okurz:
and I appreciate that you ask. I updated the description with my test results

Martchus (approved):

PR #63, opened 2020-12-29
Correctly check if a published Tumbleweed image was found

kalikiana:
- `jq` returns *null* not empty when nothing is found
- Always use `openqa-cli` in place of `openqa-client`
- Add tests for successful (dry_run) and parse error

See: [poo#81492](https://progress.opensuse.org/issues/81492)

### DIFF
--- a/trigger-openqa_in_openqa
+++ b/trigger-openqa_in_openqa
@@ -12,7 +12,7 @@ target_host_proto="${target_host_proto:-"https"}"
 dry_run="${dry_run:-"0"}"
 tw_openqa_host="${tw_openqa_host:-"https://openqa.opensuse.org"}"
 tw_group_id="${tw_group_id:-"1"}"
-openqa_client="${openqa_client:-"/usr/share/openqa/script/client"}"
+openqa_cli="${openqa_cli:-"/usr/share/openqa/script/openqa-cli"}"
 arch="${arch:-"x86_64"}"
 machine="${machine:-"64bit"}"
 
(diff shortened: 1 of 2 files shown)


trigger-openqa_in_openqa
```diff
@@ -29,14 +29,13 @@ find_latest_published_tumbleweed_image() {
         echo "Unable to find latest published Tumbleweed build."
         exit 1
     fi
-    qcow=$(/usr/share/openqa/script/openqa-cli api --host "${tw_openqa_host}" assets get \
+    qcow=$(${openqa_cli} api --host "${tw_openqa_host}" assets get \
         | jq -r "[.assets[] | select(.name | test(\"Tumbleweed-$arch-$latest_published_tw_build-Tumbleweed\\\\@$machine.qcow\"))] | .[0] | .name")
-    if ! [[ $qcow ]]; then
+    if [[ "$qcow" = "null" ]]; then
```
Martchus, line 34: What about the case when the JSON is completely invalid? Maybe `$qcow` is still empty then. I also remember that I've tested this case and maybe in some cases `$qcow` is really just empty. I'd simply check for both (null and empty) here to be sure.

perlpunk: In case the JSON is invalid, jq and the script should abort because of `bash -e`.
But also checking for the empty string doesn't sound like a bad idea.

kalikiana: If the JSON is invalid there's no return value:

    parse error: Unfinished JSON term at EOF at line 2, column 0

I can't come up with a case where the return value is the empty string.

okurz: this change makes sense to me. In 9a1ae1d mkittler changed from "sed" to "jq" so instead of an empty string we now get the null value from jq

okurz (approved):

PR #66, opened 2021-01-12
Drop the extra "get" in find_latest_published_tumbleweed_image

kalikiana:
That argument has no meaning and is at best confusing.


Martchus (approved):
This was likely supposed to be the HTTP method but I'm aware that it should actually be passed with `-X GET`.

PR #78, opened 2021-05-31
Trigger investigation jobs against the same repo as original job

Martchus:
* Otherwise investigation jobs for jobs with custom `CASEDIR` might
  incomplete with `isotovideo died: Could not find '…' in complete history`
* See https://progress.opensuse.org/issues/92905

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -19,6 +19,7 @@ echoerr() { echo "$@" >&2; }
 clone() {
     id="${1:?"Need 'job_id'"}"
     name_suffix="${2+":$2"}"
+    refspec=${3+$3}
     job_data=$(openqa-cli api --json --host "$host_url" jobs/"$id")
     # shellcheck disable=SC2181
     [[ $? != 0 ]] && echoerr "unable to query job data for $id: $job_data" && return 1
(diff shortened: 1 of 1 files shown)


openqa-investigate
```diff
@@ -27,10 +29,19 @@ clone() {
...
+        clone_settings+=("CASEDIR=${repo%#*}#${refspec}")
+    fi
+    clone_settings+=("${@:4}")
     # clear "PUBLISH_" settings to avoid overriding production assets
     # shellcheck disable=SC2207
-    publish_keys=($(echo "$job_data" | jq -r '.job.settings | keys[] | select (startswith("PUBLISH_")) | . + "="'))
-    out=$($clone_call "$host_url/tests/$id" TEST="$name" _GROUP_ID=0 BUILD= "${publish_keys[@]}" "${@:3}")
+    clone_settings+=($(echo "$job_data" | jq -r '.job.settings | keys[] | select (startswith("PUBLISH_")) | . + "="'))
+    out=$($clone_call "$host_url/tests/$id" "${clone_settings[@]}")
+    [[ $dry_run = 1 ]] && echo "$out"
```
okurz, line 44: are you abusing `$dry_run` as a debug flag? The call further down with `$client_call` should already provide the proper "dry-run" call. What you are interested here should be visible if you call the script with `-x` already

Martchus: `-x` makes the output *very* verbose. I want to show the clone invocation like the dry-run mode would actually do unless in this case because we capture the output here. The client call below doesn't show the arguments passed to the clone script.

PR #79, opened 2021-06-09
os-autoinst-obs-auto-submit: Look for updates in all packages in target project

okurz:


os-autoinst-obs-auto-submit
```diff
@@ -126,32 +126,33 @@ TMPDIR=${TMPDIR:-$(mktemp -d)}
 trap 'rm -rf "$TMPDIR"' EXIT
 
 (cd "$TMPDIR"
-for package in os-autoinst openQA os-autoinst-distri-opensuse-deps; do
-    $osc service wait "$src_project"/$package
-    $osc co -S "$src_project"/$package
+auto_submit_packages="$(osc ls "$dst_project" | grep -v '\-test$')"
```
Martchus, line 129: Note that quotes are generally not needed on assignments, e.g. the following would also work if the command outputs spaces:
```suggestion
auto_submit_packages=$(osc ls "$dst_project" | grep -v '\-test$')
```

os-autoinst-obs-auto-submit
```diff
@@ -126,32 +126,33 @@ TMPDIR=${TMPDIR:-$(mktemp -d)}
...
+for package in $auto_submit_packages; do
+    $osc service wait "$src_project"/"$package"
+    $osc co -S "$src_project"/"$package"
     if test "$package" == "os-autoinst-distri-opensuse-deps"; then
         $osc cat "$submit_target" "$package" "$package.spec" > "$package-factory.spec"
-        if diff -u $package-factory.spec "$src_project"/$package/_service:obs_scm:$package.spec | grep "^[+-]Requires"; then
+        if diff -u "$package"-factory.spec "$src_project"/"$package"/_service:obs_scm:"$package".spec | grep "^[+-]Requires"; then
```
Martchus, line 135: It would likely look less noisy when using just one pair of quotes, e.g.
```suggestion
        if diff -u "$package"-factory.spec "$src_project/$package/_service:obs_scm:$package.spec" | grep "^[+-]Requires"; then
```

Martchus (commented):
I had just a few remarks regarding coding style but I don't care much about them. I was just wondering about your preference for so many quotes :-)

I assume you've tested this somehow. If it works I suppose the approach is fine.

os-autoinst-obs-auto-submit
okurz: I am adding quotes where shellcheck demands them not anywhere else

os-autoinst-obs-auto-submit
okurz: That we can do

os-autoinst-obs-auto-submit
okurz: ok, I was wrong. shellcheck does not demand these. I changed that

Martchus (approved):

okurz:
https://build.opensuse.org/request/show/899103 was generated automatically based on the change in here. What I did on top is create a manual initial https://build.opensuse.org/package/view_file/devel:openQA/xterm-console/_servicedata?expand=1 , bumped the version number in https://github.com/os-autoinst/xterm_console/commit/2532f1418ee6df6f091c4c0e036a0852f7f4e0b4 , created a new tag 1.1 https://github.com/os-autoinst/xterm_console/tree/1.1 , added an OBS token with `osc token --create devel:openQA xterm-console`, created a webhook in github https://github.com/os-autoinst/xterm_console/settings/hooks/301873713 with URL `https://build.opensuse.org/trigger/webhook?id=4986` and the token as secret, confirm that this triggers OBS services in https://build.opensuse.org/package/show/devel:openQA/xterm-console

PR #84, opened 2021-06-23
Handle multiple job IDs in scripts used for openQA-in-openQA test

Martchus:
See https://progress.opensuse.org/issues/88754

### DIFF
--- a/monitor-openqa_job
+++ b/monitor-openqa_job
@@ -15,28 +15,39 @@ openqa_cli="${openqa_cli:-"openqa-cli"}"
 host="${host:-"https://openqa.opensuse.org"}"
 openqa_groupid="${openqa_groupid:-"24"}"
 obs_component="${obs_component:-"package"}"
+obs_package_name="${1:-""}"
 
 # shellcheck source=/dev/null
 . "$(dirname "$0")"/_common
 
 [[ -f job_post_response ]] || (echo "Need job response status file 'job_post_response'" && exit 2)
-failures=0
+declare -A failed_versions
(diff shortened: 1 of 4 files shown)


monitor-openqa_job
```diff
@@ -14,31 +16,26 @@ host="${host:-"https://openqa.opensuse.org"}"
...
+    done
+    response=$($openqa_cli api jobs/"$job_id")
+    result=$(echo "$response" | jq -r '.job.result')
+    echo "Result of job $job_id: $result"
+    obs_package_name=$1
+    if [[ $result != 'passed' ]] && [[ -n $obs_package_name ]] ; then
+        VERSION=$(echo "$response" | jq -r '.job.settings.VERSION')
+        osc api /comments/"${obs_component}"/"$obs_package_name" | grep id= | sed -n 's/.*id="\([^"]*\)">test failed.*/\1/p' | while read -r id; do osc api -X DELETE /comment/"$id"; done
+        osc api --data="test failed, see https://openqa.opensuse.org/tests/overview?version=$VERSION&groupid=${openqa_groupid}" -X POST "/comments/${obs_component}/$obs_package_name"
+        failures=$((failures + 1))
```
okurz, line 37: So we would add a comment on OBS for each failing test but always referencing the overview page. I think we can do better :)

As you already count failures how about instead of adding up a number add all failed jobs to an array and if the array is not empty put a comment on OBS and exit with 1

okurz (changes requested):

okurz (approved):

PR #91, opened 2021-07-22
Improve error handling and print only first 15 lines

perlpunk:

We saw that we are getting a 404 Not Found response in some cases,
which returns a huge HTML page that fills our logs.

Simply get the first 15 lines for now and improve handling of
status codes later.

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -16,6 +16,7 @@ exclude_no_group="${exclude_no_group:-"true"}"
 # exclude_group_regex checks a combined string "<parent job group name> / <job group name>"
 exclude_group_regex="${exclude_group_regex:-"Development.*/ "}"
 client_args="--host $host_url"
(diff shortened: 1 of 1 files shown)


openqa-investigate
```diff
@@ -26,6 +26,7 @@ runjq() {
     rc=$?
     set -e
     (( "$rc" == 0 )) && echo "$output" && return
+    output=$(echo "$output" | head -15)
```
okurz, line 29: please replace this magic number. How about

```suggestion
jq_output_limit="${jq_output_limit:-15}"
…
    output=$(echo "$output" | head -$jq_output_limit)
```

okurz (changes requested):

openqa-investigate
perlpunk: done

Martchus (approved):

okurz (approved):

PR #92, opened 2021-07-23
openqa-investigate: Improve curl HTTP status handling

perlpunk:

If a HTTP request doesn't return 200, we do not want to continue, and we want
to know the actual URL and status.

Because `investigate` is called with `|| do-something`, -e is not
in effect, so we check each `runcurl` call explicitly.

I did the same for the `runjq` calls.

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -10,6 +10,7 @@ host="${host:-"openqa.opensuse.org"}"
 scheme="${scheme:-"https"}"
 host_url="$scheme://$host"
 dry_run="${dry_run:-"0"}"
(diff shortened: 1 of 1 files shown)


okurz (approved):

foursixnine:
@perlpunk just a quick niptick: `investigate` instead of `ivestigate` :)

perlpunk:
> @perlpunk just a quick niptick: `investigate` instead of `ivestigate` in the commit message (and title) :)

oops. fixed in title and commit :)

Martchus (approved):

PR #94, opened 2021-07-27
openqa-investigate: Improve handling of unexpected values

perlpunk:

Apparently .last_good.text is null because this url was fetched:

    https://openqa.suse.de/tests/null/file/vars.json

Since we expect a test id, we now explicitly test for digits only and
abort on anything else

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -40,9 +40,9 @@ runcurl() {
     response=$(curl -w "\n%{http_code}\n" "$@" 2>&1)
     rc=$?
     set -e
-    (( "$rc" != 0 )) && echo "curl: Error fetching ($*): $response" >&2 && return 1
+    (( "$rc" != 0 )) && echo "curl ($(caller)): Error fetching ($*): $response" >&2 && return 1
     status_code=$(echo "$response" | tail -1)
-    (( "$status_code" != 200 )) && echo "curl: Error fetching url ($*): Got Status $status_code" >&2 && return 1
+    (( "$status_code" != 200 )) && echo "curl ($(caller)): Error fetching url ($*): Got Status $status_code" >&2 && return 1
     # remove last line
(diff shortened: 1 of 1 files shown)


okurz (commented):
can you please state how the generated comments on openQA side differ before and after in such cases?

perlpunk:
@okurz I didn't understand what you asked, but I changed the `echo` to use stderr now and assume that's what you meant.
Since the line I added is clearly unexpected, it is treated as an error.

okurz:
> @okurz I didn't understand what you asked, but I changed the `echo` to use stderr now and assume that's what you meant.
> Since the line I added is clearly unexpected, it is treated as an error.

Well, I don't know if previously when encountering the error we excited early or ignored the problem. openqa-investigate writes a comment on openQA so this is where with your change the behavior might differ. I suggest you find a job with missing last good and call openqa-investigate on this job before and after and see how it differs, not only on stderr/stdout but also the content of the generated openQA comment if it would be generated at all

perlpunk:
For this command:
`echo 1851745 | host=openqa.opensuse.org ./openqa-investigate`

The comment in current master would look like:
```
Automatic investigation jobs:

* **container-host:investigate:retry**:
```
The comment after applying this PR would look like:
```
Automatic investigation jobs:

* **container-host:investigate:retry**:
```
Which is the same.

(I pretended an error case by modifying a variable in the code)

okurz:
but https://openqa.opensuse.org/tests/1851745# is no failed job so there is no investigation info with no last good. We should look for a failed job with no "last good" in history, right?

okurz:
ok, let's assume this is good :)

perlpunk:
I'm still trying to find out the actual underlying problem with this (why `.last_good` is there, not equals "not found", but `.last_good.text` is null).
This is not the final PR, it will just bring us closer to the problem.

PR #101, opened 2021-08-16
Simplify exit code evaluation

okurz:
No need for arithmetic calculations here.


Martchus (commented):
Spwning a sub-shell doesn't seem to be much better. Since we're using bash anyways, what about `[[`?

okurz:
why do you think we are spawning a subshell, it's not `(` but `[`. That's equivalent to the `test` command

Martchus:
Sorry, I didn't mean sub-shell, I mean sub-process.

okurz:
Updated to prefer `[[` to prevent unnecessary sub-processes

Martchus (approved):

_common
```diff
@@ -36,9 +36,9 @@ runcurl() {
     response=$(curl -w "\n%{http_code}\n" "$@" 2>&1)
     rc=$?
     set -e
-    (( "$rc" != 0 )) && echo "curl ($(caller)): Error fetching ($*): $response" >&2 && return 1
+    [[ "$rc" != 0 ]] && echo "curl ($(caller)): Error fetching ($*): $response" >&2 && return 1
```
perlpunk, line 39: When using `[[ ]]`, then using `-ne` would be more correct than `!=`, right?

PR #104, opened 2021-08-27
Add a link to the original job in the investigation ones

ilausuch:
Add a link pointing the original job from every investigation job
created from this. And also a link to the job from the investigation
job is cloned from

Example: https://openqa.opensuse.org/tests/1886978#comments

https://progress.opensuse.org/issues/95746

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -58,6 +58,7 @@ clone() {
     echo "* **$name**: $url"
     clone_id=${out/:*/}; clone_id=${clone_id/*#/}
     $client_call --json --data "{\"priority\": $((base_prio + prio_add))}" -X PUT jobs/"$clone_id" >/dev/null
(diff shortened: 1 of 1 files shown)


ilausuch:
> Looking at the ticket I'm not sure this is the purpose of the ticket.

This PR appears as a result of researching how these scripts work, and following the suggestion "ones and provide a comment back on the original job"

However feel free to use this ticket, or close it if you consider that is out of scope

okurz:
> This PR appears as a result of researching how these scripts work, and following the suggestion "ones and provide a comment back on the original job"

ah, yes. But your PR seems to provide the backward which we already have: "Provide a comment on the investigation job pointing to the original job …" instead of "comment on the original job pointing to the investigation job about the result of retry"

kalikiana:
I'll assume this was obsoleted by #109 as per [poo#95746](https://progress.opensuse.org/issues/95746).

ilausuch:
Thanks for closing

PR #105, opened 2021-09-08
Add User-Agent header to openqa-cli calls

perlpunk:
This way we have more information in our logs (if the scripts are making
any requests, or how many)

Related issue: https://progress.opensuse.org/issues/96959

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -31,7 +31,7 @@ clone() {
     id=${2:?"Need 'id'"}
     name_suffix=${3+":$3"}
     refspec=${4+$4}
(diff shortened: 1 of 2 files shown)


Martchus (approved):

kalikiana (approved):

perlpunk:
This now leads to `GET /api/v1/openqa-label-known-issues HTTP/1.1" 404`
because of the space in the header argument. Will try to come up with a fix or revert

Martchus:
Maybe we should make it a real array?

perlpunk:
@Martchus yep, see #106

PR #111, opened 2021-09-18
Restart the job which fails in 'No space left on device'

Amrysliu:
Related: https://progress.opensuse.org/issues/98682


okurz (changes requested):
Is there a specific reason you want to extend the "issues without ticket"? I suggest to use standard "auto-review" which works based on existing tickets by just including the corresponding regex in the subject line of a ticket. However "No space left on device" is very generic and in general I would not retrigger such jobs. Only in some special cases like no space on a worker pool dir we know that retriggering can help but the label should only be applied in such cases in a very temporary matter.

Amrysliu:
As discussed in weekly unblock, we re-trigger the job by adjusting ticket's subject.

PR #113, opened 2021-09-27
openqa-monitor-investigation-candidates: Fix combination of queries with 'union'

okurz:
The command argument of psql only outputs the evaluation of the last
command so the first part was never properly processed.


Martchus (approved):
I cannot tell whether this SQL syntax is correct out of my head but I assume you've tested it.

okurz:
> I cannot tell whether this SQL syntax is correct out of my head but I assume you've tested it.

I tested the SQL queries manually on OSD and also verified that the number of jobs returned is same for before and after

PR #116, opened 2021-10-05
Add dependencies.yaml and update-deps target

kalikiana:


dist/rpm/scripts.spec
```diff
@@ -0,0 +1,53 @@
+#
+# spec file for package os-autoinst
```
Martchus, line 2:
```suggestion
# spec file for package os-autoinst/scripts
```

dist/rpm/scripts.spec
```diff
@@ -0,0 +1,53 @@
...
+# The following line is generated from dependencies.yaml
+%define test_version_only_requires perl(Mojo::IOLoop::ReadWriteProcess) >= 0.28
+# The following line is generated from dependencies.yaml
+%define test_requires %build_requires %spellcheck_requires %test_base_requires %yamllint_requires perl(Inline::Python) perl(YAML::PP)
+# The following line is generated from dependencies.yaml
+%define devel_requires %python_style_requires %test_requires ShellCheck perl(Devel::Cover) perl(Devel::Cover::Report::Codecov) perl(Perl::Tidy)
+BuildRequires:  %test_requires %test_version_only_requires
+Requires:       %main_requires
+# Optional dependency for Python test API support
+Recommends:     perl(Inline::Python)
```
Martchus, line 46: I doubt we need most of this here.

Martchus (changes requested):

dependencies.yaml
```diff
@@ -0,0 +1,46 @@
...
+  perl(Data::Dump):
+  perl(Data::Dumper):
+  perl(Getopt::Long):
+  perl(File::Basename):
+  perl(File::Rsync):
+  perl(File::Temp):
+  perl(JSON):
+  perl(Mojo::File):
+  perl(Mojo::URL):
+  perl(Mojo::UserAgent):
```
okurz, line 37: And where do we use all this?

mimi1vx (changes requested):
spec file:

should have same filename as name of package 
is incomplete .. missing almost all sections

PR #121, opened 2021-12-03
Add force_result feature and automatic tests

perlpunk:
Issue: https://progress.opensuse.org/issues/102374

Tested with:

* Replace the `issues=` line with:
```
    issues='102374
foo bar auto_review:"Test died: Machine didn.t shut down":retry:force_result:failed'
```
* Call
```
    echo https://openqa.opensuse.org/tests/1990037 | ./openqa-label-known-issues
```

I also added automatic test, so far for label_on_issue and functions below.

I added [Test::More bash](https://github.com/ingydotnet/test-more-bash) as a subrepo, see README

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Static checks, unit- and integration tests
+      - name: unit- and integration tests
+        run: |
+         make test-unit
+
+  style:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Static checks
         run: |
          sudo apt-get install yamllint shellcheck
-         make test
+         make checkstyle
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,11 @@ all:
 test: checkstyle test-unit
 
 .PHONY: test-unit
-test-unit:
+test-unit: test-more-bash
 	prove -r test/
+
+test-more-bash:
+	git clone https://github.com/ingydotnet/test-more-bash.git --depth 1 -b 0.0.5
 
 .PHONY: test-online
 test-online:
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,11 @@
 all:
 
 .PHONY: test
-test: checkstyle
+test: checkstyle test-unit
+
+.PHONY: test-unit
+test-unit:
+	prove -r test/
 
 .PHONY: test-online
 test-online:
--- a/README.md
+++ b/README.md
@@ -17,18 +17,32 @@ progress.opensuse.org within https://progress.opensuse.org/projects/openqav3
 or any subproject of it, commonly
 https://progress.opensuse.org/projects/openqatests/ , openQA jobs can be
 automatically labeled with the corresponding ticket and optionally retriggered
-where it makes sense. For this the subject line of a ticket must include text
-following the format `auto_review:"<search_term>"[:retry]` with
-`<search_term>` being the perl extended regex to search for and an optional
-boolean switch `:retry` after the quoted search term to instruct for
-retriggering the according openQA job. Example: `auto_review:"error 42
-found":retry`. The search terms are crosschecked against the logfiles and
+where it makes sense.
+
+For this the subject line of a ticket must include text following the format
+`auto_review:"<search_term>"[:retry][:force_result:<result>]`
+
+* `<search_term>`: the perl extended regex to search for
+* `:retry`: (optional) boolean switch after the quoted search term to instruct
+   for retriggering the according openQA job.
+* `:force_result:<result>`: (optional) give the job a special label which forces the result
+   to be the given string
+
+Examples:
+* `auto_review:"error 42 found"`.
+* `auto_review:"error 42 found":retry`.
+* `auto_review:"error 42 found":force_result:softfailed`.
+* `auto_review:"error 42 found":retry:force_result:softfailed`.
+
+The search terms are crosschecked against the logfiles and
 "reason" field of the openQA jobs. A multi-line search is possible, for
-example using the `<search_term>` `(?s)something to match earlier.*something
-to match some lines further down`. Other double quotes in the subject line
-than around the search term should be avoided. Also be careful to not specify
-a too generic search term to prevent false matches of job failures unrelated
-to the specified ticket.
+example using the `<search_term>`
+
+  `(?s)something to match earlier.*something to match some lines further down`
+
+Other double quotes in the subject line than around the search term should be
+avoided. Also be careful to not specify a too generic search term to prevent
+false matches of job failures unrelated to the specified ticket.
 
 * [openqa-monitor-incompletes](https://github.com/os-autoinst/scripts/blob/master/openqa-monitor-incompletes)
   queries the database of an openQA instance (ssh access is necessary) and
@@ -120,6 +134,22 @@ If this is too much hassle for you feel free to provide incomplete pull
 requests for consideration or create an issue with a code change proposal.
 
 ### Local testing
+
+#### Functional testing
+
+This is done with the [Test::More bash
+library](https://github.com/ingydotnet/test-more-bash). It is included
+into this repository via a
+[git-subrepo](https://github.com/ingydotnet/git-subrepo) under `external/`.
+
+
+    # only a few functions from openqa-label-known-issues so far
+    make test
+
+#### Style checks
+
+    make checkstyle
+
 #### openqa-label-known-issues
 Generate a list of recent incomplete jobs of your local openQA instance. Here's an example using `psql`:
 
--- a/README.md
+++ b/README.md
@@ -138,9 +138,8 @@ requests for consideration or create an issue with a code change proposal.
 #### Functional testing
 
 This is done with the [Test::More bash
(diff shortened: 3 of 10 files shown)


openqa-label-known-issues
```diff
@@ -104,7 +110,16 @@ label_on_issues_from_issue_tracker() {
...
     echo "$issues" | (while read -r issue; do
         read -r subject
-        after=${subject#*\"} && search=${after%\"*} && [[ ${#search} -ge $min_search_term ]] && label_on_issue "$id" "$search" "poo#$issue $subject" "${after//*\":retry*/1}" && break; done)
+        after=${subject#*\"}
+        search=${after%\"*}
+        if [[ ${#search} -ge $min_search_term ]]; then
+            if [[ $after =~ :force_result:([a-z_]+) ]]; then
+                force_result_on_issue "$id" "poo#$issue" "${BASH_REMATCH[1]}"
+            fi
+            label_on_issue "$id" "$search" "poo#$issue $subject" "${after//*\":retry*/1}" "$force_result" && break
```
okurz, line 119: Should we turn this around so that in case posting this fails at least is tried first?

okurz (changes requested):
Looks nice. Can you also add a description in the README as documentation?

openqa-label-known-issues
perlpunk: > Should we turn this around so that in case posting this fails at least is tried first?

I fail to parse this sentence...

okurz: Nevermind. Just leave it like it is for now. We will see how it goes :)

kalikiana (approved):

okurz:
Change is fine. Improve the commits and undraft for approval

perlpunk:
@okurz  It's not yet finished. In the current state it's broken.

perlpunk:
Now it's ready for review. I had to rearrange the functions a bit and put some of them into `_common` to be able to unit-test them, so unfortunately the change and the rearrangement is in one commit.
Added test-more-bash as a git-subrepo.

README.md
```diff
@@ -17,18 +17,32 @@ progress.opensuse.org within https://progress.opensuse.org/projects/openqav3
...
 automatically labeled with the corresponding ticket and optionally retriggered
-where it makes sense. For this the subject line of a ticket must include text
-following the format `auto_review:"<search_term>"[:retry]` with
-`<search_term>` being the perl extended regex to search for and an optional
-boolean switch `:retry` after the quoted search term to instruct for
-retriggering the according openQA job. Example: `auto_review:"error 42
-found":retry`. The search terms are crosschecked against the logfiles and
+where it makes sense.
+
+For this the subject line of a ticket must include text following the format
```
Martchus, line 22:
```suggestion
For this the subject line a ticket must include text following the format:
```

Martchus (commented):
Looks generally good, only found a typo.

Martchus (approved):

Makefile
```diff
@@ -1,8 +1,12 @@
 .PHONY: all
 all:
 
+.PHONE: test-all
+test-all: checkstyle test
+
 .PHONY: test
-test: checkstyle
+test:
+	prove test/*
```
okurz, line 9: `prove -r` should work

okurz (changes requested):
I consider test-more-bash rather remote to our development (unlike what we do for os-autoinst-common) so I recommend against using subrepo to include that here. I would not even use git submodules here but simply rely on test-more-bash to be provided in a separate directory. Within the CI pipeline we can just clone it and within a local environment we can leave it to users to checkout but provide convenient setup tools, e.g. compare to https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/Makefile#L11

Makefile
kalikiana: I'm guessing that's slower if it needs to search the whole tree?

Makefile
```diff
@@ -1,8 +1,12 @@
 .PHONY: all
 all:
 
+.PHONE: test-all
```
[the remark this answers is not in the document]
perlpunk: done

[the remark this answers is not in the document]
perlpunk: done

Makefile
perlpunk: I made it `prove -r test/` now.

perlpunk:
> I consider test-more-bash rather remote to our development

I don't see that as a reason against subrepo. The repo is not that big.
Opinions from anyone else?

kalikiana:
> > I consider test-more-bash rather remote to our development
> 
> I don't see that as a reason against subrepo. The repo is not that big. Opinions from anyone else?

Before that I would ask why this isn't packaged. If we mainly want to make it *available* rather than develop and it's used like a *library* I would regard it as an external dependency first.

Makefile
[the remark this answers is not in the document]
okurz: I would have not expected anything else because it's not automatically tested so will rot :) still serves as an example

Martchus:
I approved the PR because I'm not strictly against using Git subrepo here. However, I would also prefer to keep things separate here for reasons mentioned by @okurz . In addition, the subrepo approach is clearly *not* scaling as it is merely copying the contents after all. Having to do a separate checkout (e.g. just next to the repo) is not the worst if it is documented. It can even be simpler for the user because there are no new commands to learn about.

kalikiana:
> That is not really helpful though. It isn't packaged and doing that is certainly more work than subrepo/submodule/clone

If you already discussed this, you could have mentioned it 😉️

No objections from me then. The subrepo ensures that we consistently use the same version, especially if it isn't packaged.

kalikiana (dismissed):
> That is not really helpful though. It isn't packaged and doing that is certainly more work than subrepo/submodule/clone

If you already discussed this, you could have mentioned it 😉️

No objections from me then. The subrepo ensures that we consistently use the same version, especially if it isn't packaged.

perlpunk:
I will remove the subrepo.

Just commenting on the arguments:

> [A separate checkout] can even be simpler for the user because there are no new commands to learn about.

There is nothing to learn for the user. It's a `git subrepo clone` once, and no user has to care about it ever. For the user, it's the **easiest** solution.

Maybe you mean the maintainers.
Having a problem with learning new commands would speak against every new technology.

perlpunk:
I removed it. It will now be cloned when running `make test-unit`.

(If test-more-bash is ever updated and we want the new version, we need to make sure that the version is correct if it has already been cloned before.)

okurz (approved):

README.md
```diff
@@ -120,6 +134,21 @@ If this is too much hassle for you feel free to provide incomplete pull
 requests for consideration or create an issue with a code change proposal.
 
 ### Local testing
+
+#### Functional testing
+
+This is done with the [Test::More bash
+library](https://github.com/ingydotnet/test-more-bash).
+It will automatically cloned.
```
Martchus, line 142:
```suggestion
It will be automatically cloned.
```

Martchus (approved):

README.md
perlpunk: done

.github/workflows/ci.yml
```diff
@@ -7,7 +7,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Static checks, unit- and integration tests
+      - name: unit- and integration tests
+        run: |
+         make test-unit
+
+  style:
```
okurz, line 14: As you add a dedicated section here please also extend .mergify.yml in https://github.com/os-autoinst/scripts/blob/master/.mergify.yml#L11 and accordingly https://github.com/os-autoinst/scripts/blob/master/.mergify.yml#L20 with

```patch
       - "status-success~=test"
+      - "status-success~=style"
```

otherwise mergify would merge if tests pass but the style checks which you extended into a separate section could still fail. As an alternative of course you could keep them combined under simply "test". By the way, I would also prefer to not need to explicitly spell out each kind of test in .mergify as this is error prone as soon as we introduce another test job. In os-autoinst and openQA we could go with the alternative to look for the codecov report because only if *all* tests pass then also codecov is called. Maybe eventually we can reach something similar here as well with a fan-in job that combines all results into a final status check that we can check for in mergify

okurz (changes requested):
> > That is not really helpful though. It isn't packaged and doing that is certainly more work than subrepo/submodule/clone
> 
> If you already discussed this, you could have mentioned it wink

Well, we haven't discussed this in before just that I think asking for an external dependency to be packaged first is IMHO too much effort to get this PR merged.

@kalikiana your review is pending, would you like to approve this as well?

.github/workflows/ci.yml
perlpunk: Ok. I really like that the different tests are displayed on their own, so it's easy to see where the problem is.

Adding a test that succeeds when all others succeeded should be possible. But maybe for another PR in the future.

okurz (approved):

okurz (changes requested):
argh, wait. You need to touch *both* L11 and L22 in .mergify.yml

.github/workflows/ci.yml
okurz: please also add for https://github.com/os-autoinst/scripts/blob/master/.mergify.yml#L20

perlpunk: oh sorry. done

okurz (approved):

okurz:
With the changes to mergify config we need to merge manually

PR #124, opened 2021-12-20
Add option for the dependent jobs run

jankohoutek:
Add the option to run the jobs also with dependencies

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -27,7 +27,7 @@ def parse_args():
     parser.add_argument('--dry-run', action='store_true',
                         help='Do not do any action on openQA')
     parser.add_argument('--include-depends', action='store_true', default=False,
-                        help='Restart also the jobs on which investigated one is dependant')
+                        help='Restart also the jobs on which the investigated one depends')
     args = parser.parse_args()
     verbose_to_log = {
         0: logging.CRITICAL,
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -26,6 +26,8 @@ def parse_args():
                         help='The openQA test URL for which to trigger bisection investigation jobs')
     parser.add_argument('--dry-run', action='store_true',
                         help='Do not do any action on openQA')
+    parser.add_argument('--include-depends', action='store_true', default=False,
+                        help='Restart also the jobs on which investigated one is dependant')
     args = parser.parse_args()
     verbose_to_log = {
         0: logging.CRITICAL,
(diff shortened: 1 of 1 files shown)


openqa-trigger-bisect-jobs
```diff
@@ -44,7 +46,9 @@ def call(cmds, dry_run=False):
     return subprocess.call((['echo', 'Simulating: '] if dry_run else []) + cmds)
 
 
-def openqa_clone(cmds, dry_run, default_opts=['--skip-chained-deps', '--within-instance'], default_cmds=['_GROUP=0']):
+def openqa_clone(cmds, dry_run, include_depends, default_opts=['--skip-chained-deps', '--within-instance'], default_cmds=['_GROUP=0']):
+    if include_depends == True:
+      default_opts = ['--within-instance','--parental-inheritance']
```
okurz, line 51:
```suggestion
    if include_depends:
      default_opts = ['--within-instance','--parental-inheritance']
```

openqa-trigger-bisect-jobs
```diff
@@ -26,6 +26,8 @@ def parse_args():
                         help='The openQA test URL for which to trigger bisection investigation jobs')
     parser.add_argument('--dry-run', action='store_true',
                         help='Do not do any action on openQA')
+    parser.add_argument('--include-depends', action='store_true', default=False,
+                        help='Restart also the jobs on which investigated one is dependant')
```
okurz, line 30:
```suggestion
    parser.add_argument('--include-depends', action='store_true', default=False,
                        help='Restart also the jobs on which the investigated one depends')
```

okurz (changes requested):
please phrase the git commit message according to https://github.com/os-autoinst/scripts#rules-for-commits , i.e. with imperative mood

okurz (commented):
The option certainly does what you want but the idea for "openqa-trigger-bisect-jobs" was that it does what is needed to do to bisect any regression. And how should the caller decide if they need the option "--include-depends" or not? The problem that motivated you was a package change that affected the parent installation job already, right? I am wondering, if we would simply always do that, what would be the drawbacks?

kalikiana:
> The option certainly does what you want but the idea for "openqa-trigger-bisect-jobs" was that it does what is needed to do to bisect any regression. And how should the caller decide if they need the option "--include-depends" or not? The problem that motivated you was a package change that affected the parent installation job already, right? I am wondering, if we would simply always do that, what would be the drawbacks?

This doesn't seem to be linked to a ticket or issue but [#95783](https://progress.opensuse.org/issues/95783#note-41) touches on very similar points.

okurz:
closing due to no response. At any point feel free to address the open points and re-open.

PR #129, opened 2022-02-02
Further fixes for openqa-investigate

Martchus:
instead of running into Bash error about wrong field index

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -64,9 +64,15 @@ clone() {
     [[ $dry_run = 1 ]] && echo "$out"
 
     # keep track of handled jobs
-    for job_id in "${handled_ids[@]}"; do handled_job_ids["$job_id"]=1; done
-    # print name/urls
-    for url in "${urls[@]}"; do echo "* **$name**: $host_url${url/tests\//t}"; done
+    if [[ ${handled_ids[*]} ]]; then
+        for job_id in "${handled_ids[@]}"; do handled_job_ids["$job_id"]=1; done
+        # print name/urls
+        for url in "${urls[@]}"; do echo "* **$name**: $host_url${url/tests\//t}"; done
+    else
+        errors=$(echo "$out" | runjq -r '.errors[]')
+        echoerr "Unable to restart $id: ${errors:-no error message returned}"
+        return 2
+    fi
 }
 
 trigger_jobs() {
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -183,7 +183,7 @@ main() {
     client_prefix=''
     [ "$dry_run" = "1" ] && client_prefix="echo"
     set +u
-    if [[ -z "$client_call" ]]; then
+    if [[ -z "${client_call[*]}" ]]; then
         client_call=(openqa-cli "${client_args[@]}")
         if [[ -n "$client_prefix" ]]; then
             client_call=("$client_prefix" "${client_call[@]}")
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -151,7 +151,7 @@ investigate() {
     if [[ "$old_name" =~ ":investigate:" ]]; then
         [[ ! "$old_name" =~ investigate:retry$ ]] && echo "Job is ':investigate:' already, skipping investigation" && return 0
         investigate_origin="$(echo "$job_data" | runjq -r '.job.settings.OPENQA_INVESTIGATE_ORIGIN')" || return 1
-        origin_job_id=${investigate_origin#$host_url/t}
+        origin_job_id=${investigate_origin#"$host_url/t"}
         # meanwhile the original job might have been deleted already, handle
         # gracefully
         out=$("${client_call[@]}" -X POST jobs/"$origin_job_id"/comments text="Investigate retry job: $host_url/t$id failed, likely not a sporadic test" 2>/dev/null) || \
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -59,8 +59,8 @@ clone() {
 
     # clone job, possibly restarting dependencies
     out=$("${client_call[@]}" -X POST "jobs/$id/restart" 'clone=0' "prio=$((base_prio + prio_add))" "${restart_settings[@]}")
-    handled_ids=$(echo "$out" | runjq '.result[] | keys[] | tonumber')
-    urls=$(echo "$out" | runjq -r '.test_url[] | values[]')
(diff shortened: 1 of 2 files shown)


perlpunk:
shouldn't this:
```
handled_ids=$(echo "$out" | runjq '.result[] | keys[] | tonumber')
```
be this?
```
handled_ids=($(echo "$out" | runjq '.result[] | keys[] | tonumber'))
```
you are treating it as an array later

perlpunk:
What was the error message you got?
I can see this in the journal:
```
Feb 02 09:00:11 ariel openqa-gru[31671]: /opt/os-autoinst-scripts/openqa-investigate: line 67: 2172200
Feb 02 09:00:11 ariel openqa-gru[31671]: 2172201: syntax error in expression (error token is "2172201")
Feb 02 09:20:32 ariel openqa-gru[16443]: /opt/os-autoinst-scripts/openqa-investigate: line 67: 2172103
```

perlpunk:
Maybe this would be a good opportunity to add the `clone` function to `_common` and add automated tests for it.

openqa-investigate
```diff
@@ -19,13 +19,15 @@ exclude_name_regex="${exclude_name_regex:-":investigate:"}"
 exclude_no_group="${exclude_no_group:-"true"}"
 # exclude_group_regex checks a combined string "<parent job group name> / <job group name>"
 exclude_group_regex="${exclude_group_regex:-"Development.*/ "}"
-client_args=(api --header 'User-Agent: openqa-investigate (https://github.com/os-autoinst/scripts)' --host "$host_url")
+client_args=(api --header 'User-Agent: openqa-investigate (https://github.com/os-autoinst/scripts)' --host "$host_url" --apikey 1A0E49A69396DCD1 --apisecret F0F16CA0DD1B2F48 )
```
okurz, line 22: uuuuh …

Martchus: Already removed

openqa-investigate
```diff
@@ -59,14 +59,20 @@ clone() {
 
     # clone job, possibly restarting dependencies
     out=$("${client_call[@]}" -X POST "jobs/$id/restart" 'clone=0' "prio=$((base_prio + prio_add))" "${restart_settings[@]}")
-    handled_ids=$(echo "$out" | runjq '.result[] | keys[] | tonumber')
+    handled_ids=($(echo "$out" | runjq '.result[] | keys[] | tonumber'))
```
okurz, line 62:
```In openqa-investigate line 62:
    handled_ids=($(echo "$out" | runjq '.result[] | keys[] | tonumber'))
                 ^-- SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
```

perlpunk:
Looking at the log output, I think that's not enough testing:
```
unable to query job data for 42: 
openqa-investigate: line 62: -X: command not found
Unable to restart 42: no error message returned
test/02-investigate.t ......... ok
```
`client_call` is never set, so it doesn't even call `openqa-cli`.

Also I think we should avoid unhandled output in the test output.

openqa-investigate
```diff
@@ -193,4 +200,4 @@ main() {
     exit "$error_count"
 }
 
-main "$@"
+[[ ${TESTING:-} ]] || main "$@"
```
perlpunk, line 203: To avoid the extra variable, this _could_ be replaced by
```
caller 0 >/dev/null || main "$@"
```
but I admit that the documentation is not very clear on that.

perlpunk: I'm also a fan of putting functions into a "library" (here `_common`) and source this instead, but admittedly in bash that can be more complicated because of global variables

okurz: I'd like the caller approach. For the second suggestion Martchus and me decided to keep the functions in this file as long as the functions are actually not shared with other scripts

openqa-investigate
```diff
@@ -6,8 +6,9 @@
 # http://redsymbol.net/articles/unofficial-bash-strict-mode/
 set -euo pipefail
 
+scriptdir=$(dirname "${BASH_SOURCE[0]}")
```
okurz, line 9: You don't need to define a variable, just use it in-place

okurz (changes requested):
overlooked tina's comment

okurz:
oh, too late :D

perlpunk:
> overlooked tina's comment

I should have requested changes instead of just comment

perlpunk:
And it's not only about unhandled output:
`client_call is never set, so it doesn't even call openqa-cli.`
If you look at the test, it's also wrong. I didn't mentioned it because I thought my comment would make it obvious:
```
cli_rc=1
openqa-cli() {
    return $cli_rc
}

rc=0
clone 41 42 || rc=$?
is "$rc" 1 'fails when unable to query job data'

cli_rc=0
clone 41 42 || rc=$?
is "$rc" 2 'fails when no jobs could be restarted'
```
So how can the last `rc` be 2 if you set `cli_rc` to 0?

okurz:
> And it's not only about unhandled output: `client_call is never set, so it doesn't even call openqa-cli.` If you look at the test, it's also wrong. I didn't mentioned it because I thought my comment would make it obvious:
> 
> ```
> cli_rc=1
> openqa-cli() {
>     return $cli_rc
> }
> 
> rc=0
> clone 41 42 || rc=$?
> is "$rc" 1 'fails when unable to query job data'
> 
> cli_rc=0
> clone 41 42 || rc=$?
> is "$rc" 2 'fails when no jobs could be restarted'
> ```
> 
> So how can the last `rc` be 2 if you set `cli_rc` to 0?

Well, we just return a faked successful run of the client but it's not returning any useful job data so we end up in the error handling branch that Martchus introduced in this PR which returns 2 in line 74.

perlpunk:
@okurz  no, openqa-cli is never called:

Look at the code!
Test output:
```
unable to query job data for 42: 
26
openqa-investigate: line 61: -X: command not found
27
Unable to restart 42: no error message returned
28
test/02-investigate.t ......... ok
```
code, line 61:
```
out=$("${client_call[@]}" -X POST "jobs/$id/restart" 'clone=0' "prio=$((base_prio + prio_add))" "${restart_settings[@]}")
```
client_call is never set because we never call `main`!

okurz:
Yes, ok. I mean that we skip over an earlier invocation of openqa-cli (line 41) by faking a successful execution and later we silently ignore the error of unset client_call (ugly and should be improved) and that however helps us to reach into the API error handling branch.

PR #130, opened 2022-02-02
Fix URL in openqa-investigate comments (to include scheme and host)

okurz:


okurz:
This is likely still not enough because somehow test results don't show up any more and links are longer now, e.g. see https://openqa.opensuse.org/tests/2172606#comments

Martchus (approved):

okurz:
pair-programmed with mkittler

PR #131, opened 2022-02-03
Exclude jobs with problematic dependencies again (until problems are fixed)

Martchus:
There are several problems with restarting MM jobs, see
https://progress.opensuse.org/issues/95783#note-14.

This change brings back the guard for excluding those jobs we had in
9845f20fc2836b62b7c90f2bb735fa568fa0a361 but allows to reconsider those
jobs by setting a variable.

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -19,6 +19,7 @@ exclude_name_regex="${exclude_name_regex:-":investigate:"}"
 exclude_no_group="${exclude_no_group:-"true"}"
 # exclude_group_regex checks a combined string "<parent job group name> / <job group name>"
 exclude_group_regex="${exclude_group_regex:-"Development.*/ "}"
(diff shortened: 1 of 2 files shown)


Martchus:
Adding the not-ready label as this is only for the case we decide it is necessary after all.

okurz (changes requested):
Commented in the ticket. TL;DR: keep 'em enabled

https://stats.openqa-monitor.qa.suse.de/d/nRDab3Jiz/openqa-jobs-test?viewPanel=12&orgId=1&from=now-7d&to=now shows that we reached near-zero last night except for ppc so at least MM investigation jobs don't completely block everything.

But good that you have this prepared. This can wait for more feedback and from other users.

okurz:
You could rebase for fun and easy merging in case we still need it, e.g. if over the weekend an instance goes havok :)

PR #136, opened 2022-02-28
Improve openqa-trigger-bisect-jobs

perlpunk:

* Return early in case the job does not have (changed) OS_TEST_ISSUES settings
* Print informative error message in case of
  * Invalid JSON
  * Unsuccessful HTTP request
* Add tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ test: checkstyle test-unit
 .PHONY: test-unit
 test-unit: test-more-bash
 	prove -r test/
+	py.test
 
 test-more-bash:
 	git clone https://github.com/ingydotnet/test-more-bash.git --depth 1 -b 0.0.5
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -6,10 +6,13 @@ import re
 import requests
 import subprocess
 import sys
-from urllib.parse import urljoin, urlparse
+import json
+from urllib.parse import urljoin, urlparse, urlunparse
 
 logging.basicConfig()
 log = logging.getLogger(sys.argv[0] if __name__ == '__main__' else __name__)
+GOOD = '-'
+BAD = '+'
 
 
 class CustomFormatter(argparse.ArgumentDefaultsHelpFormatter, argparse.RawDescriptionHelpFormatter):
@@ -47,34 +50,49 @@ def call(cmds, dry_run=False):
 def openqa_clone(cmds, dry_run, default_opts=['--skip-chained-deps', '--within-instance'], default_cmds=['_GROUP=0']):
     return call(['openqa-clone-job'] + default_opts + cmds + default_cmds, dry_run)
 
+def fetch_url(url, request_type='text'):
+    try:
+        content = requests.get(url.geturl())
+        content.raise_for_status()
+    except requests.exceptions.RequestException as e:
+        log.error("Error while fetching %s: %s"% (url.geturl(), str(e)))
+        raise(e)
+    raw = content.content
+    if request_type == 'json':
+        try:
+            content = content.json()
+        except json.decoder.JSONDecodeError as e:
+            log.error("Error while decoding JSON from %s -> >>%s<<: %s" % (url.geturl(), raw, str(e)))
+            raise(e)
+    return content
 
-def main():
-    args = parse_args()
-    investigation_url = urljoin(args.url + '/', 'investigation_ajax')
-    log.debug('Retrieving investigation info from %s' % investigation_url)
-    out = requests.get(investigation_url)
-    log.debug('Received investigation info: %s' % out)
-    investigation = out.json()
-    os_test_issues = [line for line in investigation['diff_to_last_good'].split('\n') if 'OS_TEST_ISSUES' in line]
-    if len(os_test_issues) != 2:
-        return
-    good, bad = [set(re.compile(': "([^"]*)",').search(line).group(1).split(',')) for line in os_test_issues]
-    removed, added = list(good - bad), list(bad - good)
-    log.debug('removed: %s, added: %s' % (removed, added))
+def main(args):
     parsed_url = urlparse(args.url)
-    test_url = urljoin(parsed_url.geturl(), '/api/v1/jobs/' + parsed_url.path.lstrip('/tests/'))
-    log.debug('Retrieving additional job data from %s' % test_url)
-    test_data = requests.get(test_url)
+    investigation_url = urlparse(urlunparse((parsed_url.scheme, parsed_url.netloc, parsed_url.path + '/investigation_ajax', '', '', '')))
+    log.debug('Retrieving investigation info from %s' % investigation_url.geturl())
+    investigation = fetch_url(investigation_url, request_type='json')
+    log.debug('Received investigation info: %s' % investigation)
+    changes = { GOOD: set(), BAD: set() }
+    for line in investigation['diff_to_last_good'].split('\n'):
+        search = re.compile('([+-]) +"OS_TEST_ISSUES" *: "([^"]*)",').search(line)
+        if search:
+            changes[search.group(1)] = set(search.group(2).split(','))
+    if not changes[BAD] or not changes[GOOD]:
+        return
+    removed, added = list(changes[GOOD] - changes[BAD]), list(changes[BAD] - changes[GOOD])
+    log.debug('removed: %s, added: %s' % (removed, added))
+    test_url = urlparse(urlunparse((parsed_url.scheme, parsed_url.netloc, '/api/v1/jobs/' + parsed_url.path.lstrip('/tests/'), '', '', '')))
+    log.debug('Retrieving additional job data from %s' % test_url.geturl())
+    test_data = fetch_url(test_url, request_type='json')
     log.debug('Received job data: %s' % test_data)
-    test = test_data.json()['job']['settings']['TEST']
+    test = test_data['job']['settings']['TEST']
     log.debug("Found test name '%s'" % test)
-    for issue in added:
+    for issue in sorted(added):
         log.info("Triggering one bisection job without issue '%s'" % issue)
-        new = ','.join(bad - set([issue]))
+        new = ','.join(sorted(list(changes[BAD] - set([issue]))))
         log.debug("New set of OS_TEST_ISSUES='%s'" % new)
         test_name = test + ':investigate:bisect_without_%s' % issue
         openqa_clone([args.url, 'OS_TEST_ISSUES=' + new, 'TEST=' + test_name, 'OPENQA_INVESTIGATE_ORIGIN=' + args.url], args.dry_run)
 
-
 if __name__ == '__main__':
-    main()
+    main(parse_args())
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - run: |
+          pip install --upgrade pip
+          pip install pytest mocker pytest-mock
+          python --version
+          pytest --version
       - name: unit- and integration tests
         run: |
          make test-unit
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 /test-more-bash/
+/tests/__pycache__/
+/__pycache__/
--- a/tests/data/python-requests/openqa.opensuse.org/api/v1/jobs/7848818
+++ b/tests/data/python-requests/openqa.opensuse.org/api/v1/jobs/7848818
@@ -1,0 +1,15 @@
+{
+  "job": {
+    "id": 7848818,
+    "priority": 50,
+    "result": "failed",
+    "settings": {
+      "OS_TEST_ISSUES": "21637,21770,21926,21954,22030,22077,22085,22192",
+      "TEST": "foo"
+    },
+    "state": "done",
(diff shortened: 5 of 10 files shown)


test/02-trigger-bisect-jobs.t
```diff
@@ -0,0 +1,42 @@
...
+openqa-clone-job --skip-chained-deps --within-instance file://openqa.opensuse.org/tests/7848818 OS_TEST_ISSUES=21637,21770,21926,21954,22030,22077,22085 TEST=foo:investigate:bisect_without_22192 OPENQA_INVESTIGATE_ORIGIN=file://openqa.opensuse.org/tests/7848818 _GROUP=0
+END
+)
+is "$out" "$expected" "bisect works as expected"
+
+out=$(./openqa-trigger-bisect-jobs -vv --url file://openqa.opensuse.org/tests/123 2>&1) || true
+like "$out", "Error while decoding JSON" "JSON error reported"
+
+out=$(./openqa-trigger-bisect-jobs -vv --url file://openqa.opensuse.org/tests/1234 2>&1) || true
+is "$out" "" "Empty output where we do not have two OS_TEST_ISSUESlines"
```
okurz, line 36:
```suggestion
is "$out" "" "Empty output where we do not have two OS_TEST_ISSUES lines"
```

test/02-trigger-bisect-jobs.t
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
```
okurz, line 1: It's unconventional to have a shell script as test for a python script. For now it looks simple enough but I guess soon we should go for pytest

openqa-trigger-bisect-jobs
```diff
@@ -47,28 +48,60 @@ def call(cmds, dry_run=False):
...
+        path = 'test/data/python-requests/' + path
+        with open(path, "r") as request:
+            raw = content = request.read()
+    elif url.scheme in ["http", "https"]:
+        decodejson = lambda c: content.json()
+        try:
+            content = requests.get(url.geturl())
+        except requests.exceptions.RequestException as e:
+            log.error("Error while fetching "+url.geturl()+": "+str(e))
+            raise(e)
```
okurz, line 65: why catch these cases at all when you don't handle them except for logging?

openqa-trigger-bisect-jobs
```diff
@@ -47,28 +48,60 @@ def call(cmds, dry_run=False):
 def openqa_clone(cmds, dry_run, default_opts=['--skip-chained-deps', '--within-instance'], default_cmds=['_GROUP=0']):
     return call(['openqa-clone-job'] + default_opts + cmds + default_cmds, dry_run)
 
+def fetch_url(url, request_type='text'):
+    if url.scheme == "file":
+        decodejson = lambda c: json.loads(c)
+        path = url.geturl()
+        path = path[7:]
```
okurz, line 55: I would avoid the magic number. I guess you want to strip the initial `file://` part? Then there are better ways to do that with an url object

openqa-trigger-bisect-jobs
```diff
@@ -47,28 +48,60 @@ def call(cmds, dry_run=False):
...
+    log.debug('Received investigation info: %s' % investigation)
+    good = bad = set()
+    for line in investigation['diff_to_last_good'].split('\n'):
+        search = re.compile('([+-]) +"OS_TEST_ISSUES" *: "([^"]*)",').search(line)
+        if search:
+            issues = set(search.group(2).split(','))
+            if search.group(1) == '+':
+                bad = issues
+            else:
+                good = issues
```
okurz, line 91: I would try to avoid this procedural code. Another aspect is that it's less efficient although that's not a strong concern here. Why is it necessary to change this? What problematic cases did you encounter?

openqa-trigger-bisect-jobs
```diff
@@ -47,28 +48,60 @@ def call(cmds, dry_run=False):
...
     args = parse_args()
-    investigation_url = urljoin(args.url + '/', 'investigation_ajax')
-    log.debug('Retrieving investigation info from %s' % investigation_url)
-    out = requests.get(investigation_url)
-    log.debug('Received investigation info: %s' % out)
-    investigation = out.json()
-    os_test_issues = [line for line in investigation['diff_to_last_good'].split('\n') if 'OS_TEST_ISSUES' in line]
-    good, bad = [set(re.compile(': "([^"]*)",').search(line).group(1).split(',')) for line in os_test_issues]
+    parsed_url = urlparse(args.url)
+    investigation_url = parsed_url._replace(path=parsed_url.path + '/investigation_ajax')
```
okurz, line 79: there is a reason why the method `_replace` is internal. We should not use it

openqa-trigger-bisect-jobs
```diff
@@ -47,28 +48,60 @@ def call(cmds, dry_run=False):
...
     return call(['openqa-clone-job'] + default_opts + cmds + default_cmds, dry_run)
 
+def fetch_url(url, request_type='text'):
+    if url.scheme == "file":
+        decodejson = lambda c: json.loads(c)
+        path = url.geturl()
+        path = path[7:]
+        path = 'test/data/python-requests/' + path
+        with open(path, "r") as request:
+            raw = content = request.read()
```
okurz, line 58: this looks like you want to replicate more or less what https://github.com/os-autoinst/openqa_review/blob/master/openqa_review/browser.py does with loading/saving for caching+testing purposes, is it? If we need to do the same here I wonder if there isn't already a better way in the python community to mock requests for test data

okurz (changes requested):

openqa-trigger-bisect-jobs
perlpunk: Because the exception itself does not print the url and/or content it failed for?
Remember the problems we had with the gru logs, having failures where we neither had the url no the content?
Do we want  to have the same problems back after we add openqa-trigger-bisect-jobs or should I rather add the exception handling **now**?

openqa-trigger-bisect-jobs
perlpunk: Well, the previous code was failing awfully uninformative when the regex didn't match. `OS_TEST_ISSUES` can be part of another setting, who knows? And then we will have these error messages that there is no attribute `group` in the gru log.

If you think the script is fine as it is, then I will unassign myself from the progress ticket again.

openqa-trigger-bisect-jobs
perlpunk: Do you have another idea how to set the path of `parsed_url`?
Like `$url->path('/foo/bar');` for Mojo::URL? I searched for it and didn't find anything.

openqa-trigger-bisect-jobs
```diff
@@ -47,28 +48,60 @@ def call(cmds, dry_run=False):
 def openqa_clone(cmds, dry_run, default_opts=['--skip-chained-deps', '--within-instance'], default_cmds=['_GROUP=0']):
     return call(['openqa-clone-job'] + default_opts + cmds + default_cmds, dry_run)
 
+def fetch_url(url, request_type='text'):
+    if url.scheme == "file":
+        decodejson = lambda c: json.loads(c)
+        path = url.geturl()
+        path = path[7:]
+        path = 'test/data/python-requests/' + path
```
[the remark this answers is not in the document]
perlpunk: That won't work.
`path` is a local path.
For a url `file://openqa.opensuse.org/tests/123/investigation_ajax` the local path will be
`test/data/python-requests/openqa.opensuse.org/tests/123/investigation_ajax`, so the hostname is included.

test/02-trigger-bisect-jobs.t
perlpunk: done

openqa-trigger-bisect-jobs
perlpunk: can you give an example of a better way?

perlpunk:
Hm. Seems like tests are not run for drafts?

perlpunk:
I can't trigger the CI. Something's wrong.

openqa-trigger-bisect-jobs
```diff
@@ -47,34 +50,49 @@ def call(cmds, dry_run=False):
 def openqa_clone(cmds, dry_run, default_opts=['--skip-chained-deps', '--within-instance'], default_cmds=['_GROUP=0']):
     return call(['openqa-clone-job'] + default_opts + cmds + default_cmds, dry_run)
 
+def fetch_url(url, request_type='text'):
+    try:
+        content = requests.get(url.geturl())
+        content.raise_for_status()
+    except requests.exceptions.RequestException as e:
+        log.error("Error while fetching "+url.geturl()+": "+str(e))
```
Martchus, line 58:
```suggestion
        log.error("Error while fetching " + url.geturl() + ": " + str(e))
```
It would be nice to use spaces in other occurrences as well.

Makefile
```diff
@@ -7,6 +7,7 @@ test: checkstyle test-unit
 .PHONY: test-unit
 test-unit: test-more-bash
 	prove -r test/
+	py.test tests/trigger-bisect-jobs.py
```
okurz, line 10: Make that

```suggestion
	py.test
```

and rename tests/trigger-bisect-jobs.py to tests/test_trigger_bisect_jobs.py

okurz (changes requested):

openqa-trigger-bisect-jobs
Martchus: I've just read https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse and it only returns a named tuple but one cannot set the attributes manually. The docs even state the usage of `_replace` explicitly in one example so I suppose it isn't the worst to use it - you're just following an example from official documentation here.

Martchus: Maybe it is still cleaner to simply construct the `ParseResult` object you want directly, e.g.:

```
from urllib.parse import ParseResult
ParseResult(scheme='', netloc='', params='', query='', fragment='', path='/foo/bar').geturl()
```

Btw, that kind of construction is also mentioned in https://docs.python.org/3/librar/urllib.parse.html#urllib.parse.urlparse if you scroll down a bit.

openqa-trigger-bisect-jobs
perlpunk: done (I'm using `%` now)

Makefile
perlpunk: done

openqa-trigger-bisect-jobs
perlpunk: obsolete

openqa-trigger-bisect-jobs
perlpunk: replaced with a dictionary

perlpunk:
https://github.com/os-autoinst/scripts/runs/5736548596?check_suite_focus=true#step:4:36
```
==================================== ERRORS ====================================
________________________ ERROR at setup of test_trigger ________________________
file /home/runner/work/scripts/scripts/tests/test_trigger_bisect_jobs.py, line 46
  def test_trigger(mocker):
E       fixture 'mocker' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, doctest_namespace, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.

/home/runner/work/scripts/scripts/tests/test_trigger_bisect_jobs.py:46
_______________________ ERROR at setup of test_problems ____________________
...
```

perlpunk:
Apparently in some versions of pytest one has to add `@pytest.fixture` above a function, in order to use `mocker`.
Funnily, locally I did not need this.
There is more than one way to do it, python, and it totally depends on where something is running. Why?

Now the leftover failure is the warning
```
tests/test_trigger_bisect_jobs.py:8
36
  /home/runner/work/scripts/scripts/tests/test_trigger_bisect_jobs.py:8: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
37
    import imp
```
So I can now either spend my time trying to learn how to make pytest ignore this warning for now, or I have to give `importlib` a second try, despite its bad documentation.

**Any help appreciated!**

openqa-trigger-bisect-jobs
nicksinger: I want to mention https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunparse regarding url construction.

perlpunk: I replaced `_replace()` with one of the correct ways to do it (apparently there is more than one way!):
```
investigation_url = urlparse(urlunparse((parsed_url.scheme, parsed_url.netloc, parsed_url.path + '/investigation_ajax', '', '', '')))
```
beautiful!

okurz:
> **Any help appreciated!**

Seems like you could fix it? The check output looks clean now.

perlpunk:
For the record:
https://stackoverflow.com/questions/19009932/import-arbitrary-python-source-file-python-3-3
It says, that the code I use (load_module) will be deprecated, too, and suggests an alternative.
I tried:
```
import importlib.machinery
import importlib.util
 loader = importlib.machinery.SourceFileLoader('openqa', rootpath + '/openqa-trigger-bisect-jobs')
spec = importlib.util.spec_from_loader(loader.name, loader)
openqa = importlib.util.module_from_spec(spec)
loader.exec_module(openqa)
```
but I get `ModuleNotFoundError: No module named 'openqa'`

PR #141, opened 2022-04-06
Add openqa-trigger-bisect-jobs to hook

perlpunk:

I also added a unit test for the hook script.

Manual test: https://openqa.opensuse.org/tests/2282708
Bisect jobs:
* https://openqa.opensuse.org/tests/2284335
* https://openqa.opensuse.org/tests/2284336

### DIFF
--- a/openqa-label-known-issues-and-investigate-hook
+++ b/openqa-label-known-issues-and-investigate-hook
@@ -10,4 +10,14 @@ id="${1:?"Need 'job_id'"}"
 host="${host:-"openqa.opensuse.org"}"
 scheme="${scheme:-"https"}"
 host_url="$scheme://$host"
-echo "$host_url/tests/$id" | "$(dirname "$0")"/openqa-label-known-issues 3>&1 1>/dev/null 2>&3- | sed -n 's/\(\S*\) : Unknown issue, to be reviewed.*$/\1/p' | "$(dirname "$0")"/openqa-investigate
+dirname=$(dirname "$0")
+script_dir=${script_dir:-$dirname}
+
+investigate-and-bisect() {
+    read -r test
+    echo "$test" | "$script_dir/openqa-investigate"
+    "$script_dir/openqa-trigger-bisect-jobs" --url "$test"
+}
+
+echo "$host_url/tests/$id" | "$script_dir/openqa-label-known-issues" 3>&1 1>/dev/null 2>&3- | sed -n 's/\(\S*\) : Unknown issue, to be reviewed.*$/\1/p' | investigate-and-bisect
+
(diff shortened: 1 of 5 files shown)


openqa-label-known-issues-and-investigate-hook
```diff
@@ -10,4 +10,14 @@ id="${1:?"Need 'job_id'"}"
...
 scheme="${scheme:-"https"}"
 host_url="$scheme://$host"
-echo "$host_url/tests/$id" | "$(dirname "$0")"/openqa-label-known-issues 3>&1 1>/dev/null 2>&3- | sed -n 's/\(\S*\) : Unknown issue, to be reviewed.*$/\1/p' | "$(dirname "$0")"/openqa-investigate
+dirname=$(dirname "$0")
+script_dir=${script_dir:-$dirname}
+
+investigate-and-bisect() {
+    read -r test
+    echo "$test" | "$script_dir/openqa-investigate"
+    "$script_dir/openqa-trigger-bisect-jobs" --url "$test"
```
Martchus, line 19: Looks like `openqa-trigger-bisect-jobs` will break jobs with directly chained dependencies as it passes `--skip-chained-deps` as `openqa-clone-job` argument. I suppose  `openqa-trigger-bisect-jobs` needed to be adjusted similar to `openqa-investigate` until we can use it in production for jobs with directly chained and parallel dependencies (see https://progress.opensuse.org/issues/95783#note-39). So it would make sense to add an early-abort in `openqa-trigger-bisect-jobs` (or before we call it) for such jobs (similar to `openqa-trigger-bisect-jobs`).

Martchus (changes requested):

openqa-label-known-issues-and-investigate-hook
perlpunk: > So it would make sense to add an early-abort in openqa-trigger-bisect-jobs (or before we call it) for such jobs

How can I find out if it is "such a job"?

perlpunk: You mean this check?
```
'(.job.children["Parallel"] | length) + (.job.parents["Parallel"] | length) + (.job.children["Directly chained"] | length) + (.job.parents["Directly chained"] | length)'
```

Martchus: Yes, this check.

perlpunk: #142

perlpunk: #142 merged

Martchus (approved):

PR #144, opened 2022-04-11
Bisect only failed jobs with diff_to_last_good

perlpunk:
But I saw this in our logs:

    Traceback (most recent call last):
      File "/opt/os-autoinst-scripts/openqa-trigger-bisect-jobs", line 120, in <module>
        main(parse_args())
      File "/opt/os-autoinst-scripts/openqa-trigger-bisect-jobs", line 79, in main
        for line in investigation['diff_to_last_good'].split('\n'):
    KeyError: 'diff_to_last_good'

Jobs might not have a diff_to_last_good entry.

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -76,6 +76,8 @@ def main(args):
     investigation = fetch_url(investigation_url, request_type='json')
     log.debug('Received investigation info: %s' % investigation)
     changes = { GOOD: set(), BAD: set() }
+    if 'diff_to_last_good' not in investigation:
+        return
     for line in investigation['diff_to_last_good'].split('\n'):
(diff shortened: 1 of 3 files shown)


okurz (changes requested):
The change is ok but this is not about "failed" or "passed" but rather that there can be jobs which do not have investigation info or jobs that never had a last good but always failed.

Martchus (approved):

perlpunk:
I looked into the gru journal, looked at the error timestamp, then looked into the openqa_gru log fo the timestamp, found a minion id, for example https://openqa.suse.de/minion/jobs?id=4258325
And then I looked into the job https://openqa.suse.de/tests/8506934

mimi1vx (approved):

okurz (approved):

PR #145, opened 2022-04-12
Email notifications for unreviewed issues

perlpunk:
Issue: https://progress.opensuse.org/issues/91605

I tested it on o3 with
```
echo http://openqa.opensuse.org/tests/2270756 | host=openqa.opensuse.org scheme=http ./openqa-label-known-issues
```
The email looks like:
```
Subject: Unknown issue to be reviewed (Group 24)                                                                    

http://openqa.opensuse.org/tests/2270756 : Unknown issue, to be reviewed ->
http://openqa.opensuse.org/tests/2270756/file/autoinst-log.txt
Reason: null
```

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -17,6 +17,8 @@ issue_marker="${issue_marker:-"auto_review%3A"}"
 issue_query="${issue_query:-"https://progress.opensuse.org/projects/openqav3/issues.json?limit=200&subproject_id=*&subject=~${issue_marker}"}"
 reason_min_length="${reason_min_length:-"8"}"
 grep_timeout="${grep_timeout:-5}"
+email_unreviewed=${email_unreviewed:-true}
+from_email=${from_email:-openqa-label-known-issues@open.qa}
 to_review=()
 curl_args=(--user-agent "openqa-label-known-issues")
 client_args=(api --header 'User-Agent: openqa-label-known-issues (https://github.com/os-autoinst/scripts)' --host "$host_url")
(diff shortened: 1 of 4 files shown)


openqa-label-known-issues
```diff
@@ -17,6 +17,8 @@ issue_marker="${issue_marker:-"auto_review%3A"}"
 issue_query="${issue_query:-"https://progress.opensuse.org/projects/openqav3/issues.json?limit=200&subproject_id=*&subject=~${issue_marker}"}"
 reason_min_length="${reason_min_length:-"8"}"
 grep_timeout="${grep_timeout:-5}"
+email_unreviewed=${email_unreviewed:-true}
+from_email=${from_email:-okurz@suse.de}
```
okurz, line 21: Why me? :)

How about

```suggestion
from_email=${from_email:-openqa-label-known-issues@open.qa}
```

which is of course not a valid email address and we should select something infrastructure specfiic wherever we use it

test/data/group24.json
```diff
@@ -0,0 +1,18 @@
...
+    "exclusively_kept_asset_size": 37640732672,
+    "id": 24,
+    "keep_important_logs_in_days": 365,
+    "keep_important_results_in_days": 3650,
+    "keep_logs_in_days": 60,
+    "keep_results_in_days": 90,
+    "name": "openQA",
+    "parent_id": null,
+    "size_limit_gb": 40,
+    "sort_order": 26
```
okurz, line 16: Please delete all keys that we don't care about to make that test data easier to understand.

okurz (changes requested):

openqa-label-known-issues
perlpunk: done

test/data/group24.json
perlpunk: done

Martchus (approved):
Considering the manual test and unit tests I suppose it works :-)
The mails are then supposed to show up in Slack, right?

okurz:
> Considering the manual test and unit tests I suppose it works :-) The mails are then supposed to show up in Slack, right?

That's a possibility, yes. But basically we can try to develop that independantly and leave it to job group owners to decide on which email address the put there as target.

okurz (approved):

perlpunk:
> The mails are then supposed to show up in Slack, right?

Yes, you should have seen one today at 12:00 in our channel.

PR #147, opened 2022-04-20
openqa-investigate: Use the right word

Vogtinator:


okurz:
Thank you. I stumbled over this so many times and always forgot to make that trivial but important fix :)

Vogtinator:
Same! When I saw it today I decided it will be the last time.

PR #149, opened 2022-04-26
Ensure writable homedir when calling mutt

perlpunk:
Sometimes we run as a user with a readonly homedir, and mutt
wants to create a 'sent' file. Haven't found an easy way to tell
mutt not to create that file.

Issue: https://progress.opensuse.org/issues/91605


kalikiana (approved):
This looks a bit ugly, but I also couldn't find an alternative after reading mutt docs

PR #153, opened 2022-05-04
Only consider queries in subproject

kalikiana:
Fixes: https://progress.opensuse.org/issues/109857


okurz (changes requested):
This would go too far. We don't want to shutdown auto-review tickets in general, only the "openqa-review+force-result" ones. So it can not be as simple as replacing the query. It needs to be more fine grained to look into only the new project (or maybe even better the tracker) but only for the force-result tickets. How about keeping this query at is but only going into the path for "force-result" if the tracker is of type "openqa-force-result"?

okurz:
https://github.com/os-autoinst/scripts/pull/154 might help.

If not you could extend the condition in _common:63

kalikiana:
> #154 might help.
> 
> If not you could extend the condition in _common:63

I suppose we don't need the new project then. The tracker is what locks down permissions. `comment_on_job` will never be reached with your change.

PR #155, opened 2022-05-06
Ensure any bugref in force_result labels to be parsed as a real bugref

okurz:
If openQA would detect a label in a line within a comment then only a
label would be referenced however for auto-review using force_result
labels I think we should highlight the actual ticket used as reference.
This commit is putting the comment (duplicated) in a separate line so
that openQA will render the comment with an according bugref link, not
the label per se.

Related progress issue: https://progress.opensuse.org/issues/109857

### DIFF
--- a/_common
+++ b/_common
@@ -61,7 +61,9 @@ runcurl() {
 comment_on_job() {
     local id=$1 comment=$2 force_result=${3:-''}
     if $enable_force_result && [[ -n $force_result ]]; then
-        comment="label:force_result:$force_result:$comment"
+        comment="label:force_result:$force_result:$comment
+
+$comment"
     fi
     "${client_call[@]}" -X POST jobs/"$id"/comments text="$comment"
 }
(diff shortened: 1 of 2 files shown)


_common
```diff
@@ -61,7 +61,9 @@ runcurl() {
 comment_on_job() {
     local id=$1 comment=$2 force_result=${3:-''}
     if $enable_force_result && [[ -n $force_result ]]; then
-        comment="label:force_result:$force_result:$comment"
+        comment="label:force_result:$force_result:$comment
+
+$comment"
```
kalikiana, line 66: I realize this is mentioned in the description... but duplication seems like a bad idea to me. This should be fixed on the openQA side

kalikiana (changes requested):

_common
okurz: Let's add this change as is for now and discuss a change in openQA later, can we? :)

PR #161, opened 2022-05-19
Add log excerpt to notification emails

perlpunk:


perlpunk:
that was fast. i just added something missing. but it will be ok. will make a followup pr

PR #162, opened 2022-05-20
Emphasize the important of specific search terms

kalikiana:
Adding emphasis via formatting based on a suggestion by Ioannis


README.md
```diff
@@ -43,8 +43,8 @@ example using the `<search_term>`
   `(?s)something to match earlier.*something to match some lines further down`
 
 Other double quotes in the subject line than around the search term should be
-avoided. Also be careful to not specify a too generic search term to prevent
-false matches of job failures unrelated to the specified ticket.
+avoided. Also be careful to **not specify a too generic search term to prevent
+false matches** of job failures unrelated to the specified ticket.
```
b10n1k, line 47: i was thinking of it more like a warning. What is you quote this with `>` in a new line, if it is not too much trouble. wdyt?

b10n1k (changes requested):
i think it doesnt emphasize a lot with bold (at least not on my black mode browser). So i had a small suggestion. if not i could go with just *bold*

Martchus:
s/the important/the importance/

README.md
kalikiana: Like this? That looks weird imho. This isn't a quote.

> Also be careful to **not specify a too generic search term to prevent
false matches** of job failures unrelated to the specified ticket.

Martchus: > That looks weird imho.

True.

b10n1k: not sure why this is weird but it stands out as a warning. Anyway it was a suggestion. I am not expert in "look and feel" of the page and maybe is better to merge and figure it out later. Feel free to go on as you like

kalikiana: @b10n1k Like I said, I think it's weird because it looks like a quote 🙂️

I'm thinking easier phrasing makes the emphasis stand out more.

perlpunk (approved):

PR #163, opened 2022-05-20
Render job group name related to reviewed issues

kalikiana:


_common
```diff
@@ -116,7 +116,8 @@ handle_unknown() {
         group_mailto=$(echo "$group_description" | perl -n -wE'm/.*MAILTO: (\S+).*/ and say $1' | head -1)
         group_mailto=${group_mailto:-$notification_address}
         if [[ -n "$group_mailto" ]]; then
-            echo "$header"$'\n'"Reason: $reason"$'\n'"$excerpt" | mailx -s "Unknown issue to be reviewed (Group $group_id)" -r "openqa-label-known-issues <$from_email>" "$group_mailto"
+            group_description=$(echo "$group_data" | runjq -r '.[0].name' ) || true
+            echo "$header"$'\n'"Reason: $reason"$'\n'"$excerpt" | mailx -s "Unknown issue to be reviewed (Group $group_id $group_name)" -r "openqa-label-known-issues <$from_email>" "$group_mailto"
```
Martchus, line 120: Where will `group_name` come from?

_common
```diff
@@ -116,7 +116,8 @@ handle_unknown() {
         group_mailto=$(echo "$group_description" | perl -n -wE'm/.*MAILTO: (\S+).*/ and say $1' | head -1)
         group_mailto=${group_mailto:-$notification_address}
         if [[ -n "$group_mailto" ]]; then
-            echo "$header"$'\n'"Reason: $reason"$'\n'"$excerpt" | mailx -s "Unknown issue to be reviewed (Group $group_id)" -r "openqa-label-known-issues <$from_email>" "$group_mailto"
+            group_description=$(echo "$group_data" | runjq -r '.[0].name' ) || true
```
b10n1k, line 119: Does this conflict with [L115](https://github.com/os-autoinst/scripts/pull/163/files#diff-d43d416085698e2b7d04ae7e385fec724b027e46fb9870fb76ada26a8bac931fR115)

b10n1k (commented):
Something doesnt feel right. where is the group_name definition?

_common
okurz, line 120:
```suggestion
            group_name=$(echo "$group_data" | runjq -r '.[0].name' )
            echo "$header"$'\n'"Reason: $reason"$'\n'"$excerpt" | mailx -s "Unknown issue to be reviewed (Group $group_name $group_id)" -r "openqa-label-known-issues <$from_email>" "$group_mailto"
```

kalikiana: From the line that was not meant to duplicate the existing variable 😅

okurz: I guess we can also make it even shorter with just `Unknown issue`

kalikiana: I'm going for `Unreviewed issue (Group 11 FooBar)`

_common
```diff
@@ -116,7 +116,8 @@ handle_unknown() {
         group_mailto=$(echo "$group_description" | perl -n -wE'm/.*MAILTO: (\S+).*/ and say $1' | head -1)
         group_mailto=${group_mailto:-$notification_address}
         if [[ -n "$group_mailto" ]]; then
-            echo "$header"$'\n'"Reason: $reason"$'\n'"$excerpt" | mailx -s "Unknown issue to be reviewed (Group $group_id)" -r "openqa-label-known-issues <$from_email>" "$group_mailto"
+            group_name=$(echo "$group_data" | runjq -r '.[0].name' ) || true
```
perlpunk, line 119: I know it's not checked by shellcheck, but I would add the new variable to line 103 (`local ...`)

okurz: And please see my earlier suggestion. I don't think we should ignore an error trying to parse the group name. We do that for the description which can be non-existant but we shouldn't for the name.

okurz (changes requested):

okurz (approved):

PR #164, opened 2022-05-25
Add dependency Text::Markdown

perlpunk:
This will be needed for the HTML email generation

Issue: https://progress.opensuse.org/issues/111215

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -8,6 +8,7 @@ requires 'Data::Dumper';
 requires 'FindBin';
 requires 'Getopt::Long';
 requires 'Mojo::File';
+requires 'Text::Markdown';
 requires 'YAML::PP';
 
 on 'test' => sub {
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -26,6 +26,7 @@ main_requires:
   perl(Getopt::Long):
   perl(FindBin):
   perl(Mojo::File):
+  perl(Text::Markdown):
   perl(YAML::PP):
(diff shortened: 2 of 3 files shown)


okurz (commented):
hm, don't we need the according update in the spec file then?

perlpunk:
Right... added now.
That actually reveals that we have no test here to check that.

perlpunk:
I added a test

xt/01-make-update-deps.t
```diff
@@ -0,0 +1,28 @@
...
+chdir $build_dir;
+my @out = qx{$make_cmd};
+my $rc = $?;
+die "Could not run $make_cmd: rc=$rc, out: @out" if $rc;
+
+my @status = grep { not m/^\?/ } qx{git -C "$Bin/.." status --porcelain};
+ok(!@status, "No changed files after '$make_cmd'") or diag @status;
+
+done_testing;
+
```
okurz, line 28: That's a copy-paste from os-autoinst. I would favor to not duplicate. How about putting the test into os-autoinst-common and call it from there? As alternative we could symlink from external into xt/

okurz (changes requested):

xt/01-make-update-deps.t
perlpunk: ok, but I will split it out of this PR so we can make actual progress.

Martchus (approved):

okurz (approved):

PR #170, opened 2022-06-24
openqa-investigate: Provide support for multi-machine scenarios

Martchus:
Latest commit: Sync investigation of parallel clusters via openQA comment

* Instead of only considering parallel parents, just do the investigation
  for any job with parallel dependencies
    * Avoid having to run the investigation script for all job results
    * Sync via an openQA comment instead (to avoid running the same
      investigation twice; abort if a concurrent job already does the
      investigation of the cluster)
    * Use `--max-depth 0` to clone all jobs in the parallel cluster,
      regardless whether we're starting from a parallel parent or child
        * Has no effect on other dependency types since we're
            * using `--skip-chained-deps` anyways
            * *not* using `--clone-children`
            * still excluding directly chained dependencies
* Write an investigation comment on the job we're actually investigating
  and on the first job in the cluster (for the synchronization)
* See https://progress.opensuse.org/issues/95783#note-58

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -42,7 +42,7 @@ clone() {
 
     name="$(echo "$job_data" | runjq -r '.job.test'):investigate$name_suffix" || return $?
     base_prio=$(echo "$job_data" | runjq -r '.job.priority') || return $?
-    clone_settings=("TEST=$name" '_GROUP_ID=0' 'BUILD=')
+    clone_settings=("TEST+=:investigate$name_suffix" '_GROUP_ID=0' 'BUILD=')
     if [[ $refspec ]]; then
         casedir=$(echo "$job_data" | runjq -r '.job.settings.CASEDIR') || return $?
         [[ $casedir == null ]] && casedir=''
@@ -242,7 +242,7 @@ main() {
         fi
     fi
     set -u
-    clone_call="${clone_call:-"$client_prefix openqa-clone-job --skip-chained-deps --max-depth 0 --within-instance"}"
+    clone_call="${clone_call:-"$client_prefix openqa-clone-job --skip-chained-deps --max-depth 0 --parental-inheritance --within-instance"}"
     error_count=0
     # shellcheck disable=SC2013
     for i in $(cat - | sed 's/ .*$//'); do
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -26,7 +26,7 @@ curl_args=(--user-agent "openqa-investigate")
 echoerr() { echo "$@" >&2; }
 
 clone() {
-    local origin id name_suffix refspec job_data unsupported_cluster_jobs name base_prio clone_settings casedir repo out url clone_id
+    local origin id name_suffix refspec job_data unsupported_cluster_jobs parallel_parents pending_cluster_jobs name base_prio clone_settings casedir repo out url clone_id
     origin=${1:?"Need 'origin'"}
     id=${2:?"Need 'id'"}
     name_suffix=${3+":$3"}
@@ -34,9 +34,19 @@ clone() {
     job_data=$(openqa-cli "${client_args[@]}" --json jobs/"$id")
     # shellcheck disable=SC2181
     [[ $? != 0 ]] && echoerr "unable to query job data for $id: $job_data" && return 1
-    unsupported_cluster_jobs=$(echo "$job_data" | runjq -r '(.job.children["Parallel"] | length) + (.job.parents["Parallel"] | length) + (.job.children["Directly chained"] | length) + (.job.parents["Directly chained"] | length)') || return $?
+
+    # fail on jobs with directly chained dependencies (not supported)
+    unsupported_cluster_jobs=$(echo "$job_data" | runjq -r '(.job.children["Directly chained"] | length) + (.job.parents["Directly chained"] | length)') || return $?
     [[ $unsupported_cluster_jobs != 0 ]] \
-        && echoerr "unable to clone job $id: it is part of a parallel or directly chained cluster (not supported)" && return 2
+        && echoerr "Unable to clone job $id: it is part of a directly chained cluster (not supported)" && return 2
+
+    # fail on parallel children (not supported)
+    # note: We already skip the investigation for parallel children to handle it together with the parent. However, the last good job
+    #       of the parallel parent might be a parallel child so we still need to check whether we have a parallel child.
+    parallel_parents=$(echo "$job_data" | runjq -r '(.job.parents["Parallel"] | length)') || return $?
+    [[ $parallel_parents != 0 ]] \
+        && echoerr "Unable to clone job $id: it is a parallel child; the structure of the parallel cluster might have changed since the last good (not supported)" && return 2
+
     name="$(echo "$job_data" | runjq -r '.job.test'):investigate$name_suffix" || return $?
     base_prio=$(echo "$job_data" | runjq -r '.job.priority') || return $?
     clone_settings=("TEST=$name" '_GROUP_ID=0' 'BUILD=')
@@ -119,6 +129,37 @@ trigger_jobs() {
     fi
 }
 
+check_dependencies_to_determine_whether_to_skip_or_postpone() {
+    local id=$1 job_data=$2 parallel_parents dependency_data pending_cluster_jobs cluster_jobs relevant_cluster_results
+
+    # skip parallel children (will be handled together with parallel parent)
+    # note: For this to work, this script needs to be executed for all job results (so we won't miss the investigation in case the parallel parent has passed).
+    parallel_parents=$(echo "$job_data" | runjq -r '(.job.parents["Parallel"] | length)') || return $?
+    [[ $parallel_parents != 0 ]] \
+        && echoerr "Skipping investigation of job $id: Investigation will be done on the parallel parent" && return 0
+
+    # postpone if not all dependencies are done/cancelled
+    # note: This "AJAX" route is normally used to render the dependencies tab in the web UI.
+    dependency_data=$(openqa-cli "${client_args[@]}" --apibase '' --json tests/"$id"/dependencies_ajax)
+    pending_cluster_jobs=$(echo "$dependency_data" | runjq -r '[.nodes[] | select(.state != "done" and .state != "cancelled")] | length') || return $?
+    [[ $pending_cluster_jobs != 0 ]] \
+        && echoerr "Postponing to investigate job $id: waiting until pending dependencies have finished" && return 142
+
+    # determine number of jobs in cluster with relevant result
+    # note: The "$id, …" is to always also consider the job itself (even if it is just a single job and therefore not in a cluster).
+    cluster_jobs=$(echo "$dependency_data" | runjq -r "[$id, [.cluster[] | select(contains([$id]))]] | flatten | unique") || return $?
+    # shellcheck disable=SC2016
+    relevant_cluster_results=$(echo "$dependency_data" | runjq --argjson cluster_jobs "$cluster_jobs" -r '[.nodes[] | select([.id] | inside($cluster_jobs)) | .result | select(inside("failed", "incomplete", "timeout_exceeded"))] | length') || return $?
+
+    # skip investigation if no jobs with relevant results are part of the cluster
+    # note: We need to run the investigate script for all jobs (see note from before) and therefore rely on checking the job/cluster result here.
+    [[ $relevant_cluster_results == 0 ]] \
+        && echoerr "Skip investigation of job $id: job is not failed/incomplete/timeout_exceeded, same counts for whole job cluster" && return 0
+
+    # do not skip the job
+    return 255
+}
+
 # crosscheck
 # 1. current job/build + current test -> check if reproducible/sporadic
 # 2. current job/build + last good test (+ last good needles) -> check for
@@ -151,6 +192,10 @@ investigate() {
         echoerr "Job already has a clone, skipping investigation. Use the env variable 'force=true' to trigger investigation jobs"
         return 0
     fi
+
+    check_dependencies_to_determine_whether_to_skip_or_postpone "$id" "$job_data"; rc=$?
+    [[ $rc != 255 ]] && return $rc
+
     [[ "$old_name" =~ $exclude_name_regex ]] && echo "Job name '$old_name' matches \$exclude_name_regex '$exclude_name_regex', skipping investigation" && return 0
     group="$(echo "$job_data" | runjq -r '.job.parent_group + " / " + .job.group')" || return $?
     [[ "$group" = " / " ]] && [[ "$exclude_no_group" = "true" ]] && echo "Job w/o job group, \$exclude_no_group is set, skipping investigation" && return 0
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -149,7 +149,7 @@ check_dependencies_to_determine_whether_to_skip_or_postpone() {
     # note: The "$id, …" is to always also consider the job itself (even if it is just a single job and therefore not in a cluster).
     cluster_jobs=$(echo "$dependency_data" | runjq -r "[$id, [.cluster[] | select(contains([$id]))]] | flatten | unique") || return $?
     # shellcheck disable=SC2016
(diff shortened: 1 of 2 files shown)


openqa-investigate
```diff
@@ -151,6 +161,29 @@ investigate() {
         echoerr "Job already has a clone, skipping investigation. Use the env variable 'force=true' to trigger investigation jobs"
         return 0
     fi
+
+    # skip parallel children (will be handled together with parallel parent)
+    # note: For this to work, this script needs to be executed for all job results (so we won't miss the investigation in case the parallel parent has passed).
+    parallel_parents=$(echo "$job_data" | runjq -r '(.job.parents["Parallel"] | length)') || return $?
+    [[ $parallel_parents != 0 ]] \
+        && echoerr "Skipping investigation of job $id: it will be handled together with its parallel parent" && return 0
```
okurz, line 169:
```suggestion
        && echoerr "Skipping investigation of job $id: Investigation will be done on the parallel parent" && return 0
```

to avoid the potentially ambiguous "it"-reference and making it a tad shorter

openqa-investigate
```diff
@@ -151,6 +161,29 @@ investigate() {
...
+
+    # postpone if not all dependencies are done/cancelled
+    dependency_data=$(openqa-cli "${client_args[@]}" --apibase '' --json tests/"$id"/dependencies_ajax)
```
okurz, line 172: if we need this hack how about extending the openQA API as needed?

openqa-investigate
```diff
@@ -151,6 +161,29 @@ investigate() {
...
+    # determine number of jobs in cluster with relevant result
+    # note: The ", $id" is to always also consider the job itself (even if it is just a single job and therefore not in a cluster).
+    cluster_jobs=$(echo "$dependency_data" | runjq -r "[.cluster[] | select(contains([$id])), $id] | flatten | unique") || return $?
+    relevant_cluster_results=$(echo "$dependency_data" | runjq --argjson cluster_jobs "$cluster_jobs" -r '[.nodes[] | select([.id] | inside($cluster_jobs)) | .result | select(inside("failed", "incomplete", "timeout_exceeded"))] | length') || return $?
+
+    # skip investigation if no jobs with relevant results are part of the cluster
+    # note: We need to run the investigate script for all jobs (see note from before) and therefore rely on checking the job/cluster result here.
+    [[ $relevant_cluster_results == 0 ]] \
+        && echoerr "Skip investigation of job $id: job is not failed/incomplete/timeout_exceeded, same counts for whole job cluster" && return 142
+
```
okurz, line 186: Please extract some code as you are making the "investigate" function too big

openqa-investigate
```diff
@@ -26,17 +26,27 @@ curl_args=(--user-agent "openqa-investigate")
...
-        && echoerr "unable to clone job $id: it is part of a parallel or directly chained cluster (not supported)" && return 2
+        && echoerr "Unable to clone job $id: it is part of a directly chained cluster (not supported)" && return 2
+
+    # fail on parallel children (not supported)
+    # note: We already skip the investigation for parallel children to handle it together with the parent. However, the last good job
+    #       of the parallel parent might be a parallel child so we still need to check whether we have a parallel child.
+    parallel_parents=$(echo "$job_data" | runjq -r '(.job.parents["Parallel"] | length)') || return $?
+    [[ $parallel_parents != 0 ]] \
+        && echoerr "Unable to clone job $id: it is a parallel child; the structure of the parallel cluster might have changed since the last good (not supported)" && return 2
+
```
okurz, line 49: please extract some code as the function "clone" becomes too big

okurz (changes requested):

openqa-investigate
Martchus: I could move the route into the API namespace. For now I would refrain from it (since everything is still a bit experimental and we don't need to commit to a certain API for "real external" users yet). Of course we need to take care that we don't break this route without also adjusting this script but I suppose that this mistake is quite unlikely to happen.

openqa-investigate
Martchus: I suppose for extending tests that will be necessary anyways :-)

openqa-investigate
okurz: Ok, leave it as is for now

openqa-investigate
```diff
@@ -119,6 +129,37 @@ trigger_jobs() {
...
+    dependency_data=$(openqa-cli "${client_args[@]}" --apibase '' --json tests/"$id"/dependencies_ajax)
+    pending_cluster_jobs=$(echo "$dependency_data" | runjq -r '[.nodes[] | select(.state != "done" and .state != "cancelled")] | length') || return $?
+    [[ $pending_cluster_jobs != 0 ]] \
+        && echoerr "Postponing to investigate job $id: waiting until pending dependencies have finished" && return 142
+
+    # determine number of jobs in cluster with relevant result
+    # note: The "$id, …" is to always also consider the job itself (even if it is just a single job and therefore not in a cluster).
+    cluster_jobs=$(echo "$dependency_data" | runjq -r "[$id, [.cluster[] | select(contains([$id]))]] | flatten | unique") || return $?
+    # shellcheck disable=SC2016
+    relevant_cluster_results=$(echo "$dependency_data" | runjq --argjson cluster_jobs "$cluster_jobs" -r '[.nodes[] | select([.id] | inside($cluster_jobs)) | .result | select(inside("failed", "incomplete", "timeout_exceeded"))] | length') || return $?
```
Martchus, line 152: Looks like this is actually wrong and should be

```suggestion
    relevant_cluster_results=$(echo "$dependency_data" | runjq --argjson cluster_jobs "$cluster_jobs" -r '[.nodes[] | select([.id] | inside($cluster_jobs)) | .result | select(inside("failed"))] | length') || return $?
```
because for incomplete/timeout_exceeded we only run `openqa-label-known-issues-hook`.

openqa-investigate
```diff
@@ -119,6 +129,37 @@ trigger_jobs() {
     fi
 }
 
+check_dependencies_to_determine_whether_to_skip_or_postpone() {
+    local id=$1 job_data=$2 parallel_parents dependency_data pending_cluster_jobs cluster_jobs relevant_cluster_results
+
+    # skip parallel children (will be handled together with parallel parent)
+    # note: For this to work, this script needs to be executed for all job results (so we won't miss the investigation in case the parallel parent has passed).
```
Martchus, line 136: Looks like that's currently not possible unless one would list all possible job results in the config (to have a line like `job_done_hook_enable_failed = 1` for every job result).

openqa-investigate
okurz: correct

okurz (approved):

okurz (approved):

Martchus:
Pushed a new commit. It is still WIP and I need to adjust tests but feedback is appreciated. See the commit message for details.

openqa-investigate
```diff
@@ -145,18 +132,24 @@ check_dependencies_to_determine_whether_to_skip_or_postpone() {
...
 
-    # do not skip the job
+    comment_id=$("${client_call[@]}" -X POST jobs/"$first_cluster_job_id"/comments text="Starting investigation for $id" | runjq -r '.id') || return $?
+    first_comment_id=$("${client_call[@]}" -X GET jobs/"$first_cluster_job_id"/comments | runjq -r '[.[] | select(.text | contains("investigation"))] | sort_by(.id) | first | .id') || return $?
+
+    # delete comment again in case a concurrent job could start the investigation before us
+    if [[ $comment_id != $first_comment_id ]]; then
+        echoerr "Skip investigation of job $id: job cluster is already being investigated, see comment on $first_cluster_job_id"
+        "${client_call[@]}" -X DELETE jobs/"$first_cluster_job_id"/comments/"$comment_id" && return 0
+    fi
```
okurz, line 150: this looks very clean to me, good!

openqa-investigate
```diff
@@ -145,18 +132,24 @@ check_dependencies_to_determine_whether_to_skip_or_postpone() {
...
+sync_via_investigation_commennt() {
+    local id=$1 first_cluster_job_id=$2
 
-    # do not skip the job
+    comment_id=$("${client_call[@]}" -X POST jobs/"$first_cluster_job_id"/comments text="Starting investigation for $id" | runjq -r '.id') || return $?
+    first_comment_id=$("${client_call[@]}" -X GET jobs/"$first_cluster_job_id"/comments | runjq -r '[.[] | select(.text | contains("investigation"))] | sort_by(.id) | first | .id') || return $?
+
+    # delete comment again in case a concurrent job could start the investigation before us
+    if [[ $comment_id != $first_comment_id ]]; then
+        echoerr "Skip investigation of job $id: job cluster is already being investigated, see comment on $first_cluster_job_id"
```
okurz, line 148: s/Skip/Skipping/

openqa-investigate
```diff
@@ -205,13 +202,22 @@ investigate() {
...
 
+    # sync by writing initial investigation comment (edited later)
+    comment_id=$(sync_via_investigation_commennt "$id" "$first_cluster_job_id"); rc=$?
+    [[ $rc != 255 ]] && return $rc
+
     out=$(trigger_jobs "$id" "${@:2}")
     $verbose && echo "$0, id: '$id', out: '$out'"
     [[ $out ]] || return 0
-    comment="Automatic investigation jobs:
+    comment="Automatic investigation jobs for $id:
```
okurz, line 212: `s/for $id/for job $id/` to make it very obvious what the number means

okurz (changes requested):

Martchus:
It works in production on a simple case, see https://openqa.opensuse.org/tests/2456131#comments. It also skips it correctly on the 2nd run:
```
$ echo 'https://openqa.opensuse.org/tests/2456131' | env exclude_group_regex='foobar'  ./openqa-investigate 
{"id":285792}
$ echo 'https://openqa.opensuse.org/tests/2456131' | env exclude_group_regex='foobar'  ./openqa-investigate 
Skipping investigation of job 2456131: job cluster is already being investigated, see comment on job 2456131
```

It also worked on a failed parallel child, see https://openqa.opensuse.org/tests/2456116#comments. That restarted the cluster correctly (no chained parents, just the cluster as expected). It also created the comment on the first job in the cluster and edited it later, see https://openqa.opensuse.org/tests/2456115#comments. A second run was also skipped as expected, as well as a run on the parent:

```
$ echo 'https://openqa.opensuse.org/tests/2456116' | env dry_run= exclude_group_regex='foobar'  ./openqa-investigate 
Cloning parents of opensuse-Tumbleweed-DVD-aarch64-Build20220706-salt-minion@aarch64
Cloning children of opensuse-Tumbleweed-DVD-aarch64-Build20220706-salt-master@aarch64
Cloning parents of opensuse-Tumbleweed-DVD-aarch64-Build20220706-salt-minion@aarch64
Cloning children of opensuse-Tumbleweed-DVD-aarch64-Build20220706-salt-master@aarch64
Cloning parents of opensuse-Tumbleweed-DVD-aarch64-Build20220628-salt-minion@aarch64
Cloning children of opensuse-Tumbleweed-DVD-aarch64-Build20220628-salt-master@aarch64
Cloning parents of opensuse-Tumbleweed-DVD-aarch64-Build20220628-salt-minion@aarch64
Cloning children of opensuse-Tumbleweed-DVD-aarch64-Build20220628-salt-master@aarch64
{"id":285794}
{"id":285795}
$ echo 'https://openqa.opensuse.org/tests/2456116' | env dry_run= exclude_group_regex='foobar'  ./openqa-investigate 
Skipping investigation of job 2456116: job cluster is already being investigated, see comment on job 2456115
$ echo 'https://openqa.opensuse.org/tests/2456115' | env dry_run= exclude_group_regex='foobar'  ./openqa-investigate 
Skipping investigation of job 2456115: job cluster is already being investigated, see comment on job 2456115
```

(I've cancelled all investigation jobs again manually.)

okurz (changes requested):
> It works in production on a simple case, see https://openqa.opensuse.org/tests/2456131#comments. It also skips it correctly on the 2nd run:
> 
> ```
> $ echo 'https://openqa.opensuse.org/tests/2456131' | env exclude_group_regex='foobar'  ./openqa-investigate 
> {"id":285792}
> $ echo 'https://openqa.opensuse.org/tests/2456131' | env exclude_group_regex='foobar'  ./openqa-investigate 
> Skipping investigation of job 2456131: job cluster is already being investigated, see comment on job 2456131
> ```
> 
> It also worked on a failed parallel child, see https://openqa.opensuse.org/tests/2456116#comments. That restarted the cluster correctly (no chained parents, just the cluster as expected). It also created the comment on the first job in the cluster and edited it later, see https://openqa.opensuse.org/tests/2456115#comments.

Looking at https://openqa.opensuse.org/tests/2456116#comment-285795 the first job pair is:
* salt-minion:investigate:retry: https://openqa.opensuse.org/t2457333 https://openqa.opensuse.org/t2457334
Out of those both the *second* is correctly called "opensuse-Tumbleweed-DVD-aarch64-salt-minion:investigate:retry@aarch64" and not within any job group, but the *first* is still with the original name "opensuse-Tumbleweed-DVD-aarch64-Build20220706-salt-master@aarch64" and within the original job group and build. That should be avoided as we don't want to pollute production builds.

Martchus:
I suppose I need to enable parental inheritance for that. However, then we have the same problem as before with the restart approach - we cannot have job-specific settings. For `TEST` we could use `+=` but for and if `OPENQA_INVESTIGATE_ORIGIN` is always the same we can likely live with that. However, for `CASEDIR` that doesn't make necessarily sense.

Martchus:
I suppose we can also live with using the same `CASEDIR` everywhere. So I've added another commit to apply the job settings for the whole cluster, see its commit message.

test/02-investigate.t
```diff
@@ -61,6 +86,51 @@ is "$rc" 0 'success regardless of actually triggered jobs'
...
 like "$out" "exclude_no_group is set, skipping investigation"
+
+rc=0
+out=$(investigate 30 2>&1) || rc=$?
+is "$rc" 142 'investigation postponed because other job in cluster is not done'
+like "$out" "Postponing to investigate job 30: waiting until pending dependencies have finished"
+
+rc=0
+out=$(force=true investigate 31 2>&1) || rc=$?
+is "$rc" 0 'success when job is skipped (because of exclude_no_group annd job w/o group)'
```
okurz, line 100: s/annd/and/

test/02-investigate.t
```diff
@@ -61,6 +86,51 @@ is "$rc" 0 'success regardless of actually triggered jobs'
...
+# delete certain files used to trace whether API calls happened
+for trace_file in comment_1234_updated comment_for_job_31_created comment_1234_deleted; do
+    [[ -f $trace_file ]] && unlink "$trace_file"
+done
+
+# test finalizing investigation comment when no investigation jobs were needed
+rc=0
+out=$(force=true finalize_investigation_comment 31 30 1234 '' 2>&1) || rc=$?
+is "$rc" 0 'success if no investigation jobs needed to be created after all'
+[[ -f comment_1234_deleted ]] && comment_1234_deleted=1 || comment_1234_deleted=0
```
okurz, line 124: this could be

```suggestion
[[ ! -f comment_1234_deleted ]]; comment_1234_deleted=$?
```

okurz (changes requested):

test/02-investigate.t
Martchus: It somehow doesn't work; if the condition is false this is considered an error and the test stops. Only using `set +e` before would make it work.

Putting
```
some-function() {
    return 255
}
another-function() {
    foo=$(some-function ); rc=$?
    [[ $rc != 255 ]] && return $rc
    return 0
}
another-function
```
there gives me the same problem.

But I don't really understand it because why is then
```
    dependency_data=$(query_dependency_data_or_postpone "$id" "$job_data"); rc=$?
    [[ $rc != 255 ]] && return $rc
```
working? It also works if I explicitly invoke `set -e` before. And I'm pretty sure that it works here because if it wouldn't work the test would fail and the script would always skip the investigation.

okurz: well, nevermind then.

okurz:
> I suppose we can also live with using the same `CASEDIR` everywhere. So I've added another commit to apply the job settings for the whole cluster, see its commit message.
> 
> I also moved the previously created jobs out of the group/build.

Can you trigger a new run on o3 with openqa-investigate and investigation jobs please?

Martchus:
I did, see the second comment on https://openqa.opensuse.org/tests/2456116#comment-286104. (To be able to trigger it I temporarily edited the first comment.)

okurz (approved):

PR #171, opened 2022-07-13
Fix postponing the investigation

Martchus:
* Ensure the `main()` function of `openqa-investigate` actually passes the
  return code of the `investigate()` call
    * Only returns the return code of the last `investigate()` call but
      that should be ok since we're only investigating one job at a time
      in `openqa-label-known-issues-and-investigate-hook`
* Add `set -e` to `openqa-label-known-issues-and-investigate-hook` for
  consistent behavior regardless whether the script is called via shebang
  or not
* Tested against my local instance to see whether
  `openqa-label-known-issues` and
  `openqa-label-known-issues-and-investigate-hook` will actually return 142
* See https://progress.opensuse.org/issues/95783

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -243,12 +243,12 @@ main() {
     fi
     set -u
     clone_call="${clone_call:-"$client_prefix openqa-clone-job --skip-chained-deps --max-depth 0 --parental-inheritance --within-instance"}"
-    error_count=0
+    local rc=0
     # shellcheck disable=SC2013
     for i in $(cat - | sed 's/ .*$//'); do
-        investigate "$i" "$@" || ((error_count++)) ||:
+        investigate "$i" "$@" || rc=$?
     done
-    exit "$error_count"
+    exit $rc
 }
 
 caller 0 >/dev/null || main "$@"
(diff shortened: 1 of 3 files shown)


openqa-investigate
```diff
@@ -243,12 +243,12 @@ main() {
     fi
     set -u
     clone_call="${clone_call:-"$client_prefix openqa-clone-job --skip-chained-deps --max-depth 0 --parental-inheritance --within-instance"}"
-    error_count=0
+    local rc=0
     # shellcheck disable=SC2013
     for i in $(cat - | sed 's/ .*$//'); do
-        investigate "$i" "$@" || ((error_count++)) ||:
+        investigate "$i" "$@" || rc=$?
```
okurz, line 249: why did we have an error count in the first place when you are resorting to the simpler solution now?

Martchus: Ask that @okurz from two years ago: https://github.com/os-autoinst/scripts/commit/60a8c4c80ec4ec8ca956a0782ea78d9ccd375d45

My simpler code has the caveat to only return the return code of the last investigate call. Maybe you wanted to avoid that. However, I don't think we use the count somewhere. Note that the count was broken anyways for more than 255 errors.

okurz: Well, the subject line "Prevent immediate failure if just a single job fails" already explains that, thank you @okurz from two years ago ;) So why should we exit immediately now on the first failure?

okurz: huh, weird. Ok then :)

okurz (approved):

PR #173, opened 2022-07-14
Postpone investigation jobs only if pending jobs are in same cluster

Martchus:
Before this change, investigation jobs are postponed if *any* dependency is
still pending. For instance, a pending chained sibling would cause the
investigation of a job to be postponed. That's not needed so this change
will postpone the investigation only if there are pending jobs in the same
parallel cluster.

See https://progress.opensuse.org/issues/95783#note-64

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -123,14 +123,16 @@ trigger_jobs() {
 }
 
 query_dependency_data_or_postpone() {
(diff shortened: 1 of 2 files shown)


Martchus:
You have already approved before I could share the screenshot :-)

Well, here's nevertheless an example of a job where the investigation was needlessly postponed because of a running chained sibling:

screenshot_20220714_134228

PR #174, opened 2022-08-02
Post similarity

Akemiiii:
This script scans the autoinst-log.txt in the testresults directory. Then it finds the error messages in the autoinst-log.txt. To avoid double computing, it saves some results in a JSON file.  Next, it uses Levenshtein to calculate the text distance. Finally, it posts the results as comments by OpenQA_Client. I have converted the .pkl file into a JSON file for readability. I also changed some codes to minimize the use of walking.

15/8/2021 updated:
I rewrote this script:
1. Remove commented-out codes and add logging to the terminal.
2. Use parser to set configurable parameters, including host.
3. Remove useless fields for posting comments.
4. Separate codes into several functions and the top level codes contain some function calls.
5. Use a function to remove duplicate codes.
6. Fix some bugs.

### DIFF
--- a/openqa-post-similarity
+++ b/openqa-post-similarity
@@ -1,5 +1,6 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 
+from curses.ascii import isdigit
 import os
 import Levenshtein
 from http import client
@@ -9,6 +10,8 @@ import argparse
 import logging
 from tqdm import tqdm
 import sys
+from pyecharts import options as opts
+from pyecharts.charts import Graph
 
 """
 This script scans all autoinst-log.txt files in the testresults directory searching
@@ -119,6 +122,79 @@ def post(server, number):
         client.openqa_request('POST', 'jobs/' + origin + '/comments', data)
 
 
+def dict_slice(adict, s, e):
+    keys = adict.keys()
+    d_slice = {}
+    for k in list(keys)[s:e]:
+        d_slice[k] = adict[k]
+    return d_slice
+
+
+def draw(logger, points, geometry, save_path):
+    resolution = geometry.split("x")
+    if len(resolution) != 2:
+        logger.warning("Wrong geometry format")
+        return
+    for item in resolution:
+        if not isdigit(item):
+            logger.warning("Wrong geometry format")
+            return
+
+    msg_number = {}
+    for err_id, msg in id_msg.items():
+        if msg in msg_number:
+            msg_number[msg] += 1
+        else:
+            msg_number[msg] = 1
+    cut = dict_slice(msg_number, 0, points)
+
+    nodes_data = []
+    for msg, freq in cut.items():
+        nodes_data.append(opts.GraphNode(name=msg, symbol_size=freq))
+    
+    links_data = []
+    max_distance = 0
+    min_distance = float("inf")
+    for key1 in cut.keys():
+        for key2 in cut.keys():
+            if key1 == key2:
+                continue
+            L = Levenshtein.distance(key1, key2)
+            if L > max_distance:
+                max_distance = L
+            if L < min_distance:
+                min_distance = L
+    for key1 in cut.keys():
+        for key2 in cut.keys():
+            if key1 == key2:
+                continue
+            L = Levenshtein.distance(key1, key2)
+            if L < (max_distance + min_distance) / 2:
+                links_data.append(opts.GraphLink(source=key1, target=key2, value=Levenshtein.distance(key1, key2)))
+    c = (
+        Graph(init_opts=opts.InitOpts(height=resolution[0]+"px", width=resolution[1]+"px"))
+        .add(
+            "",
+            nodes_data,
+            links_data,
+            repulsion=400,
+            edge_length=[10, 1000],
+            layout="force",
+            gravity=0.01,
+            edge_label=opts.LabelOpts(
+                is_show=False, position="middle", formatter="{b} : {c}"
+            ),
+            label_opts=opts.LabelOpts(
+                is_show=False,
+            ),
+        )
+        .set_global_opts(
+            title_opts=opts.TitleOpts(title="Levenshtein distance")
+        )
+        .render(save_path)
+    )
+        
+
 if __name__ == '__main__':
     help_message = "This script scans all autoinst-log.txt files in the testresults directory searching for error " \
                    "messages. Then it " \
@@ -128,8 +204,15 @@ if __name__ == '__main__':
     parser.add_argument("-s", "--server", default="http://127.0.0.1:9526", type=str, help="OpenQA server URL")
     parser.add_argument("-n", "--number", default=10, type=int, help="Number of similar errors")
     parser.add_argument("-d", "--dir", default="/var/lib/openqa/testresults/", type=str, help="Directory of OpenQA test results")
+    parser.add_argument("-p", "--post", action="store_true", help="Whether post similarity to openQA website")
+    parser.add_argument("-c", "--chart", required='--geometry' in sys.argv or '--save' in sys.argv, default=0, type=int, help="Number of points in chart (If the number is 0, the chart won't be drawn)")
+    parser.add_argument("--geometry", default="1920x1080", type=str, help="Chart resolution (e.g. 1920x1080)")
+    parser.add_argument("--save", default="./similarity.html", type=str, help="Path and name to save the chart (e.g. ./similarity.html)")
     args = parser.parse_args()
     logger = log_set()
     read_id_msg(logger=logger, testdir=args.dir)
     cal_distance(logger=logger, output=args.output, number=args.number)
-    post(server=args.server, number=args.number)
+    if args.post:
+        post(server=args.server, number=args.number)
+    if args.chart != 0:
+        draw(logger=logger, points=args.chart, geometry=args.geometry, save_path=args.save)
--- a/openqa-post-similarity
+++ b/openqa-post-similarity
@@ -53,7 +53,7 @@ def read_id_msg(logger, testdir):
         all_dirs = [testdir + item for item in all_dirs]
         logs = [item + '/autoinst-log.txt' for item in all_dirs]
 
-        for path in logs:
+        for path in tqdm(logs, desc='Processing testresults directory', unit='test'):
             read_errors(path)
         id_msg_json = json.dumps(id_msg, sort_keys=False, indent=4, separators=(',', ':'))
         with open("id_msg.json", "w") as f:
@@ -67,7 +67,7 @@ def read_id_msg(logger, testdir):
         all_dirs = os.listdir(testdir)
         checked_dirs = list(id_msg.keys())
         unchecked_dirs = [i for i in all_dirs if i not in checked_dirs]
-        for job_id in unchecked_dirs:
+        for job_id in tqdm(unchecked_dirs, desc='Processing testresults directory', unit='test'):
             path = testdir + job_id + "/autoinst-log.txt"
             read_errors(path)
         id_msg_json = json.dumps(id_msg, sort_keys=False, indent=4, separators=(',', ':'))
@@ -81,7 +81,7 @@ def cal_distance(logger, output, number):
     global result
     if output:
         f = open("distance_result.txt", "w")
-    for index, (key1, value1) in enumerate(id_msg.items()):
+    for index, (key1, value1) in enumerate(tqdm(id_msg.items(), desc='Calculating message distance', unit="error")):
         calculate = {}
         for key2, value2 in id_msg.items():
             if key1 == key2:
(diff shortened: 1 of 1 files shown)


openqa-post-similarity
```diff
@@ -0,0 +1,111 @@
...
+        f.write("Job ID: " + calculate_sorted[i][0] + "\n")
+        f.write(id_msg[calculate_sorted[i][0]] + "\n")
+        matched_results.append(calculate_sorted[i][0])
+    f.write("\n")
+    result[key1] = matched_results
+
+# print(result)
+
+# Post the results in comments by OpenQA_Client
+client = OpenQA_Client(server="http://127.0.0.1:9526")
```
Martchus, line 99: Hard-coding the server's address isn't very nice.

openqa-post-similarity
```diff
@@ -0,0 +1,111 @@
...
+for origin, matched in result.items():
+    data = {'bugrefs': [], 'created': time.strftime("%Y-%m-%d %H:%M:%S", time.localtime()) + " +0000"}
```
Martchus, line 101: It should not be necessary to pass created/updated timestamps if you just want to use the current time. Note that this code only works correctly if the local time is GMT/UTC±00:00. I assume you put this hack here to have openQA display times in your local timezone but that's not generally wanted; openQA is generally supposed to use display timestamps using GMT/UTC±00:00. So please remove that hack.

openqa-post-similarity
```diff
@@ -0,0 +1,111 @@
...
+    text = "Top 10 similar failures:\r\n"
+    rendered_markdown = "<p>Top 10 similar failures:</p>\n"
+    for job_id in matched:
+        text += "[" + job_id + "](https://openqa.opensuse.org/tests/" + job_id + ")\r\n"
+        rendered_markdown += '<p><a href="https://openqa.opensuse.org/tests/' + job_id + '">' + job_id + '</a>\n'
+    data['renderedMarkdown'] = rendered_markdown
+    data['text'] = text
+    data['updated'] = data['created']
```
Martchus, line 109: This should be removed as well (for reasons stated in the other comment).

openqa-post-similarity
```diff
@@ -0,0 +1,111 @@
...
+from http import client
+from openqa_client.client import OpenQA_Client
+import time
+import json
+
+"""
+This script scans all the autoinst-log.txt in the testresults directory.
+Then find the error messages in the autoinst-log.txt.
+Use Levenshtein to calculate the text distance.
+Finally, post the results as comments by OpenQA_Client.
```
Martchus, line 14: I'd shorten and improve the wording a little bit.

```suggestion
This script scans all autoinst-log.txt files in the testresults directory searching
for error messages. Then it computes the Levenshtein distance and posts the
results as openQA comments.
```

Note that it would likely make more sense to show this text when the script is invoked via `--help` (and then of course anybody opening the file first can still read it as string literal).

openqa-post-similarity
```diff
@@ -0,0 +1,111 @@
...
+# Post the results in comments by OpenQA_Client
+client = OpenQA_Client(server="http://127.0.0.1:9526")
+for origin, matched in result.items():
+    data = {'bugrefs': [], 'created': time.strftime("%Y-%m-%d %H:%M:%S", time.localtime()) + " +0000"}
+    text = "Top 10 similar failures:\r\n"
+    rendered_markdown = "<p>Top 10 similar failures:</p>\n"
+    for job_id in matched:
+        text += "[" + job_id + "](https://openqa.opensuse.org/tests/" + job_id + ")\r\n"
+        rendered_markdown += '<p><a href="https://openqa.opensuse.org/tests/' + job_id + '">' + job_id + '</a>\n'
+    data['renderedMarkdown'] = rendered_markdown
```
Martchus, line 107: Passing the rendered markdown explicitly should not be required. In fact, openQA will simply ignore it.

Martchus (changes requested):

openqa-post-similarity
```diff
@@ -0,0 +1,111 @@
...
+This script scans all the autoinst-log.txt in the testresults directory.
+Then find the error messages in the autoinst-log.txt.
+Use Levenshtein to calculate the text distance.
+Finally, post the results as comments by OpenQA_Client.
+"""
+
+# gather all the autoinst-log.txt
+BASE_DIR = os.getenv("OPENQA_BASEDIR")
+logs = []
+for root, dirs, files in os.walk(BASE_DIR + "//openqa//testresults"):
```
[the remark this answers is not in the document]
Akemiiii: I used escape characters. Do you mean using r"" is better?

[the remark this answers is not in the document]
Martchus: Escaping on what level? I haven't used `os.walk` for ages but the string-literal definitely does not need escaping. You're confusing `/` with `\`.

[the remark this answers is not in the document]
Akemiiii: You are right. Thank you.

Martchus:
We usually only merge PR branches on master but not the other way around. So please rebase your branch against master (instead of merging it with master). There shouldn't be a merge commit on your branch.

Akemiiii:
> We usually only merge PR branches on master but not the other way around. So please rebase your branch against master (instead of merging it with master). There shouldn't be a merge commit on your branch.

Do you mean I should combine my commits by rebase？

Martchus:
> Do you mean I should combine my commits by rebase？

"combining" (which is actually called "squashing" in Git) via rebase is something we can do as well via GitHub. So you don't necessarily need to squash your commits. That's the option mentioned in my last paragraph.

If you want to keep your commits separate, then you'll have to improve them. That's what the first few paragraphs of my last comment are about. And yes, getting rid of the merge commit means rebasing against master (but not necessarily squashing all commits).

Akemiiii:
> If you want to keep your commits separate, then you'll have to improve them.

So every commit should represent a complete thing?

Martchus:
Not necassariy. However, the commit messages should be clear in the context of this repository. And e.g. "Add help text" is not very clear as it is not clear to which script it relates. And again, there shouldn't be a merge commit on the branch.

openqa-post-similarity
```diff
@@ -0,0 +1,126 @@
...
+logger.setLevel(logging.INFO)
+
+help_message = "This script scans all autoinst-log.txt files in the testresults directory searching for error messages. Then it " \
+    "computes the Levenshtein distance and posts the results as openQA comments. "
+parser = argparse.ArgumentParser(description=help_message)
+args = parser.parse_args()
+
+# gather all the autoinst-log.txt
+logger.info("Get environment variables")
+BASE_DIR = os.getenv("OPENQA_BASEDIR")
```
perlpunk, line 28: [`getenv`](https://docs.python.org/3/library/os.html#os.getenv) supports a default value, so I suggest:
```suggestion
BASE_DIR = os.getenv("OPENQA_BASEDIR", "/var/lib")
```

openqa-post-similarity
```diff
@@ -0,0 +1,126 @@
...
+            anchor = 0
+            begin = text.find("Test died", anchor)
+            while begin != -1:
+                anchor = begin + 1
+                end = text.find("\n", begin)
+                id_msg[path.split("\\")[1]] = text[begin:end]
+                begin = text.find("Test died", anchor)
+    id_msg_json = json.dumps(id_msg, sort_keys=False, indent=4, separators=(',', ':'))
+    with open("id_msg.json", "w") as f:
+        f.write(id_msg_json)
```
perlpunk, line 81: This code looks like a copy of the above code, so it should be extracted to a function.

openqa-post-similarity
```diff
@@ -0,0 +1,126 @@
...
+    data = {'bugrefs': [], 'created': time.strftime("%Y-%m-%d %H:%M:%S", time.localtime()) + " +0000"}
+    text = "Top 10 similar failures:\r\n"
+    rendered_markdown = "<p>Top 10 similar failures:</p>\n"
+    for job_id in matched:
+        text += "[" + job_id + "](https://openqa.opensuse.org/tests/" + job_id + ")\r\n"
+        rendered_markdown += '<p><a href="https://openqa.opensuse.org/tests/' + job_id + '">' + job_id + '</a>\n'
+    data['renderedMarkdown'] = rendered_markdown
+    data['text'] = text
+    data['updated'] = data['created']
+    data['userName'] = 'Demo'
```
perlpunk, line 125: Also this should be removed. The user of the comment will be determined via authentication through the api secret. It's not possible to set a user explicitly when posting a comment.

perlpunk (commented):
Additionally to our comments I think it would be good to have all the code in separate functions, so ideally the only toplevel code should be a function call.

openqa-post-similarity
Akemiiii: I missed your comments before. Thank you.

openqa-post-similarity
Akemiiii: I am not sure about how to use api key/secret. I should read this (https://github.com/os-autoinst/openQA-python-client/blob/2aeba7a5868f86a4150b32b65fed17a976dd6f1d/src/openqa_client/client.py#L142) as an example of authentication. Should I read api key/secret from /etc/openqa/client.conf or ~/.config/openqa/client.conf or by user input? Thank you.

Martchus: I think reading credentials from those config files makes most sense. Of course, if you add CLI argument parsing it wouldn't hurt to allow adding CLI args for it (like e.g. `openqa-clone-job` has).

Akemiiii: I see. Thank you.

openqa-post-similarity
```diff
@@ -0,0 +1,138 @@
...
+            calculate[key2] = Levenshtein.distance(value1, value2)
+        calculate_sorted = sorted(calculate.items(), key=lambda x: x[1], reverse=False)
+        if output:
+            f.write("Index: " + str(index) + "\n")
+            f.write("Original error message:\n")
+            f.write("Job ID: " + key1 + "\n")
+            f.write(value1 + "\n")
+            f.write("matched error message (top 10):\n")
+        matched_results = []
+        for i in range(10):
```
okurz, line 103: I suggest to make this value "10" also a configurable parameter passed from the argument parser

Akemiiii: Modified. Thank you.

openqa-post-similarity
```diff
@@ -0,0 +1,139 @@
...
+This script scans all autoinst-log.txt files in the testresults directory searching
+for error messages. Then it computes the Levenshtein distance and posts the
+results as openQA comments.
+"""
+DEBUG = True
+BASE_DIR = os.getenv("OPENQA_BASEDIR", "/var/lib/")
+if DEBUG:
+    TEST_DIR = BASE_DIR + "/repos/scripts/testresults/"
+else:
+    TEST_DIR = BASE_DIR + "/openqa/testresults/"
```
okurz, line 23: As you already use argparse I suggest a command line option for this as well

Akemiiii: I have added an argument and set /var/lib/openqa/testresults/ as the default value

okurz (approved):

Martchus (approved):

Martchus:
Adding the not-ready label as we might want to squash commits because there's still a merge commit on this branch and commit messages are quite unspecific in the context of this repository (as it is not clear to which script they belong).

Akemiiii:
Do you mean I can use "git rebase -interactive" to squash commits after improving commits?

Martchus:
No, you use "git rebase -interactive" to improve commits. And if you don't then we can squash commits automatically on merge via GitHub.

Martchus (changes requested):
Looks like you rebased against a wrong commit. The PR branch should not contain unrelated commits.

Martchus (approved):
Commits look good now.

openqa-post-similarity
```diff
@@ -0,0 +1,135 @@
...
+            id_msg = json.load(f)
+        all_dirs = os.listdir(testdir)
+        checked_dirs = list(id_msg.keys())
+        unchecked_dirs = [i for i in all_dirs if i not in checked_dirs]
+        for job_id in unchecked_dirs:
+            path = testdir + job_id + "/autoinst-log.txt"
+            read_errors(path)
+        id_msg_json = json.dumps(id_msg, sort_keys=False, indent=4, separators=(',', ':'))
+        with open("id_msg.json", "w") as f:
+            f.write(id_msg_json)
```
perlpunk, line 78: I still think there is lots of duplicated code here in the if and else block. The last 3 lines are exactly the same, and it feels like the rest of the code could also be shared if it was rewritten.
But I think it's ok for now, as we are not using it in production yet.

perlpunk (approved):
The next step for using it in production would be to add tests.

openqa-post-similarity
Akemiiii: Thank you. I will rewrite it as soon as possible.

okurz:
https://akemiiii.github.io/2022/09/08/GSoC-project/ is the related GSoC project report

PR #175, opened 2022-08-03
trigger-openqa_in_openqa: Fix lookup of latest Tumbleweed qcow

okurz:
We can generalize the search for assets by looking for any machine
suffix. Lately the qcow generating test scenario was changed to run on
the machine variant "64bit-3G" so our lookup did not match anymore.

Tested manually locally by calling

```
openqa-cli api --host https://openqa.opensuse.org assets | jq -r '[.assets[] | select(.name | test("Tumbleweed-x86_64-20220801-Tumbleweed\\@64bit.*.qcow")) | select(.size != null)] | .[0] | .name'
```

which yields "opensuse-Tumbleweed-x86_64-20220801-Tumbleweed@64bit-3G.qcow2

Related progress issue: https://progress.opensuse.org/issues/114944

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -78,8 +78,10 @@ def main(args):
     changes = { GOOD: set(), BAD: set() }
     if 'diff_to_last_good' not in investigation:
         return
+    diff_to_last_good = investigation['diff_to_last_good'].split('\n')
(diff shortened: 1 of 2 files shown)


okurz:
whoopsie, see https://github.com/os-autoinst/scripts/pull/176

PR #180, opened 2022-08-15
Add a wrapper for openqa-cli for better error handling

perlpunk:
Now we will see the arguments and the calling command and line

Issue: https://progress.opensuse.org/issues/115178


openqa-investigate
```diff
@@ -242,6 +242,8 @@ main() {
         client_call=(openqa-cli "${client_args[@]}")
         if [[ -n "$client_prefix" ]]; then
             client_call=("$client_prefix" "${client_call[@]}")
+        else
+            client_call=("runcli" "${client_call[@]}")
```
okurz, line 246: can this be combined into something like

```suggestion
        client_prefix="${client_prefix:-runcli}"
        client_call=("$client_prefix" "${client_call[@]}")
```

Martchus (commented):
Looks generally good. Maybe it is also worth improving logging options in `openqa-cli` itself.

openqa-investigate
perlpunk: done

okurz (approved):

PR #181, opened 2022-08-16
Do not send E-Mail notifications for restarted jobs

alvarocarvajald:
Proposal: avoid sending E-Mail notifications for jobs that were automatically restarted (for example, via **RETRY** setting) by checking if the job has been cloned.

### DIFF
--- a/_common
+++ b/_common
@@ -131,7 +131,9 @@ handle_unknown() {
         group_description=$(echo "$group_data" | runjq -r '.[0].description' ) || true
         group_mailto=$(echo "$group_description" | perl -n -wE'm/.*MAILTO: (\S+).*/ and say $1' | head -1)
         group_mailto=${group_mailto:-$notification_address}
-        if [[ -n "$group_mailto" ]]; then
+        # Avoid sending notifications for jobs that were restarted
+        clone_id="$(echo "$job_data" | runjq -r '.job.clone_id')"
+        if [[ -n "$group_mailto" && "$clone_id" == 'null' ]]; then
             group_name=$(echo "$group_data" | runjq -r '.[0].name')
             job_name=$(echo "$job_data" | runjq -r '.job.name')
             job_result=$(echo "$job_data" | runjq -r '.job.result')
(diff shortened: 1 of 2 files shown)


perlpunk:
I pushed a fix for the test

alvarocarvajald:
> I pushed a fix for the test

Oh, thanks @perlpunk

okurz (changes requested):
This goes in the right direction but IMHO we shouldn't even have the complete script called on jobs that are already retried. Maybe a regression in openQA?

okurz (approved):
Thought about it and it looks like your approach is actually completely fine. We might want to consider moving the logic that early-aborts on retry some lines above so that we abort `handle_unknown` earlier but we should think about that further. So good as is, thank you!

_common
```diff
@@ -131,7 +131,9 @@ handle_unknown() {
         group_description=$(echo "$group_data" | runjq -r '.[0].description' ) || true
         group_mailto=$(echo "$group_description" | perl -n -wE'm/.*MAILTO: (\S+).*/ and say $1' | head -1)
         group_mailto=${group_mailto:-$notification_address}
-        if [[ -n "$group_mailto" ]]; then
+        # Avoid sending notifications for jobs that were restarted
+        clone_id="$(echo "$job_data" | runjq -r '.job.clone_id')"
```
Martchus, line 135:
```suggestion
        clone_id=$(echo "$job_data" | runjq -r '.job.clone_id')
```

Martchus (approved):

_common
alvarocarvajald: Did not notice the double double quotes there. It works though:
```
$ job_data='{"job": {"name": "foo", "result": "failed"}}'
$ A="$(echo "$job_data" | jq -r '.job.clone_id')"
$ echo $A
null
$
```
However this got merged right after your approval. Should I submit a new PR to get rid of the outermost double quotes, or leave as is?

Martchus: You can leave it as-is. I've just mentioned it because quotes are not necessary when assigning variables like this. (They to no harm of course.)

PR #182, opened 2022-08-18
Fix bisect to support _TEST_REPOS

mimi1vx:


mimi1vx:
unfortuanetly it still has pretty big issue in logic --> ALL var settings must be used together, it now create jobs containing only one bisected variable which is incorrect

perlpunk:
Maybe split out your reformatting to a new PR?
I am also working on the script and just created a PR.
If your reformatting was in its own PR it could have been merged already and I would have used that as a base.

mimi1vx:
@perlpunk splitted, reformat merged .. 

still part where are jobs created and posted is fundamentally wrong :(

tests/test_trigger_bisect_jobs.py
```diff
@@ -247,15 +249,19 @@ def test_network_problems():
...
 
 def test_issue_types():
-    investigation = """
-    - "OS_TEST_ISSUES": "1,2,3,4",
-    + "OS_TEST_ISSUES": "1,2,3,4,5",
-    - "OTHER_TEST_ISSUES": "23",
-    + "OTHER_TEST_ISSUES": "24",
-    + "DUMMY_TEST_ISSUES": "25,26,27",
-    """
+    investigation = '- "OS_TEST_ISSUES": "1,2,3,4",\n+ "OS_TEST_ISSUES": "1,2,3,4,5",\n- "OTHER_TEST_ISSUES": "23",\n+ "OTHER_TEST_ISSUES": "24",\n+ "DUMMY_TEST_ISSUES": "25,26,27",'
```
perlpunk, line 252: I'm not sure why this change was made.
This is much less readable than before.

PR #185, opened 2022-08-22
openqa-trigger-bisect-jobs: fix trigger logic in bisect to use all variables

mimi1vx:
Original code used all *_TEST_* test variables as independent entities.
But in reality we need remove incident from all variables.

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
 import argparse
+from functools import total_ordering
+import hashlib
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


baierjan (approved):

openqa-trigger-bisect-jobs
```diff
@@ -219,35 +257,50 @@ def main(args):
...
-                    "TEST=" + test_name,
-                    "OPENQA_INVESTIGATE_ORIGIN=" + args.url,
-                    "MAINT_TEST_REPO="
-                ],
-                args.dry_run,
+        log.debug("[%s] removed: %s, added: %s" % (key, removed_key, added_key))
+        added += added_key
+
+    # whole sort is to simplify testability of code
+    for issue in sorted(list({i.incident_id for i in added}), key=int):
```
kalikiana, line 273: The comment already says this is for the benefit of unit tests. Which might be a bit wasteful if it's otherwise not needed. Did you look into alternatives? It kind of looks like all this does is order the log messages. Maybe they can be sorted afterwards?

kalikiana (dismissed):
See my question with regard to sorting. Also please try and pick more meaningful titles and commit messages here. "fix" doesn't tell the reader what you're trying to achieve and you're not explaining how you're doing it. Also, a ticket reference might help add context.

openqa-trigger-bisect-jobs
mimi1vx: I looked , unfortuanetly `MagicMock` from `unittest` excepts exact values in assert so we are here

btw. If you look to original code --> it has more sorts only for sake of unit tests

kalikiana (approved):

openqa-trigger-bisect-jobs
perlpunk: The sorts are for unit tests, but also for humans. How many of those issues do we have at a maximum? 30? Then the sort will not make any visible difference to the total runtime of the bisect script.

PR #187, opened 2022-09-15
Trigger test to generate pflash vars image

perlpunk:
Issue: https://progress.opensuse.org/issues/113794

### DIFF
--- a/.github/workflows/schedule.yml
+++ b/.github/workflows/schedule.yml
@@ -1,0 +1,21 @@
+---
+name: schedule
+on:
+  push:
+    branches: ['*pflash*']
+  schedule:
+    - cron: '23 5 * * 0'
+  workflow_dispatch:
+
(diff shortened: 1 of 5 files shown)


.github/workflows/schedule.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  schedule:
+    - cron: '23 5 * * 0'
+  workflow_dispatch:
+
+jobs:
+  schedule:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.opensuse.org/devel/openqa/ci/containers/base:latest
```
okurz, line 12: You can use registry.opensuse.org/home/okurz/container/containers/tumbleweed:openQA-client and then don't need to call `zypper ref` and the client install

okurz (approved):

PR #191, opened 2022-12-21
investigate: Skip passed tests

perlpunk:
In order to enable _TRIGGER_JOB_DONE_HOOK=1 for investigate jobs we first have to make sure it exits early for passed jobs.

Later we can add the requested comment logic.


openqa-investigate
```diff
@@ -187,6 +187,12 @@ investigate() {
     local rc=0
 
     job_data=$(openqa-cli "${client_args[@]}" --json jobs/"$id")
+    state="$(echo "$job_data" | runjq -r '.job.state')" || return $?
+    result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
+    # Skip all unfinished or passed jobs for now
+    if [[ "$state" != 'done' || "$result" == passed ]]; then
+        return
+    fi
```
okurz, line 195: Likely we want to have this skipping in https://github.com/os-autoinst/scripts/blob/master/openqa-label-known-issues instead of here

perlpunk:
See #192 instead

PR #193, opened 2022-12-21
Set _TRIGGER_JOB_DONE_HOOK=1 for investigation jobs

perlpunk:

Then passed jobs will also trigger the hooks for post investigation processing.

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -47,7 +47,7 @@ clone() {
 
     name="$(echo "$clone_job_data" | runjq -r '.job.test'):investigate$name_suffix" || return $?
     base_prio=$(echo "$clone_job_data" | runjq -r '.job.priority') || return $?
-    clone_settings=("TEST+=:investigate$name_suffix" '_GROUP_ID=0' 'BUILD=')
+    clone_settings=("TEST+=:investigate$name_suffix" '_TRIGGER_JOB_DONE_HOOK=1' '_GROUP_ID=0' 'BUILD=')
     if [[ $refspec ]]; then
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

openqa-investigate
```diff
@@ -47,7 +47,7 @@ clone() {
 
     name="$(echo "$clone_job_data" | runjq -r '.job.test'):investigate$name_suffix" || return $?
     base_prio=$(echo "$clone_job_data" | runjq -r '.job.priority') || return $?
-    clone_settings=("TEST+=:investigate$name_suffix" '_GROUP_ID=0' 'BUILD=')
+    clone_settings=("TEST+=:investigate$name_suffix" ''_TRIGGER_JOB_DONE_HOOK=1' _GROUP_ID=0' 'BUILD=')
```
Martchus, line 50: The quoting looks a little odd.

Martchus (changes requested):

openqa-investigate
perlpunk: Oops. fixed

perlpunk: and pushed...

Martchus (approved):

kalikiana (approved):

okurz (approved):

PR #196, opened 2023-01-02
Use test-tap's try and has functions

ingydotnet:
Based on #194

### DIFF
--- a/test/01-label-known-issues.t
+++ b/test/01-label-known-issues.t
@@ -11,66 +11,52 @@ mock-client() {
     client_output+="client_call $@"$'\n'
 }
 
-nl=$'\n'
 client_call=(mock-client "${client_call[@]}")
 logfile1=$dir/data/01-os-autoinst.txt.1
 logfile2=$dir/data/01-os-autoinst.txt.2
 
-rc=0
-comment_on_job 123 Label || rc=$?
-is "$rc" 0 'successful comment_on_job'
-is "$client_output" "client_call -X POST jobs/123/comments text=Label$nl" 'comment_on_job works'
+try-client-output() {
+  out=$logfile1
+  client_output=''
+  try "$*"' && echo "$client_output"'
+}
 
-rc=0
-search_log 123 'foo.*bar' "$logfile1" || rc=$?
+try-client-output comment_on_job 123 Label
+is "$rc" 0 'successful comment_on_job'
+is "$got" "client_call -X POST jobs/123/comments text=Label" 'comment_on_job works'
+
+try search_log 123 'foo.*bar' "$logfile1"
 is "$rc" 0 'successful search_log'
 
-rc=0
-search_log 123 'foo.*bar' "$logfile2" || rc=$?
+try search_log 123 'foo.*bar' "$logfile2"
 is "$rc" 1 'failing search_log'
 
-rc=0
-output=$(search_log 123 'foo [z-a]' "$logfile2" 2>&1) || rc=$?
+try "search_log 123 'foo [z-a]' $logfile2"
 is "$rc" 2 'search_log with invalid pattern'
-like "$output" 'range out of order in character class' 'correct error message'
+like "$got" 'range out of order in character class' 'correct error message'
 
-rc=0
-client_output=''
-out=$logfile1
-label_on_issue 123 'foo.*bar' Label 1 softfailed || rc=$?
+try-client-output label_on_issue 123 'foo.*bar' Label 1 softfailed
 expected="client_call -X POST jobs/123/comments text=Label
-client_call -X POST jobs/123/restart
-"
+client_call -X POST jobs/123/restart"
 is "$rc" 0 'successful label_on_issue'
-is "$client_output" "$expected" 'label_on_issue with restart and disabled force_result'
+is "$got" "$expected" 'label_on_issue with restart and disabled force_result'
 
-rc=0
-client_output=''
-out=$logfile1
-enable_force_result=true label_on_issue 123 'foo.*bar' Label 1 softfailed || rc=$?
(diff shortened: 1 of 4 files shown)


okurz:
@mergify rebase

test/02-investigate.t
```diff
@@ -64,53 +64,45 @@ openqa-cli() {
...
+has "$got" '1234' 'comment ID returned'
 
 # test syncing via investigation comment; we're second
-rc=0
-out=$(force=true sync_via_investigation_comment 32 32 2>&1) || rc=$?
+try force=true sync_via_investigation_comment 32 32
 is "$rc" 0 'skip with success if we do not own first investigation comment'
-like "$out" '' 'no output when skipping'
+# XXX What is this testing?
+like "$got" '' 'no output when skipping'
```
kalikiana, line 105: Should be https://github.com/os-autoinst/scripts/blob/master/openqa-investigate#L230

kalikiana: Generally there's a message that we're skipping. So this should ensure we don't in this case

test/01-label-known-issues.t
```diff
@@ -73,22 +73,22 @@ testurl=https://openqa.opensuse.org/api/v1/jobs/2291399
...
+has "$out" 'Subject: Unreviewed issue (Group 24 openQA)' "send-email subject like expected"
+has "$out" 'From: openqa-label-known-issues <foo@bar>' "send-email from called like expected"
+has "$out" 'To: dummy@example.com.dummy' "send-email to like expected"
 like "$out" '8<.*Backend process died.*>8' 'Log excerpt in mail'
-like "$out" 'Content-Type: text/html' 'mail has text part'
-like "$out" 'Content-Type: text/plain' 'mail has HTML part'
-like "$out" '<li>Name: foo' 'mail contains job name'
+has "$out" 'Content-Type: text/html' 'mail has text part'
+has "$out" 'Content-Type: text/plain' 'mail has HTML part'
+has "$out" 'Name: foo' 'mail contains job name'
```
perlpunk, line 82: why did you remove the `<li>`?

test/04-common.t
```diff
@@ -15,14 +15,13 @@ failure() {
...
+try runcli success a b c
 is $rc 0 "runcli success"
-is "$output" "SUCCESS a b c" "runcli successful output"
+is "$got" "SUCCESS a b c" "runcli successful output"
 
-output=$(runcli failure a b c 2>&1) || rc=$?
+try runcli failure a b c
 is $rc 23 "runcli failure"
-like "$output", "test/04-common.t.*failure a b c.*oh noe" "runcli failure output"
+like "$got", "failure a b c.*oh noe" "runcli failure output"
```
perlpunk, line 24: I noticed you removed the `test/04-common.t.*` from the match.
This was actually making sure that the error message contained information where it happened:
```
openqa-cli (34 test/04-common.t): 
```
With this PR it says:
```
openqa-cli (168 /home/tina/openqa-devel/repos/scripts/bpan/.bpan/lib/test-tap.bash):
```
So I assume it needs to go one caller level deeper somewhere. Not sure if that's possible with `try`.

test/01-label-known-issues.t
ingydotnet: Fixed. See 5e8bef9 for details.

test/04-common.t
ingydotnet: Right. There is stuff in test-tap-bash to adjust the call stack level on error messages.

I need to figure out a patch if you need this to have the file name.

BTW, just noticed the likely unintended comma in that line :)

I'll work on a fix for this.

ingydotnet: Fixed in da12258

test/02-investigate.t
ingydotnet: @kalikiana the `like` testing function applies a regex to a string, so this is testing that the output _contains_ the empty string (regex), which would always be true.

I tried changing this to `is "$got" '' 'no output when skipping'` which I would expect is what you want, but the test fails.

I'll leave that for you to investigate. :)

okurz (commented):
+1

okurz (approved):

_common
```diff
@@ -36,7 +37,7 @@ runcli() {
     rc=$?
     set -e
     [[ "$rc" == 0 ]] && echo "$output" && return
-    warn "openqa-cli ($(caller)): (${args[*]}) rc: $rc >>>$output<<<"
+    warn "openqa-cli ($(caller "$runcli_caller_level")): (${args[*]}) rc: $rc >>>$output<<<"
```
perlpunk, line 40: I was hoping for something more generic in the `try` function, e.g. in perl you can do:
```
some test_helper_function {
    local $Test::Builder::Level = $Test::Builder::Level + 1;
    ...
}
```
The error message including the caller we wouldn't just need for the `runcli` function, but I would expect the right caller level in every function we test.
OTOH the level just wrong in the tests, so I'd prefer to remove this commit again.

perlpunk: Maybe it could be done by adding a wrapper around the `caller` function that uses such a `level` variable, and `try` will increase it.

ingydotnet: That strategy worked in b4455c0.

ingydotnet:
Squashed commits getting the warning caller to work in [test/04-common.t](https://github.com/os-autoinst/scripts/pull/196/files#diff-033dbd7ac4c2e9fab2c941a9379c1bc220198d43cff36d547a9b36a4abcb5909)

See 0804876

perlpunk (approved):
thanks!

ingydotnet:
@perlpunk , you're welcome. Will rebase into #197 now.

kraih (approved):

test/02-investigate.t
perlpunk: yeah, leave it like that and we will investigate it

PR #200, opened 2023-01-18
Make hook script easier to test

perlpunk:

By adding the
    
    caller 0 >/dev/null || hook "$@"
    
line one can source the script and test its functions without having to call
the whole script.
    
By making the calls to the scripts pathless one can mock
it in the test easily with a function.
That's also what a user can do if they want to call the
hook script and use a different openqa-investigate script/function.

### DIFF
--- a/openqa-label-known-issues-and-investigate-hook
+++ b/openqa-label-known-issues-and-investigate-hook
@@ -5,22 +5,25 @@ set -eo pipefail
 # "hook script" intended to be called by openQA instances taking a job ID as
 # parameter and forwarding a complete job URL to "openqa-label-known-issues"
 # on stdin and all left unknowns to "openqa-investigate"
+dir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)
+PATH=$dir:$PATH
 
-id="${1:?"Need 'job_id'"}"
 host="${host:-"openqa.opensuse.org"}"
 scheme="${scheme:-"https"}"
 host_url="$scheme://$host"
-dirname=$(dirname "$0")
-script_dir=${script_dir:-$dirname}
 
 investigate-and-bisect() {
     local rc=0 test
     read -r test
-    echo "$test" | "$script_dir/openqa-investigate" || rc=$?
+    echo "$test" | openqa-investigate || rc=$?
     [[ "$rc" -eq 142 ]] && return "$rc"
-    "$script_dir/openqa-trigger-bisect-jobs" --url "$test" || rc=$?
+    openqa-trigger-bisect-jobs --url "$test" || rc=$?
     return "$rc"
 }
 
-echo "$host_url/tests/$id" | "$script_dir/openqa-label-known-issues" | sed -n 's/\[\([^]]*\)\].*Unknown issue, to be reviewed.*/\1/p' | investigate-and-bisect
+hook() {
+    local id="${1:?"Need 'job_id'"}"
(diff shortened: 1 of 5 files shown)


openqa-label-known-issues-and-investigate-hook
```diff
@@ -6,21 +6,22 @@ set -eo pipefail
...
 scheme="${scheme:-"https"}"
 host_url="$scheme://$host"
-dirname=$(dirname "$0")
-script_dir=${script_dir:-$dirname}
 
 investigate-and-bisect() {
     local rc=0 test
     read -r test
-    echo "$test" | "$script_dir/openqa-investigate" || rc=$?
+    echo "$test" | "$(dirname "$0")"/openqa-investigate || rc=$?
```
Martchus, line 16: It would make sense to still use the `$script_dir` variable so `dirname` needs to be invoked only once.

openqa-label-known-issues-and-investigate-hook
```diff
@@ -6,21 +6,22 @@ set -eo pipefail
...
     [[ "$rc" -eq 142 ]] && return "$rc"
-    "$script_dir/openqa-trigger-bisect-jobs" --url "$test" || rc=$?
+    "$(dirname "$0")"/openqa-trigger-bisect-jobs --url "$test" || rc=$?
     return "$rc"
 }
 
-echo "$host_url/tests/$id" | "$script_dir/openqa-label-known-issues" | sed -n 's/\[\([^]]*\)\].*Unknown issue, to be reviewed.*/\1/p' | investigate-and-bisect
+hook() {
+    id="${1:?"Need 'job_id'"}"
```
Martchus, line 23:
```suggestion
    local id="${1:?"Need 'job_id'"}"
```

Martchus (changes requested):

openqa-label-known-issues-and-investigate-hook
```diff
@@ -6,21 +6,22 @@ set -eo pipefail
...
+    echo "$host_url/tests/$id" | "$(dirname "$0")"/openqa-label-known-issues | sed -n 's/\[\([^]]*\)\].*Unknown issue, to be reviewed.*/\1/p' | investigate-and-bisect
+}
 
+caller 0 >/dev/null || main "$@"
```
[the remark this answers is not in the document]
perlpunk: fixed

openqa-label-known-issues-and-investigate-hook
perlpunk: The main point is, I can't keep it as it is. I can't use a path to the script because I want to override it via functions in the test.
And that's also what other people can do if they want a different path.

I'm not sure how often I have the use case that I want to run this hook but the scripts from a certain directory.
In the beginning, when you wrote the hook, it was just `"$(dirname "$0")"/openqa-investigate`.

perlpunk: > It would make sense to still use the `$script_dir` variable so `dirname` needs to be invoked only once.

I fixed by setting `dir` at the beginning, and adding it to `PATH`.

perlpunk:
I added more details to the commit message

openqa-label-known-issues-and-investigate-hook
perlpunk: done

okurz (approved):

Martchus (approved):

PR #201, opened 2023-01-19
Make wrapper scripts for batch processing

perlpunk:
Make openqa-label-known-issues and openqa-investigate take a single url

and provide  wrapper scripts {openqa-label-known-issues,openqa-investigate}-multi for
batch operations.

This makesthe scripts easier to use and test.

We can call a single command for a single url that we pass as a
parameter.
Having to pipe the url makes it harder to use.
Also with a command for a single url we can easier write tests for
this command.
    
Related issue: https://progress.opensuse.org/issues/98862

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ bpan:
 
 .PHONY: test-online
 test-online:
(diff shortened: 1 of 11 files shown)


okurz (commented):
+1

okurz (approved):

PR #202, opened 2023-01-19
Reuse job_data in openqa-investigate

perlpunk:
Most of the times origin and id are the same, so better not query the data everytime we call clone.

Also rename the variable to make it clear it might be different from the outer job_data variable.

Related issue: https://progress.opensuse.org/issues/98862


okurz (commented):
+1

okurz (approved):

PR #205, opened 2023-01-23
Fix openqa-investigate exception handling

perlpunk:
Introduced by https://github.com/os-autoinst/scripts/pull/201/commits/060bbebfa24e6eb0a79492b2f383fe5ac8014da2

By making openqa-investigate operating on only a single id the

    || = rc=$?

was removed. As a result, failing commands inside the investigate() function would return on error, so query_dependency_data_or_postpone() and sync_via_investigation_comment() would just result in an exit 255.

So we didn't get any investigate jobs since this change.

Somehow that part doesn't seem to be tested well. Also it's a bit hard to follow the script's lofic here because we transport results with certain exit codes.

I'll check if I can improve the test in my current PR for posting comments.


openqa-investigate
```diff
@@ -231,7 +231,7 @@ investigate() {
     # have more, ambiguous potential changes that we need to bisect on
 
     # sync by writing initial investigation comment (edited later)
-    comment_id=$(sync_via_investigation_comment "$id" "$first_cluster_job_id"); rc=$?
+    comment_id=$(sync_via_investigation_comment "$id" "$first_cluster_job_id") || rc=$?
```
Martchus, line 234: What if `rc` is still 255 from the previous call and `sync_via_investigation_comment` returns 0? Then we wouldn't override the previous 255 and continue instead of returning 0. But returning 0 is what this code should do then, right?

okurz (approved):

Martchus (changes requested):
Maybe I'm wrong but it would make sense to answer the question from my inline comment first.

openqa-investigate
perlpunk: good point! I need to reset rc

perlpunk: #206

PR #206, opened 2023-01-23
Reset rc before next function call

perlpunk:
Otherwise it could still be set to non-zero from the previous call Thanks @Martchus


openqa-investigate
```diff
@@ -231,6 +231,7 @@ investigate() {
     # have more, ambiguous potential changes that we need to bisect on
 
     # sync by writing initial investigation comment (edited later)
+    rc=0
```
kalikiana, line 234: Could this be covered by a test?

okurz:
+1

Martchus (approved):

PR #207, opened 2023-01-25
Make regex more performant

perlpunk:
Having several `.*` in a regex with the `s` flag can involve a lot of backtracking on a logfile, but in reality only one of the .* must be line spanning.

Looking into the gru logs, this regex was timing out very often. Removing the s flag and replacing the line spanning .* with [\S\s]* will make it faster

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -91,7 +91,7 @@ label_on_issues_from_issue_tracker() {
 }
 
 label_on_issues_without_tickets() {
-    if label_on_issue "$id" '(?s)([dD]ownload.*failed.*404).*Result: setup failure' 'label:non_existing asset, candidate for removal or wrong settings'; then return
+    if label_on_issue "$id" '([dD]ownload.*failed.*404)[\S\s]*Result: setup failure' 'label:non_existing asset, candidate for removal or wrong settings'; then return
     elif label_on_issue "$id" 'File .*\.yaml.* does not exist at .*scheduler.pm' 'label:missing_schedule_file'; then return
     elif label_on_issue "$id" 'Compilation failed in require at .*isotovideo line 28.' 'label:schedule_compilation_error'; then return
     elif label_on_issue "$id" 'qemu-img: Could not open .*: No such file or directory' 'label:missing_asset'; then return
--- a/test/01-label-known-issues.t
+++ b/test/01-label-known-issues.t
@@ -2,7 +2,7 @@
 
 source test/init
 
-plan tests 25
+plan tests 26
 
 source _common
 
@@ -14,6 +14,7 @@ mock-client() {
 client_call=(mock-client "${client_call[@]}")
 logfile1=$dir/data/01-os-autoinst.txt.1
 logfile2=$dir/data/01-os-autoinst.txt.2
+logfile3=$dir/data/01-os-autoinst.txt.3
 
 try-client-output() {
   out=$logfile1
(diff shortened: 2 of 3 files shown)


perlpunk:
There is some weird behaviour with the `try` function:
https://github.com/os-autoinst/scripts/actions/runs/4010410344/jobs/6886889427#step:5:10
```
#   Failed test 'successful search_log'
#   at test/01-label-known-issues.t line 37.
#     got: '2'
#   expected: '0'
# >>/home/runner/work/scripts/scripts/bpan/.bpan/lib/test-tap.bash: eval: line 168: syntax error near unexpected token `('
# /home/runner/work/scripts/scripts/bpan/.bpan/lib/test-tap.bash: eval: line 168: `search_log 123 ([dD]ownload.*failed.*404)[\S\s]*Result: setup failure /home/runner/work/scripts/scripts/test/data/01-os-autoinst.txt.3'<<
```
If I reduce the regex to `Download` only it works.
If I just remove the `()` from the regex, e.g. `[dD]ownload.*failed.*404[\S\s]*Result: setup failure`, then the error message is:
``
# >>grep failed: cmd=>grep -qPzo '[dD]ownload.*failed.*404[Ss]*Result:' 'setup'< output='timeout: invalid time interval ‘failure’
# Try 'timeout --help' for more information.'<<
```
Adding an extra backslash `[\\S\\s]` doesn't really help, I can see slashes in the error message, but the error is the same.

**Comment by perlpunk** [comment]
I showed @ingydotnet the error, and he'll take a look.
Maybe we can go without a test for this or use the traditional output capturing here meanwhile.

**Review by okurz — APPROVED** [review]

**Comment by ingydotnet** [comment]
The bug is in https://github.com/perlpunk/scripts/blob/improve-regex/test/01-label-known-issues.t#L36

Bash regexes are not pcre. `[\s\S]` should just be `.`
and for the non-s-flag dot you might try `[^$'\n']`

```
$ (v=$'foo\nbar'; [[ $v =~ (.*) ]] && echo "${BASH_REMATCH[0]}")
foo
bar
$ (v=$'foo\nbar'; [[ $v =~ ([^$'\n']*) ]] && echo "${BASH_REMATCH[0]}")
foo
$ (v=$'foo\nbar'; [[ $v =~ ([^\n]*) ]] && echo "${BASH_REMATCH[0]}")   # WRONG (just means not 'n')
foo
bar
$ 
```

**Comment by ingydotnet** [comment]
That's a problem but not the bug. Sorry. :\
I'll get the test passing.

**Comment by ingydotnet** [comment]
Passes for me with this patch:

```
+ git diff
diff --git a/test/01-label-known-issues.t b/test/01-label-known-issues.t
index f6b0821..9af928e 100644
--- a/test/01-label-known-issues.t
+++ b/test/01-label-known-issues.t
@@ -26,17 +26,14 @@ try-client-output comment_on_job 123 Label
 is "$rc" 0 'successful comment_on_job'
 is "$got" "client_call -X POST jobs/123/comments text=Label" 'comment_on_job works'
 
-try search_log 123 'foo.*bar' "$logfile1"
+try "search_log 123 'foo.*bar' '$logfile1'"
 is "$rc" 0 'successful search_log'
 
-try search_log 123 'foo.*bar' "$logfile2"
+try "search_log 123 'foo.*bar' '$logfile2'"
 is "$rc" 1 'failing search_log'
 
-diag "##################################"
-try search_log 123 '([dD]ownload.*failed.*404)[\S\s]*Result: setup failure' "$logfile3"
+try "search_log 123 '([dD]ownload.*failed.*404)[\s\S]*Result: setup failure' '$logfile3'"
 is "$rc" 0 'successful search_log'
-diag ">>$got<<"
-diag "##################################"
 
 try "search_log 123 'foo [z-a]' $logfile2"
 is "$rc" 2 'search_log with invalid pattern'
```

perlpunk:
Updated PR, thanks @ingydotnet !

Issue #208, opened 2023-01-26
Offer to code review this repo's Bash


ingydotnet:
I've looked over the Bash code in this repo and many things are very good, but a lot of things could be improved.

I've been using Bash for 10 years now, as seriously as I use Perl, Ruby, Python, Node etc.
I've even recently written a Bash package management system (CPAN, RubyGems, PyPI, NPM) called BPAN.

I'd like to offer a code review of this repo for you. I'll make lots of suggestions and you can choose to apply the ones you like.
Most of the suggestions would just be minor syntax refactoring. Nothing (or very little) would be a major change suggestion.

But before I do, please let me know whether or not this would be appreciated.

Cheers, Ingy

okurz:
Sure, you already helped in previous pull requests and I would be happy to see more suggestions

ingydotnet:
Great. I think this will be fun and I'm sure I'll end up learning a lot as well. I still am constantly learning new/better ways to do things with Bash.

@okurz I started #209 with a dumb commit. I think making each suggestion a commit of just a couple lines, but a full explanation in the commit message might be a decent way to do this. If you agree then please say so in a comment or a 👍 in #209 and I'll get started.  :)

okurz:
Yep, good approach

okurz:
thanks for the bash script review. Since the initial issue we also could extend test coverage and improve style. Meanwhile for more complex scripts we reworked those in Python which IMHO provides more streamlined, easier and also faster tooling to ensure good style and test coverage

PR #209, opened 2023-01-26
WIP - Bash code review

ingydotnet:
Per https://github.com/os-autoinst/scripts/issues/208 I wish to code review the Bash content and make suggestions on possible improvements.

I haven't done this before so I'm not sure of the best way to go about it.
I'm thinking to use this pull request, where I will make a commit for each suggestion with a couple of code changes to demonstrate the objective.
Then we can can discuss it.

I don't expect the PR to ever be merged, but you might end up cherry-picking some commits in it.

Let me know if this approach works for you and I'll start adding commits soon after.


okurz:
The change itself is fine. Feel welcome to continue adding more in this PR or create an individual PR right away. I don't mind trivial PRs :)

ingydotnet:
Started out today making a bunch of changes to the Makefile.
Tried to make them into many separate commits, which is tricky when you get on a refactoring roll.
I spend nearly as much time working on Makefiles as I do with Bash. :)

To be clear, this code review will likely happen in fits and spurts over the next several (7-10?) days, as I have time and ambition.

I'll get into the Bash files tomorrow...

ingydotnet:
Just pushed a bunch of Bash review commits. Kept almost all changes to just `_common` as it had a lot to offer.

I still have lots more to get to, but out of time for today.

Makefile
```diff
@@ -1,37 +1,59 @@
...
-	git clone https://github.com/bpan-org/bpan.git --depth 1
+test-python:
+	$(call run-with,py.test,$@,\
+	py.test tests)
 
-.PHONY: test-online
 test-online:
-	cat ./tests/incompletes | env dry_run=1 bash -ex ./openqa-label-known-issues-multi
-	env dry_run=1 ./trigger-openqa_in_openqa
+	dry_run=1 bash -x ./openqa-label-known-issues-multi < ./tests/incompletes
```
okurz, line 31: I'd prefer to keep the `|` here. Any good reason why you prefer `<`?

okurz:
about "Add a run-with Makefile function" I am not sure it's a good idea to "just" warn. I would prefer a fatal failure.

Makefile
okurz: oh, sorry. You explained that well in the commit message

_common
```diff
@@ -67,9 +69,9 @@ runcurl() {
     response=$(curl -w "\n%{http_code}\n" "$@" 2>&1)
     rc=$?
     set -e
-    [[ "$rc" != 0 ]] && warn "curl ($(caller)): Error fetching ($*): $response" && return 1
+    [[ "$rc" == 0 ]] || die "curl ($(caller)): Error fetching ($*): $response"
     status_code=$(echo "$response" | tail -1)
-    [[ "$status_code" != 200 ]] && warn "curl ($(caller)): Error fetching url ($*): Got Status $status_code" && return 1
+    [[ "$status_code" == 200 ]] || die "curl ($(caller)): Error fetching url ($*): Got Status $status_code"
```
okurz: but calling die is having a different effect than just returning from this function. I have not checked all calls of runcurl but if you want to replace the usage here then we should make sure it's effectively having the same effect

_common
```diff
@@ -39,7 +39,7 @@ runcli() {
     output=$("${args[@]}" 2>&1)
     rc=$?
     set -e
-    [[ "$rc" == 0 ]] && echo "$output" && return
+    [[ $rc -eq 0 ]] && echo "$output" && return
```
okurz: I honestly knew about the subtleties of `==` vs. `eq` in bash but still prefered `==` as a more readable variant. You wrote what we *should* use but if there isn't a more compelling reason I'd prefer to just keep the "=="

okurz:
everyone, https://github.com/os-autoinst/scripts/pull/211 was merged but I highly recommend everyone to read all the git commit messages including the ones of already merged commits as @ingydotnet's explanations are really good and helpful for the future as well :)

ingydotnet:
> about "Add a run-with Makefile function" I am not sure it's a good idea to "just" warn. I would prefer a fatal failure.

See 31022d7

_common
ingydotnet: That's a very good point.
`exit` is not same as `return`, though with `set -e` returning non-zero could trigger an exit depending on how it was called.

This inspires me to create a test coverage facility.

_common
ingydotnet: I suppose if shellcheck isn't griping then it's very likely ok.
I hadn't thought about using `==` for numbers before but I can't yet think of a reason it wouldn't work.
In that case I'd stick with it. I'll let you know if I find any edge cases.
I might start using this as well if it's safe. :)

ingydotnet:
Rebased master and added some more commits.

ingydotnet:
Ironically 983e105 is what I imagined would be my first review commit as it is a pet peeve.

So much cargo-culted bash scripts and stackoverflow posts over-quote...

PR #210, opened 2023-01-27
README.md: Add notes regarding line-spanning regex

okurz:
Inspired by perlpunk's notes

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,11 @@ https://raw.githubusercontent.com/os-autoinst/scripts/master/openqa-query-for-jo
 for example to look for ticket 12345 call `openqa-query-for-job-label poo#12345`
 ```
 
+*Notes:* the regex used in `<search_term>` is very powerful but also can be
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,6 +67,11 @@ https://raw.githubusercontent.com/os-autoinst/scripts/master/openqa-query-for-jo
 for example to look for ticket 12345 call `openqa-query-for-job-label poo#12345`
 ```
 
+*Notes:* the regex used in `<search_term>` is very powerful but also can be
+expensive when `(?s)` is used. Every `.*` can span over multiple lines and
+that involves a lot of backtracking and might not be needed. Use `[^\n]*`
+over `.*` where possible or use  or use line spanning matches like `[\S\s]*`.
```
perlpunk, line 73:
```suggestion
over `.*` where possible or use line spanning matches like `[\S\s]*`.
```

Martchus (approved):

PR #211, opened 2023-01-30
Cherry-pick all bash code review commits I consider fine right away

okurz:
I cherry-picked the following commits from #209
that I consider right away to merge if others agree as well. Other not included commits can still be discussed:

* Rename a function using - instead of _
* Use `warn` in _common
* Add job_post_response file to 'make clean'
* Remove usage of `env` from Makefile rules
* Split `test-unit` Makefile rule into 2 rules
* Add some standard settings to .setup.mk
* Allow user to select which files to prove
* Change bpan/ to .bpan/
* Remove unneeded .PHONY lines
* Add a 'clean' target to the Makefile
* README.md: Add notes regarding line-spanning regex

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-/bpan/
+/.bpan/
 /tests/__pycache__/
 /__pycache__/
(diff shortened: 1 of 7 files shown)


baierjan (approved):
Looks good, hopefully all `label-on-issue` calls are renamed

PR #213, opened 2023-02-13
Add openqa-get-job-schedule

grisu48:
Add script for extracting the SCHEDULE of an openQA job

### DIFF
--- a/openqa-get-job-schedule
+++ b/openqa-get-job-schedule
@@ -1,0 +1,129 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+# 
+# Extract a SCHEDULE parameter for the given job
+# 
+
+import json
+import sys
+from openqa_client.client import OpenQA_Client
+
+def get_schedule(instance, job) :
+    '''
+    Build a SCHEDULE parameter from the given job (id) on the given instance
+    '''
+    schedule = []
+    
+    ## Configure instance
+    client = OpenQA_Client(server=instance)
+    path = "jobs/%d/details" % (job)
+    job = client.openqa_request(method="GET", path=path)['job']
+    # iterate over testresults. category is only set for a new category (e.g. "installation", "console", ecc.)
+    category = ""
+    for result in job['testresults'] :
+        if "category" in result : category = result['category']
+        name = result['name']
+        # Ignore fragment
+        if '#' in name :
+            i = name.find('#')
+            name = name[:i]
+        schedule.append("%s/%s" % (category, name))
+    return schedule
+
+def can_convert_to_int(x) :
+    try :
+        int(x)
+        return True
+    except ValueError :
+        return False
+
+def print_schedule(schedule, fmt="schedule") :
+    # add 'tests/' as prefix
+    schedule = map(lambda x : 'tests/' + x, schedule)
+    if fmt == "" or fmt == "plain" or fmt == "schedule" :
+        text = ",".join(schedule)
+        print("SCHEDULE=%s" % text)
+    elif fmt == "csv":
+        text = ",".join(schedule)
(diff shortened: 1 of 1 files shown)


openqa-get-job-schedule
```diff
@@ -0,0 +1,121 @@
...
+		if "category" in result : category = result['category']
+		name = result['name']
+		# Ignore fragment
+		if '#' in name :
+			i = name.find('#')
+			name = name[:i]
+		schedule.append("%s/%s" % (category, name))
+	return schedule
+
+def is_int(x) :
```
Martchus, line 33:
```suggestion
def can_convert_to_int(x) 
```

openqa-get-job-schedule
```diff
@@ -0,0 +1,121 @@
...
+	try :
+		x = int(x)
```
Martchus, line 35: The assigned value will never the read, right? So I'd rather avoid it.

Of course it would be nice if that assignment would actually be effective. Then the function could be called `try_convert_to_int` and the converted value could be used by the caller (avoiding the 2nd conversion). Maybe something like this can be achieved by using a reference (in Perl it would be very easy, not sure about Python)?

openqa-get-job-schedule
```diff
@@ -0,0 +1,121 @@
...
+	schedule = map(lambda x : 'tests/' + x, schedule)
+	if fmt == "" or fmt == "plain" or fmt == "schedule" :
+		text = ",".join(schedule)
+		print("SCHEDULE=%s" % text)
+	elif fmt == "csv":
+		text = ",".join(schedule)
+		print("%s" % text)
+	elif fmt == "list":
+		for item in schedule:
+				print(item)
```
Martchus, line 51:
```suggestion
			print(item)
```

openqa-get-job-schedule
```diff
@@ -0,0 +1,121 @@
...
+		elif url == "--list" or url == "--plain":
+			fmt = "list"
+		elif url == "--osd" :
+			instance = "https://openqa.suse.de"
+		elif url == "--ooo" or url == "--o3" :
+			instance = "https://openqa.opensuse.org"
+		elif "/tests/" in url :         # https://openqa.opensuse.org/tests/12345
+			i = url.find("/tests/")
+			instance = url[:i]
+			job = int(url[i+7:])
```
Martchus, line 103: What if this conversion fails? I suppose the error handling could be improved in various places in this script.

Martchus (commented):
Maybe a lookup table for certain values (such as `--osd`) and regex matching for others would make more sense than this huge `if`-`elif` construct.

openqa-get-job-schedule
grisu48: In Python you pass objects and integer objects are immutable, so something like a reference to a int does not exists naturally.

Given the rather simple approach to the problem, I also feel that this would be very much overshoot the problem.

openqa-get-job-schedule
grisu48: Ah yes!

openqa-get-job-schedule
grisu48: I've added a `try..except` block for the possible `int` mismatch, but I wouldn't go further for a simple script here. WDYT?

okurz (changes requested):
Please use 4 tabs for indention. See existing files like https://github.com/os-autoinst/scripts/blob/master/openqa-trigger-bisect-jobs

Martchus:
You mean 4 spaces? :-)

okurz:
> You mean 4 spaces? :-)

*facepalm*, yes please :)

okurz (approved):
I suggest to also add a test for this file

PR #214, opened 2023-02-13
Add openqa-get-job-runtime-stats

grisu48:
Adds script to create runtime statistics for a range of openQA test runs.

### DIFF
--- a/openqa-get-job-runtime-stats
+++ b/openqa-get-job-runtime-stats
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# 
+# Summary: Display the test runtime statistics for a range of openQA test runs
+# Job ranges can be defined via 7..9 (i,e, [7,8, 9]) or 5+2 (i.e. [5,6,7])
+# 
+# Usage example: ./openqa-get-job-runtime-stats https://openqa.opensuse.org/tests/100000+10
+# 
+
+import requests
+import datetime
+import time
+import json
+import sys
+import argparse
+import numpy as np
+
+def parse_t(dt_str) :
+	# "2023-01-25T10:22:21"
+	if dt_str is None : return None
+	dt_str = dt_str.strip()
+	if dt_str == "" : return None
+	return datetime.datetime.strptime(dt_str, '%Y-%m-%dT%H:%M:%S')
+
+
+class Job :
+	# Init job by the json object fetched from OpenQA
+	def __init__(self, obj) :
+		# Apply json values to this class. Yes, this is hacky.
+		for k in obj :
+			self.__dict__[k] = obj[k]
+	
+	def done(self) -> bool :
+		return self.state == "done"
+	def passed(self) -> bool :
+		return self.state == "done" and self.result == "passed"
+	def softfailed(self) -> bool :
+		return self.state == "done" and self.result == "softfailed"
+	def isNotOK(self) -> bool :
+		return self.done() and not (self.passed() or self.softfailed())
+
+	def runtime(self) :
+		# Note: Timezone parsing is not needed here
+		t_finished = parse_t(self.t_finished)
+		if t_finished is None : return 0
+		t_started = parse_t(self.t_started)
+		if t_started is None : return 0
+		diff = t_finished - t_started
+		return diff.total_seconds()
+	
+	def __item__(self, index) :
+		return self._obj[index]
+
+
+def parse_job_number(jobstr) :
+	# Range? (Start..End)
+	if ".." in jobstr :
+		i = jobstr.find("..")
+		try :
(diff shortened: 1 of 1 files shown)


openqa-get-job-runtime-stats
```diff
@@ -0,0 +1,186 @@
...
+			self.__dict__[k] = obj[k]
+	
+	def done(self) -> bool :
+		return self.state == "done"
+	def passed(self) -> bool :
+		return self.state == "done" and self.result == "passed"
+	def softfailed(self) -> bool :
+		return self.state == "done" and self.result == "softfailed"
+	def failed(self) -> bool :
+		return self.state == "done" and (self.result == "failed" or self.result == "incomplete")
```
Martchus, line 40: It is not good to call this `failed` when it is actually `failed` + `incomplete`. Note that there are also more states you might want to consider here. For instance, in openQA we have meta states/results, e.g. `OK_RESULTS` (see https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Jobs/Constants.pm#L76) and the negation is usually called `not_ok` (when used somewhere).

openqa-get-job-runtime-stats
```diff
@@ -0,0 +1,186 @@
...
+		except ValueError :
+			raise ValueError("invalid job id in job sequence")
+	
+	# Try to parse as int
+	try :
+		return [int(jobstr)]
+	except ValueError:
+		raise ValueError("invalid job identifier")
+
+def get_job_api_urls(arg, default_remote="https://openqa.opensuse.org") :
```
Martchus, line 81: This function is never called with a 2nd argument. Maybe it makes sense to make the `default_remote` actually configurable from the outside?

openqa-get-job-runtime-stats
```diff
@@ -0,0 +1,186 @@
...
+				median = np.median(runtime)
+				average = np.average(runtime)
+				stdev = np.std(runtime)
+				print("  Value range:                       %d-%d s" % (runtime.min(), runtime.max()))
+				print("  Median runtime:                    %.2f s" % (median))
+				print("  Average runtime:                   %.2f s" % (average ))
+				print("  Standard deviation:                %.2f s" % (stdev))
+				#print("  * Median-normalized values *")
+				#print("    Average runtime / median:        %.2f s" % (average / median))
+				#print("    Standard deviation / median:     %.2f s" % (stdev / median))
```
Martchus, line 182: I'd avoid checking-in disabled code. Decide whether you want it or not.

Martchus (changes requested):

openqa-get-job-runtime-stats
grisu48: Reverted the logic, please check again if you agree now.

openqa-get-job-runtime-stats
```diff
@@ -0,0 +1,187 @@
...
+		for k in obj :
+			self.__dict__[k] = obj[k]
+	
+	def done(self) -> bool :
+		return self.state == "done"
+	def passed(self) -> bool :
+		return self.state == "done" and self.result == "passed"
+	def softfailed(self) -> bool :
+		return self.state == "done" and self.result == "softfailed"
+	def failed(self) -> bool :
```
Martchus, line 39: I still would avoid calling it `failed` because that is not what it is. It is "not ok".

openqa-get-job-runtime-stats
```diff
@@ -0,0 +1,187 @@
...
+		if not self.done() : return False
+		return not (self.passed() or self.softfailed())
```
Martchus, line 41:
```suggestion
		return self.done() and not (self.passed() or self.softfailed())
```

openqa-get-job-runtime-stats
grisu48: Amended.

Martchus (approved):

grisu48:
Thank you!

okurz:
@Martchus you actually approved a python script using tabs instead of 4 spaces? How could you?!? :D

okurz:
Please change the file to use 4 tabs for indention. See existing files like https://github.com/os-autoinst/scripts/blob/master/openqa-trigger-bisect-jobs. Also please add at least a minimal corresponding test module in https://github.com/os-autoinst/scripts/tree/master/tests

Martchus:
I don't care about the indentation as long as it is consistent within one coherent piece of work. I don't think the scripts in this repo are a coherent piece of work (but rather a collection of various helpers).

Maybe it makes sense to bump the number of approvals required by mergify (to give more people a chance to disagree).

okurz:
> Maybe it makes sense to bump the number of approvals required by mergify (to give more people a chance to disagree).

Good idea -> https://github.com/os-autoinst/scripts/pull/215

PR #216, opened 2023-02-14
Require at least two approvals for automatic merge

Martchus:
Otherwise, when an external contributor creates a PR, it might be merged by only having one person from the team having a look at it.


okurz:
sorry, was like one second faster in #215 ;)

PR #217, opened 2023-02-15
Avoid sending mails about unreviewed e-mails by default

Martchus:
We received unwanted mails about unreviewed issues today and it turns out it was due to a manual invocation of `openqa-label-known-issues`. We should avoid this and only send mails when explicitly enabled.

See https://progress.opensuse.org/issues/124565


kalikiana (approved):

okurz:
As soon as you ensured that the according setting is set on o3+osd you can remove the "not-ready" label again

okurz (approved):

Martchus:
o3 is already configured, we're only waiting for https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/795

perlpunk (approved):

Martchus:
The change has been deployed on OSD.

PR #218, opened 2023-02-21
Improve wording in mails about unreviewed issues

Martchus:
* Make it clear that this is about *test* issues (and not about openQA itself like mails from logwarn)
* Avoid stating that the error is *likely* within the last lines before the shutdown; it could just be an outdated needle or a product bug
* State possible reasons for the failure below
* Give suggestions how to avoid receiving more of those mails for the same issue
* See https://progress.opensuse.org/issues/124212

### DIFF
--- a/_common
+++ b/_common
@@ -123,10 +123,10 @@ snip_end="    # --- >8 ---"
 handle_unknown() {
     local testurl=$1 out=$2 reason=$3 group_id=$4 email_unreviewed=$5 from_email=$6 notification_address=${7:-} job_data=${8:-} dry_run=${9:-}
     local group_data group_name group_description group_mailto header email excerpt=$snip_start
-    local job_name job_result job_info
+    local job_name job_result info
     to_review+=("$testurl ${reason:0:50}")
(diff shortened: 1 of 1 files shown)


_common
```diff
@@ -144,10 +144,17 @@ handle_unknown() {
...
             job_name=$(echo "$job_data" | runjq -r '.job.name')
             job_result=$(echo "$job_data" | runjq -r '.job.result')
-            job_info="* Name: $job_name
+            info="* Name: $job_name
 * Result: $job_result
-* Reason: $reason"
-            email="[$job_name]($testurl)"$'\n'"$header"$'\n'"$job_info"$'\n'$'\n'"$excerpt"
+* Reason: $reason
+
+It might be a product bug, an outdated needle, test code needing adaption or a
```
kalikiana, line 151:
```suggestion
It might be a product bug, an outdated needle, test code needing adaptation or a
```

Although the words are similar, adaption is something else. An adaption that dogs have compared to wolves is being able to process more starches found in human foods.

perlpunk (approved):

_common
Martchus: Looks like you're right (https://grammarist.com/usage/adaption-adaptation). Thanks for pointing things like this out (only that way one can learn).

kalikiana (approved):

PR #219, opened 2023-02-22
Skip comments about failing openQA-in-openQA test

Martchus:
Skip commenting about failing openQA-in-openQA tests on OBS by default as those comments are redundant to mails about unreviewed issues sent by the hook script.

The code takes care of cleaning up old comments. However, this does not help with mail notifications leading to
https://progress.opensuse.org/issues/124694. So to fix this, it is the easiest to simply avoid creating those comments.

### DIFF
--- a/monitor-openqa_job
+++ b/monitor-openqa_job
@@ -16,6 +16,7 @@ host="${host:-"https://openqa.opensuse.org"}"
 openqa_groupid="${openqa_groupid:-"24"}"
 obs_component="${obs_component:-"package"}"
 obs_package_name="${1:-""}"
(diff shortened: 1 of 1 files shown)


monitor-openqa_job
```diff
@@ -40,6 +40,7 @@ for job_id in $(job_ids job_post_response); do
 done
 
 [[ ${failed_jobs[*]} ]] || exit 0
+[[ $comment_on_obs ]] || exit 1
```
perlpunk, line 43: I think this variable must be declared first because of the `-u`

perlpunk (approved):

kalikiana (approved):

PR #220, opened 2023-03-08
Avoid confusing behavior by not combining force-result labels with bugrefs

Martchus:
* As now stated in the documentation, this is not recommended as the force-result label will not be evaluated on carry-over due to the bugref (which can be confusing).
* This reverts d17bd74b2c9e1ecfa057bbd2e2cd5403b09418b9. The main intention behind this change was to still have a bugref rendered as link (as the one within the label wouldn't be rendered as such). With the openQA change 8fa3634545c132cdc3e8e3b0c55e63ae1b601c98 this is no longer required.
* See https://progress.opensuse.org/issues/123724

### DIFF
--- a/_common
+++ b/_common
@@ -83,9 +83,7 @@ openqa-api-get() {
 comment_on_job() {
     local id=$1 comment=$2 force_result=${3:-''}
     if $enable_force_result && [[ -n $force_result ]]; then
-        comment="label:force_result:$force_result:$comment
-
-$comment"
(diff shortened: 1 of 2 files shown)


baierjan (commented):
`s/carray-over/carry-over/`

Otherwise OK.

Martchus:
I've fixed the typo.

baierjan (approved):

okurz (approved):

PR #223, opened 2023-03-29
Identify product issues

perlpunk:
Spike solution which uses the retry job to fetch the results of the other investigate jobs by parsing the investigation comment.

Use the retry job to fetch the results of the other investigate jobs by
parsing the investigation comment.

Issues:
* https://progress.opensuse.org/issues/126527
* https://progress.opensuse.org/issues/109920

Demo:
investigate-identify-product-issues

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -182,18 +182,80 @@ $comment_text"
     fi
 }
 
+fetch-investigation-results() {
+    local origin_job_id=$1
+    local state job investigate_type other_id result investigate_comment comment_lines
+
+    investigate_comment=$("${client_call[@]}" -X GET jobs/"$origin_job_id"/comments | runjq -r '[.[] | select(.text | contains("Automatic investigation jobs for job") and contains(":investigate:retry**:"))] | min_by(.id)') || return $?
+    [[ $investigate_comment == 'null' ]] && return
+    readarray -t comment_lines < <(echo "$investigate_comment" | runjq -r '.text') || return $?
+
+    for line in "${comment_lines[@]}"; do
+        if [[ $line =~ :investigate:([^:*]+).*\/t([0-9]+) ]]; then
+            investigate_type=${BASH_REMATCH[1]}
+            other_id=${BASH_REMATCH[2]}
+            job=$(openqa-cli "${client_args[@]}" --json experimental/jobs/"$other_id"/status)
+            state=$(echo "$job" | runjq -r '.state') || return $?
+            # at least one job is not finished, come back later
+            [[ $state != 'done' ]] && return 142
+
+            result=$(echo "$job" | runjq -r '.result') || return $?
+            echo "$investigate_type|$other_id|$result"
+        fi
+    done
+}
+
+is-ok() {
+    [[ $1 == passed || $1 == softfailed ]]
+}
+
+is-product-issue() {
+    local origin_job_id=$1
+    local state result investigate_type passed result_lines
+    local pass_lgt='' pass_lgb='' pass_lgtb=''
+    product_issue=false
+
+    readarray -t result_lines < <(fetch-investigation-results "$origin_job_id") || return $?
+
+    for line in "${result_lines[@]}"; do
+        if [[ $line =~ ^([^|]+)\|([^|]+)\|([^|]+) ]]; then
+            investigate_type=${BASH_REMATCH[1]} result=${BASH_REMATCH[3]}
+            is-ok "$result" && passed=true || passed=false
+            if [[ $investigate_type == last_good_tests_and_build ]]; then
+                pass_lgtb=$passed
+            elif [[ $investigate_type == last_good_build ]]; then
+                pass_lgb=$passed
+            elif [[ $investigate_type == last_good_tests ]]; then
+                pass_lgt=$passed
+            fi
+        fi
+    done
+    [[ -z "$pass_lgb" ]] && return # no last good build
+    if ([[ -z "$pass_lgtb" ]] || "$pass_lgtb") && ([[ -z "$pass_lgt" ]] || ! "$pass_lgt") && "$pass_lgb"; then
+        product_issue=true
+    fi
+}
+
 post-investigate() {
     local id=$1 old_name=$2
     local rc=0 status
     [[ ! "$old_name" =~ investigate:retry$ ]] && echo "Job is ':investigate:' already, skipping investigation" && return 0
-    result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
+    # We are in the investigate:retry job now
+    retry_result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
     investigate_origin="$(echo "$job_data" | runjq -r '.job.settings.OPENQA_INVESTIGATE_ORIGIN')" || return 1
     origin_job_id=${investigate_origin#"$host_url/t"}
+
     comment="Investigate retry job: $host_url/t$id"
-    if [[ "$result" == passed || "$result" == softfailed ]]; then
-        comment+=" $result, likely a sporadic failure"
+    if is-ok "$retry_result"; then
+        comment+=" $retry_result, likely a sporadic failure"
     else
         comment+=" failed, likely not a sporadic failure"
+        local product_issue=false
+        is-product-issue "$origin_job_id" || return $?
+
+        if "$product_issue"; then
+            comment+="."$'\n'"Jobs including the last good build are ok, likely a product issue"
+        fi
     fi
 
     # meanwhile the original job might have been deleted already, handle
--- a/test/02-investigate.t
+++ b/test/02-investigate.t
@@ -3,7 +3,7 @@
 source test/init
 bpan:source bashplus +err +fs +sym
 
-plan tests 39
+plan tests 46
 
 host=localhost
 url=https://localhost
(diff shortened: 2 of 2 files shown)


openqa-investigate
```diff
@@ -180,18 +180,60 @@ $comment_text"
...
+    local job="$(echo "$job_data" | runjq -r '.job')" || return $?
+
+    local IFS=$'\n'
+    local investigate_comment=($("${client_call[@]}" -X GET jobs/"$origin_job_id"/comments | runjq -r '[.[] | select(.text | contains("Automatic investigation jobs for job") and contains(":investigate:retry**:"))] | sort_by(.id) | first | .text')) || return $?
+
+    for line in "${investigate_comment[@]}"; do
+        if [[ $line =~ :investigate:([^*]+)\*\*:.*\/t([0-9]+) ]]; then
+            investigate_type=${BASH_REMATCH[1]}
+            other_id=${BASH_REMATCH[2]}
+            if [[ $other_id -ne $retry_id ]]; then
```
Martchus, line 195:
```suggestion
            if [[ $other_id -ne "$retry_id" ]]; then
```

openqa-investigate
```diff
@@ -180,18 +180,60 @@ $comment_text"
...
+
 post-investigate() {
     local id=$1 old_name=$2
     local rc=0 status
     [[ ! "$old_name" =~ investigate:retry$ ]] && echo "Job is ':investigate:' already, skipping investigation" && return 0
-    result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
+    retry_result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
     investigate_origin="$(echo "$job_data" | runjq -r '.job.settings.OPENQA_INVESTIGATE_ORIGIN')" || return 1
     origin_job_id=${investigate_origin#"$host_url/t"}
+    warn "post-investigate $id $old_name origin_job_id=$origin_job_id"
```
Martchus, line 220: Not sure what this warning would tell me if I'd encountered it in a month.

Martchus (commented):
I guess this makes generally sense.

openqa-investigate
```diff
@@ -180,18 +180,60 @@ $comment_text"
     fi
 }
 
+fetch-investigate-results() {
+    local origin_job_id=$1 retry_id=$2
+    local state other_job_data state result investigate_type passed
+    local job="$(echo "$job_data" | runjq -r '.job')" || return $?
+
+    local IFS=$'\n'
+    local investigate_comment=($("${client_call[@]}" -X GET jobs/"$origin_job_id"/comments | runjq -r '[.[] | select(.text | contains("Automatic investigation jobs for job") and contains(":investigate:retry**:"))] | sort_by(.id) | first | .text')) || return $?
```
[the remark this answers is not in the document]
perlpunk: > Sorting the whole array is inefficient when you really just want to get the min/max element:

I copied this from your `sync_via_investigation_comment` ;-)

> By the way, I'm wondering whether we really want the comment with the lowest ID here. If we wanted the most recent comment that would be the one with the highest ID (so in my suggestion one would use `max_by`).

Normally we only should have one such investigation comment anyway. So it probably doesn't matter much. Let's suggest `max_by` for the actual ticket then.

openqa-investigate
```diff
@@ -180,18 +180,60 @@ $comment_text"
...
     local id=$1 old_name=$2
     local rc=0 status
     [[ ! "$old_name" =~ investigate:retry$ ]] && echo "Job is ':investigate:' already, skipping investigation" && return 0
-    result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
+    retry_result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
     investigate_origin="$(echo "$job_data" | runjq -r '.job.settings.OPENQA_INVESTIGATE_ORIGIN')" || return 1
     origin_job_id=${investigate_origin#"$host_url/t"}
+    warn "post-investigate $id $old_name origin_job_id=$origin_job_id"
+
+
```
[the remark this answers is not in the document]
perlpunk: That was just debugging

openqa-investigate
perlpunk: That was just debugging

openqa-investigate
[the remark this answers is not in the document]
Martchus: And I suppose I'm the author of `sync_via_investigation_comment` :-)
So I've created a PR to improve it: https://github.com/os-autoinst/scripts/pull/224

openqa-investigate
perlpunk: Hm, shellcheck doesn't complain here, but I changed it

okurz (commented):
+1

perlpunk:
I updated the screenshot with a new comment I just created on osd.

openqa-investigate
```diff
@@ -182,18 +182,82 @@ $comment_text"
...
+    local pass_lgt='' pass_lgb='' pass_lgtb=''
+    product_issue=false
+
+    readarray -t result_lines < <(fetch-investigation-results "$origin_job_id") || return $?
+
+    for line in "${result_lines[@]}"; do
+        if [[ $line =~ ^([^|]+)\|([^|]+)\|([^|]+) ]]; then
+            investigate_type=${BASH_REMATCH[1]}
+            other_id=${BASH_REMATCH[2]}
+            result=${BASH_REMATCH[3]}
```
Martchus, line 220: It looks like `other_id` is an unused variable. What about just:
```suggestion
            investigate_type=${BASH_REMATCH[1]} result=${BASH_REMATCH[3]}
```

Martchus (commented):
For Bash code it looks quite clean. I only have a few remarks.

openqa-investigate
```diff
@@ -182,18 +182,82 @@ $comment_text"
     fi
 }
 
+fetch-investigation-results() {
+    local origin_job_id=$1
+    local state job investigate_type other_id result investigate_comment comment_lines
+
+    investigate_comment=$("${client_call[@]}" -X GET jobs/"$origin_job_id"/comments | runjq -r '[.[] | select(.text | contains("Automatic investigation jobs for job") and contains(":investigate:retry**:"))] | sort_by(.id) | first') || return $?
```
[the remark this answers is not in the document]
perlpunk: done, as well as all your other remarks (except one), thanks

openqa-investigate
```diff
@@ -182,18 +182,82 @@ $comment_text"
...
+            result="$(echo "$job" | runjq -r '.result')" || return $?
+            echo "$investigate_type|$other_id|$result"
+        fi
+    done
+}
+
+is-product-issue() {
+    local origin_job_id=$1
+    local state result investigate_type passed result_lines
+    local pass_lgt='' pass_lgb='' pass_lgtb=''
```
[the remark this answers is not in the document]
Martchus: Ah, then let's do it the shellcheck way.

openqa-investigate
```diff
@@ -182,18 +182,77 @@ $comment_text"
...
+
     comment="Investigate retry job: $host_url/t$id"
-    if [[ "$result" == passed || "$result" == softfailed ]]; then
-        comment+=" $result, likely a sporadic failure"
+    if [[ "$retry_result" == passed || "$retry_result" == softfailed ]]; then
+        comment+=" $retry_result, likely a sporadic failure"
     else
         comment+=" failed, likely not a sporadic failure"
+        local product_issue=false
+        is-product-issue "$origin_job_id" || rc=$?
```
perlpunk, line 249: Actually I think I should just return `$rc` in every non-zero case here

perlpunk: done.
And I added a `is-ok` function now.

openqa-investigate
```diff
@@ -182,18 +182,79 @@ $comment_text"
...
-        comment+=" $result, likely a sporadic failure"
+    if is-ok "$retry_result"; then
+        comment+=" $retry_result, likely a sporadic failure"
     else
         comment+=" failed, likely not a sporadic failure"
+        local product_issue=false
+        is-product-issue "$origin_job_id" || return $?
+
+        if "$product_issue"; then
+            comment+="."$'\n'"Test(s) with last good build pass, likely a product issue"
```
okurz, line 256: We use the past tense and the actual result for the tests job, I would use the same here, i.e. "passed" or "soft-failed". As alternative use "ok", as we also use within openQA where we define "ok" as "passed or soft-failed"

okurz (changes requested):
Can you explain in more detail why you wait for the retry-comment and extend that instead of it's open comment? If you wanted to reduce the amount of comments and have found a good way to edit existing comments how about always updating the existing initial comment instead?

perlpunk:
> Can you explain in more detail why you wait for the retry-comment and extend that instead of it's open comment?

I don't understand that sentence at all. Can you rephrase? What waiting do you mean?
What do you mean with "it's open comment"?

> If you wanted to reduce the amount of comments and have found a good way to edit existing comments how about always updating the existing initial comment instead?

Yes, that would be possible.

okurz:
> > Can you explain in more detail why you wait for the retry-comment and extend that instead of it's open comment?
> 
> I don't understand that sentence at all. Can you rephrase? What waiting do you mean? What do you mean with "it's open comment"?

Sorry. I meant: Please explain why you add a message to the comment from the "retry" investigation job instead of creating a new comment with just your information about potential product regression

perlpunk:
> Sorry. I meant: Please explain why you add a message to the comment from the "retry" investigation job instead of creating a new comment with just your information about potential product regression

Because I'm doing that at the same time anyway. Posting two different comments didn't even occur to me.

At the time I know that the simple retry test failed, I also investigate the other investigation tests, and then I post a comment with that combined information. I have this information at the same time. It wouldn't make sense to me to just split these two sentences into two comments.

We also don't split the first investigation test list into four separate comments.

Maybe we can clarify this tomorrow in a chat.

openqa-investigate
okurz: "Jobs including the last good build are ok, likely a product issue"

okurz (approved):

PR #225, opened 2023-05-02
Fix checking for unknown issues

perlpunk:
Issue: https://progress.opensuse.org/issues/128405

The output for unown issues was changed to "Unknown test issue", but the test and the READE have never been adapted, mainly because we mock the function in our test, and the mock still uses the old string.

I will think about a way to mock less.


Martchus (approved):
`s/READE/README/`

mimi1vx (approved):

PR #230, opened 2023-05-08
Fix matching of urls

perlpunk:
It might be http or https

Otherwise we might get more unwanted output from the clone command in the
comment.

Issue: https://progress.opensuse.org/issues/128909 , https://progress.opensuse.org/issues/128405

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -63,7 +63,7 @@ clone() {
     [[ $dry_run = 1 ]] && echo "$out"
     # output is in $out, should change that text to a markdown list entry
     # Remove any text before the url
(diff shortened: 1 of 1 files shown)


kalikiana (approved):

Martchus (approved):
Note that `openqa-clone-job` supports `--json-output`.

perlpunk:
> Note that `openqa-clone-job` supports `--json-output`.

Ah, that would have been my next step. I didn't know that. It's also not documented...
I hope the warning will not end up on stdout then, otherwise the JSON will be broken...

PR #232, opened 2023-05-11
Capture and output stderr in runcli seperately

perlpunk:
Issue: https://progress.opensuse.org/issues/128909

### DIFF
--- a/_common
+++ b/_common
@@ -31,13 +31,20 @@ job_ids() {
 # an error, and the calling command and line
 runcli() {
     local args=("$@")
-    local rc output
+    local rc output errorlog errstr=
     set +e
-    output=$("${args[@]}" 2>&1)
+    errorlog=$(mktemp -t runcli-errorlog-XXXX)
+    # shellcheck disable=SC2064
+    trap "rm $errorlog" EXIT
+    output=$("${args[@]}" 2>"$errorlog")
     rc=$?
     set -e
+    if [[ -s "$errorlog" ]]; then
+        read -r errstr < "$errorlog"
+        warn "$1 ($(caller)): (${args[*]}) stderr: >>>$errstr<<<"
+    fi
     [[ "$rc" == 0 ]] && echo "$output" && return
-    warn "openqa-cli ($(caller)): (${args[*]}) rc: $rc >>>$output<<<"
+    warn "$1 ($(caller)): (${args[*]}) rc: $rc >>>$output<<<"
     return $rc
 }
 
--- a/test/04-common.t
+++ b/test/04-common.t
@@ -2,7 +2,7 @@
 
 source test/init
 
-plan tests 12
+plan tests 13
 
 source _common
 
@@ -23,7 +23,7 @@ caller() ( builtin caller 2 )
 try runcli failure a b c
 unset -f caller
 is $rc 23 "runcli failure"
-like "$got" "test/04-common.t.*failure a b c.*oh noe" "runcli failure output"
+like "$got" "test/04-common.t.*failure a b c.*stderr.*oh noe" "runcli failure output"
 
 tw_openqa_host=foo
 get_image() {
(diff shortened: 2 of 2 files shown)


test/04-common.t
```diff
@@ -56,3 +56,11 @@ latest_published_tw_builds() {
...
 is "$rc" 1 "find_latest_published_tumbleweed_image failure (no builds)"
 has "$got" "Unable to find latest published Tumbleweed builds" "Expected error message (no builds)"
+
+somecommand() {
+    echo "STDOUT"
+    echo "STDERR" >&2
+}
+
+try runcli somecommand
+like "$got" "somecommand.*>>>STDERR<<<.*STDOUT" "somecommand stdout and stderr"
```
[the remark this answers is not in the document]
perlpunk: the runcli function was previously merging both outputs, and openqa-clone-job is called with this wrapper in openqa-investigate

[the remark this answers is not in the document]
Martchus: I thought the problem wasn't `openqa-clone-job` mixing the output but this helper mixing it up later. So to me it makes sense to adjust the helper like this.

_common
```diff
@@ -31,13 +31,19 @@ job_ids() {
 # an error, and the calling command and line
 runcli() {
     local args=("$@")
-    local rc output
+    local rc output errorlog errstr=""
```
Martchus, line 34:
```suggestion
    local rc output errorlog errstr=
```

Martchus (approved):

_common
perlpunk: done

_common
```diff
@@ -31,13 +31,19 @@ job_ids() {
 # an error, and the calling command and line
 runcli() {
     local args=("$@")
-    local rc output
+    local rc output errorlog errstr=
     set +e
-    output=$("${args[@]}" 2>&1)
+    errorlog=$(mktemp -t runcli-errorlog-XXXX)
```
okurz, line 36: if you use a temporary file you should still ensure it's deletion, e.g. with

```
trap "rm $errorlog" EXIT
```

okurz (changes requested):

_common
perlpunk: done.
I just realized that running the test suite (`test/02-investigate.t`) leaves several tempfiles around, not named specifically (e.g. `/tmp/tmp.1BrG2J6Atc`). Not critical but I would like to know why they aren't deleted

perlpunk: Looks like it's from 
```
# Make auto-deleting temp files for testing subprocesses:
+fs:mktemp; comment_1234_updated=$temp
...
```
so I'l poke @ingydotnet about it

test/04-common.t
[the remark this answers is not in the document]
perlpunk: @kalikiana is it clear now?

_common
ingydotnet: Looking into it.

ingydotnet: I think I know the problem. 
The bashplus function `+fs:mktemp` sets a trap to auto-remove it. 
But in bash you can only have one trap per signal per process.
Bashplus also has a `+trap` function that allows you to have multiple traps.
But if you use a regular `trap` after `+trap` has been used it will blow it away.
Can you try changing the `trap` you just added to `+trap`?
I think that would solve the issue.

Bashplus could probably wrap the `trap` function to warn about this, but I'd need to think more about it.

_common
```diff
@@ -31,13 +31,20 @@ job_ids() {
...
```
ingydotnet, line 36: Also you could change:
```
errorlog=$(mktemp -t runcli-errorlog-XXXX)
```
to 
```
+mktemp; errorlog=$temp
```
and that would make an auto-deleting temp file for you.

ingydotnet: @perlpunk indicated that you don't use bpan/bashplus in deployment.

I made this standalone version of +trap for you to use: https://gist.github.com/ingydotnet/15a44654776f1709f04e67b1d35a307c

test/04-common.t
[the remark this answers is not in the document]
kalikiana: Ah. Yes. The >>> takes a bit of getting used to.

kalikiana (approved):

PR #233, opened 2023-06-13
Add ability to trigger full opensuse test in trigger-openqa_in_openqa

mimi1vx:
https://progress.opensuse.org/issues/127622

### DIFF
--- a/trigger-openqa_in_openqa
+++ b/trigger-openqa_in_openqa
@@ -20,6 +20,7 @@ arch="${arch:-"x86_64"}"
 machine="${machine:-"64bit"}"
 build_tag=${BUILD_TAG:-}
 client_prefix=${client_prefix:-}
+full_run=${FULL:-}
 
 main() {
     [ "$dry_run" = "1" ] && client_prefix="echo"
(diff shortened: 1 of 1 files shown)


trigger-openqa_in_openqa
```diff
@@ -46,14 +47,18 @@ trigger() {
...
+    fi
 
     ${client_prefix} "${openqa_cli}" \
         api --host "${target_host_proto}://${target_host}" -X POST isos \
         VERSION=Tumbleweed \
         DISTRI=openQA FLAVOR=dev BUILD="${build}" \
         ARCH="${arch}" BACKEND=qemu WORKER_CLASS=qemu_x86_64 \
         HDD_1="$qcow" \
-        ${ARGS} \
+        "${ARGS}" \
```
Martchus, line 61: I don't think quoting will work as the key-value-pairs actually need to be passed as separate arguments. Maybe it makes sense to make `ARGS` a bash array. Then the quoting would make sense.

mimi1vx: why?  it passes and parses the whole cli as a string

mimi1vx: + without quotes is ShellCheck unhappy

Martchus: > without quotes is ShellCheck unhappy

That's true.

> why? it passes and parses the whole cli as a string

If it works then fine :-)

Martchus:
It would be good to add a ticket reference to the commit message.

trigger-openqa_in_openqa
kalikiana: How do we know it works? Do we have tests? If we do, why don't they fail since they were not modified in this PR 🤐😉

mimi1vx:
@Martchus you were right ... it doesn't work as excepted ...

PR #235, opened 2023-06-14
BUILD magic in trigger-openqa_in_openqa isn't needed

mimi1vx:
Without it jobs have name 'TW.full.<nr>' so suffix .full isn't needed and makes build looks weird


Martchus:
Our normal style for commit messages would be "Remove not needed BUILD magic in trigger-openqa_in_openqa".

okurz (approved):

PR #236, opened 2023-06-26
openqa-trigger-bisect-jobs: Use user-agent for py-requests and skip passed jobs

asdil12:
Ticket: https://progress.opensuse.org/issues/131204

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -185,6 +185,52 @@ def find_changed_issues(investigation):
 
 def main(args):
     parsed_url = urlparse(args.url)
+    test_url = urlparse(
+        urlunparse(
+            (
+                parsed_url.scheme,
+                parsed_url.netloc,
+                "/api/v1/jobs/" + parsed_url.path.lstrip("/tests/"),
+                "",
+                "",
+                "",
+            )
+        )
+    )
+    log.debug("Retrieving job data from %s" % test_url.geturl())
+    test_data = fetch_url(test_url, request_type="json")
+    job = test_data["job"]
+    if job['result'] == 'passed':
+        log.info(
+            "Job %d (%s) is passed, skipping bisection"
+            % (job["id"], job["test"])
+        )
+        return
+    search = re.search(":investigate:", job["test"])
+    if search:
+        log.info(
+            "Job %d (%s) is already an investigation, skipping bisection"
+            % (job["id"], job["test"])
+        )
+        return
+    if job.get("clone_id") is not None:
+        log.info("Job %d already has a clone, skipping bisection" % job["id"])
+        return
+
+    children = job["children"] if "children" in job else []
+    parents = job["parents"] if "parents" in job else []
+    if (
+        "Parallel" in children
+        and len(children["Parallel"])
+        or "Directly chained" in children
+        and len(children["Directly chained"])
+        or "Parallel" in parents
+        and len(parents["Parallel"])
+        or "Directly chained" in parents
+        and len(parents["Directly chained"])
+    ):
+        return
+
     investigation_url = urlparse(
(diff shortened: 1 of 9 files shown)


openqa-trigger-bisect-jobs
```diff
@@ -183,6 +185,23 @@ def find_changed_issues(investigation):
...
+                parsed_url.netloc,
+                '/api/v1/jobs/' + parsed_url.path.rsplit('/', 1)[1],
+                "",
+                "",
+                "",
+            )
+        )
+    )
+    log.debug("Retrieving result info from %s" % result_url.geturl())
+    result = fetch_url(result_url, request_type="json")
```
perlpunk, line 201: The job data is already fetched, see line 208 in the current version.
So I suggest to not do that twice and just move the code up.

perlpunk (changes requested):
I would prefer to have the two changes in separate commits.

okurz (changes requested):
Having the keyword "and" in a git commit message is a good sign that the commit wants to be split :)

openqa-trigger-bisect-jobs
```diff
@@ -183,6 +185,23 @@ def find_changed_issues(investigation):
 
 def main(args):
     parsed_url = urlparse(args.url)
+    result_url = urlparse(
+        urlunparse(
```
Martchus, line 189: Do we really need this ridiculous combination of `urlunparse` and `urlparse`?

Martchus (changes requested):

tests/test_trigger_bisect_jobs.py
```diff
@@ -197,6 +200,13 @@ def test_problems():
     openqa.main(args)
     openqa.openqa_clone.assert_not_called()
 
+    args.url = "http://openqa.opensuse.org/tests/101"
+    try:
+        openqa.main(args)
+    except FileNotFoundError as e:
+        if str(e) == "[Errno 2] No such file or directory: 'tests/data/python-requests/openqa.opensuse.org/tests/101/investigation_ajax'":
+            raise Exception('This test should have stopped due to passed job result before hitting this issue')
```
Martchus, line 208: If I encountered this error message I would likely wonder what "this issue" is referring to.

tests/test_trigger_bisect_jobs.py
```diff
@@ -197,6 +200,13 @@ def test_problems():
     openqa.main(args)
     openqa.openqa_clone.assert_not_called()
 
+    args.url = "http://openqa.opensuse.org/tests/101"
+    try:
+        openqa.main(args)
+    except FileNotFoundError as e:
+        if str(e) == "[Errno 2] No such file or directory: 'tests/data/python-requests/openqa.opensuse.org/tests/101/investigation_ajax'":
```
Martchus, line 207: Since this is Python and not Perl we can likely do better than comparing the error message on string level. Maybe one can access the path somehow directly?

tests/test_trigger_bisect_jobs.py
```diff
@@ -197,6 +200,13 @@ def test_problems():
     openqa.main(args)
     openqa.openqa_clone.assert_not_called()
 
+    args.url = "http://openqa.opensuse.org/tests/101"
+    try:
+        openqa.main(args)
+    except FileNotFoundError as e:
```
perlpunk, line 206: I don't think we need to test an exception coming from a mocked function.
It should never occur, and if it does, then someone forgot to add the file to the test data.
So I don't see what's the motivation for this test.

asdil12: The motivation is checking if the script returns if the job result is "passed".

okurz (approved):

perlpunk (approved):

PR #237, opened 2023-06-27
Adapt bisect script to new openqa-clone-job output

Martchus:
* Use `--json-output` instead of trying to parse the human readable output
* See https://progress.opensuse.org/issues/131486


perlpunk (approved):

okurz (approved):
Ok, a real love example on openQA would be nice

okurz:
> Ok, a real love example on openQA would be nice

I wanted to write "live example" but I love "love example" anyway :)

PR #239, opened 2023-06-29
Use a unique tempfile template in openqa-label-known-issues

perlpunk:
It helps investigating in case of a problem.

Issue: https://progress.opensuse.org/issues/131465


baierjan (approved):
Quite a long name for a tempfile 😄

PR #241, opened 2023-07-03
Fix test in autosubmit

mimi1vx:
Looks like this script is trying generate Req changelog also for other packages than os-autoinst-distri-opensuse-deps

https://progress.opensuse.org/issues/132125


Martchus:
I'm also wondering why the old conditions were broken. Note that with `[[` you don't need to quote the left hand side.

okurz (approved):

baierjan (approved):

PR #242, opened 2023-07-03
Commit container packages with '--no-service'

mimi1vx:
To prevent checking non-existing spec file

https://progress.opensuse.org/issues/132125


baierjan (approved):
That's cool idea, yes.

mimi1vx:
@baierjan it works ....

Martchus (approved):

PR #244, opened 2023-07-04
Make sure exit code gets propagated to caller

perlpunk:
readarray -t foo < <(some command)

will not get the exit code of 'some command'.

Issue: https://progress.opensuse.org/issues/109920


okurz (approved):

perlpunk:
Actually I also needed to use `mapfile` in another place where `< <()` was used. pushed

Martchus (approved):

PR #245, opened 2023-07-05
Fix error handling with trap

perlpunk:
trap should only be called once, as it overwrites the previous error handler. The cleanup of the file is now done in the existing error-handler function.

Issue: https://progress.opensuse.org/issues/131465

### DIFF
--- a/_common
+++ b/_common
@@ -3,6 +3,8 @@
 # Common shell script snippets to be used when interacting with openQA
 # instances, for example over openqa-cli.
 
+errorlog=""
+
 warn() { printf '%s\n' "$@" >&2; }
 
 enable_force_result=${enable_force_result:-false}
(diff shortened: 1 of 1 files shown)


_common
```diff
@@ -35,14 +38,14 @@ runcli() {
...
     # shellcheck disable=SC2064
-    trap "rm $errorlog" EXIT
     output=$("${args[@]}" 2>"$errorlog")
     rc=$?
     set -e
     if [[ -s "$errorlog" ]]; then
         read -r errstr < "$errorlog"
         warn "$1 ($(caller)): (${args[*]}) stderr: >>>$errstr<<<"
     fi
+    rm "$errorlog"
```
Martchus, line 48: Shouldn't we clear `$errorlog` if this command succeeds?

```suggestion
    rm "$errorlog" && errorlog=
```

Martchus (commented):
This makes generally sense. I only have one remark.

okurz (approved):

_common
Martchus: Your explanation makes sense so I suppose you can just leave it as-is.

Martchus (approved):

PR #246, opened 2023-07-06
Consider only direct retry job for investigation

perlpunk:
As multiple jobs in a parallel cluster can get the same OPENQA_INVESTIGATE_ORIGIN as the direct retry job, we need to compare the original name.

Also make tests a bit more readable and structure job ids by type.

Issue: https://progress.opensuse.org/issues/132332

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -241,14 +241,18 @@ is-product-issue() {
 }
 
 post-investigate() {
-    local id=$1 old_name=$2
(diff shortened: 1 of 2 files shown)


Martchus (approved):

perlpunk:
Actually I had forgotten a test for the new code. Added now.

kalikiana (approved):
Nice!

PR #248, opened 2023-07-17
Let curl output an error message on error

perlpunk:
Normally we want it to be silent to not have any informational output we don't need, but in error cases, for example the host is not reachable, we want to see the error message.

Issue: https://progress.opensuse.org/issues/132665

### DIFF
--- a/_common
+++ b/_common
@@ -215,7 +215,7 @@ $body
 
 latest_published_tw_builds() {
     local latest_builds group_id=$1
-    latest_builds=("$(runcurl -s "$tw_openqa_host/group_overview/$group_id.json" \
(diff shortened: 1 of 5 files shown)


_common
```diff
@@ -215,7 +215,7 @@ $body
 
 latest_published_tw_builds() {
     local latest_builds group_id=$1
-    latest_builds=("$(runcurl -s "$tw_openqa_host/group_overview/$group_id.json" \
+    latest_builds=("$(runcurl -s -S "$tw_openqa_host/group_overview/$group_id.json" \
```
okurz, line 218: I commonly see (and use) `curl -sS` so I would use the shorter form here as well

okurz (changes requested):

okurz (approved):

PR #251, opened 2023-08-02
Add o3 ipxe scripts

asdil12:
https://progress.opensuse.org/issues/133160

### DIFF
--- a/ipxe/boot.ipxe
+++ b/ipxe/boot.ipxe
@@ -1,0 +1,13 @@
+#!ipxe
+
+# This script is supposed to be embedded into the ipxe.pxe and ipxe.efi bootloader images like this:
+# git clone https://github.com/ipxe/ipxe.git
+# cd ipxe/src
+# make EMBED=boot.ipxe bin/ipxe.pxe bin-x86_64-efi/ipxe.efi
+
+echo Welcome to o3 iPXE
+
+dhcp
+
+chain --replace --autofree http://${next-server}/tftpboot/ipxe/menu.ipxe ||
+chain --replace --autofree tftp://${next-server}/ipxe/menu.ipxe ||
--- a/ipxe/menu.ipxe
+++ b/ipxe/menu.ipxe
@@ -1,0 +1,128 @@
+#!ipxe
+
+# dhcp should be already done by boot script
+#dhcp
+
+#set variables
+set menu-timeout 30000
+
+:start
+menu iPXE boot menu for openqa.opensuse.org 
+
+item --gap --             ------------------------- Operating systems ------------------------------
+item --gap --             ------ (You will be prompted to edit cmdline after selecting an os) ------
+item --key l leapS1       (L)Boot openSUSE Leap15.5 from download.opensuse.org / ttyS1 (http)
+item --key e leapS2       (E)Boot openSUSE Leap15.5 from download.opensuse.org / ttyS2 (http)
+item --key a leapS1auto   (A)Boot openSUSE Leap15.5 / ttyS1 autoyast=https://is.gd/oqaay4
+item --key k tumbleweed_KVM_autoyast (K)Boot openSUSE Tumbleweed KVM server from O3 repo with autoyast/ ttyS1 (http)
+item --key z tumbleweed_XEN_autoyast (Z)Boot openSUSE Tumbleweed XEN server from O3 repo with autoyast/ ttyS1 (http)
+item --key t tumbleweedS1 (T)Boot openSUSE Tumbleweed from download.opensuse.org / ttyS1 (http)
+item --key s tumbleweedS2 (S)Boot openSUSE Tumbleweed from download.opensuse.org / ttyS2 (http)
+item --key M memtest64efi (M)Memtest 64 EFI
+item --key N nbxyzbios    (N)Netboot.xyz BIOS
+item --key N nbxyzefi     (N)Netboot.xyz EFI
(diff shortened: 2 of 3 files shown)


ipxe/boot.ipxe
```diff
@@ -0,0 +1,13 @@
+#!ipxe
+
+# this script is to be embedded into the ipxe.pxe and ipxe.efi bootloader imges
```
Martchus, line 3:
```suggestion
# This script is supposed to be embedded into the ipxe.pxe and ipxe.efi bootloader images like this:
```

ipxe/menu.ipxe
```diff
@@ -0,0 +1,121 @@
...
+goto ${selected}
+
+:leapS1
+set kernel http://download.opensuse.org/distribution/leap/15.3/repo/oss/boot/x86_64/loader/linux
+set cmdline usessh=1 sshpassword=linux network=1 install=http://download.opensuse.org/distribution/leap/15.3/repo/oss/ console=ttyS1,115200n8 root=/dev/ram0 initrd=initrd textmode=1
+set initrd http://download.opensuse.org/distribution/leap/15.3/repo/oss/boot/x86_64/loader/initrd
+goto editandboot
+
+:leapS2
+set kernel http://download.opensuse.org/distribution/leap/15.3/repo/oss/boot/x86_64/loader/linux
```
Martchus, line 41: I guess you just coped that over from ariel. Maybe it makes nevertheless sense to also add entries for Leap 15.4 and 15.5 in addition. I'm not sure whether any tests rely on the order of entries here so maybe it would make sense to append these entries (instead of inserting them before Tumbleweed).

ipxe/pxeboot.conf
```diff
@@ -0,0 +1,38 @@
...
+dhcp-option=tag:efi-x86_64-http,option:bootfile-name,http://10.150.1.11/tftpboot/ipxe/ipxe.efi
+dhcp-option=tag:efi-x86_64-http,option:vendor-class,HTTPClient
+
+# UEFI ipxe binary from Arch Linux' ipxe package, useful for testing
+#dhcp-boot=tag:efi-x86_64,arch-netboot/ipxe-arch.16e24bec1a7c.efi
+
+# UEFI ipxe binnary built that loads commands from http://martchus.no-ip.biz/ipxe/leap-15.4 to install Leap 15.4 with autoyast
+# note: This image has been used to install Leap 15.4 on ow19/ow20, see https://progress.opensuse.org/issues/111473#note-44 for
+#       details.
+#dhcp-boot=tag:efi-x86_64,ipxe-own-build/ipxe.efi
```
Martchus, line 36: I could live with having this disabled code from my previous experiments and setup work removed. So we should likely not add them to Git.

Martchus (changes requested):

Martchus (approved):
As mentioned it would likely be safer to append the new entries instead of inserting them before Tumbleweed. However, we can also just try whether it breaks tests or not.

asdil12:
> As mentioned it would likely be safer to append the new entries instead of inserting them before Tumbleweed. However, we can also just try whether it breaks tests or not.

do you happen to have a reference to a test that actually uses this?
Grepping through os-autoinst-distri-opensuse it seems that tests that perform PXE seem to bring their own supportserver.

So I would simply see if anything breaks.

Martchus:
Maybe no tests are using this anymore. I just remember that last time I temporarily broke the setup people were complaining :-)

nicksinger:
If tests break due to a changed order I think the test is broken and should be fixed instead. After all, a human would also just select what it needs and not "the second entry" ;)

ipxe/ipxe.pxe
nicksinger, line 1: I don't think binary files fit into that repo. If we want to automatically create them based on our source this should be done by a pipeline and if they are external (like memtest64.efi) they don't belong into our repo at all and rather into a package of some sort.

nicksinger (changes requested):

ipxe/ipxe.pxe
Martchus: You could also just manually upload the binaries as assets in the release section. I don't think we have any releases so far but I had nothing against if you'd just made one for that. Maybe the asset should be a tar file containing the entire relevant directory tree and there should be a script to generate that tar file.

nicksinger: might be an improvement for later but for now I'd just throw out all binaries :)

Martchus (approved):

ipxe/menu.ipxe
```diff
@@ -0,0 +1,135 @@
...
+#set variables
+set menu-timeout 30000
+
+:start
+menu iPXE boot menu for openqa.opensuse.org 
+
+item --gap --             ------------------------- Operating systems ------------------------------
+item --gap --             ------ (You will be prompted to edit cmdline after selecting an os) ------
+item --key l leapS1       (L)Boot openSUSE Leap15.3 from download.opensuse.org / ttyS1 (http)
+item --key e leapS2       (E)Boot openSUSE Leap15.3 from download.opensuse.org / ttyS2 (http)
```
mimi1vx, line 15: why 15.3 which is EOL ?

asdil12: I just kept it, because I reused the existing menu partly.

asdil12: removed the 15.3 entries and added an autoyast entry

nicksinger (approved):

PR #261, opened 2023-09-21
openqa-advanced-retrigger-jobs: Post an optional comment on matching …

kalikiana:
When re-triggering it's often desirable to also add ticket/bug labels.

See: https://progress.opensuse.org/issues/135407#note-10


okurz:
Neat idea. Can you provide an output of example execution?

Martchus (approved):

kalikiana:
> Neat idea. Can you provide an output of example execution?

Sorry, I missed your comment. Here's an example from today:

    for i in worker{33..36}; do env host=openqa.suse.de result="result='failed'" comment='Re-running to verify connection with poo#138698' ./openqa-advanced-retrigger-jobs; done

image

For example https://openqa.suse.de/tests/12749636#comments

okurz (approved):

PR #263, opened 2023-09-28
Mention when force result will not be applied

kalikiana:
To distinguish from other cases where the result was not applied and avoid having to second-guess after the fact.

See: https://progress.opensuse.org/issues/135782

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -15,6 +15,7 @@ dry_run="${dry_run:-"0"}"
 min_search_term="${min_search_term:-"16"}"
 issue_marker="${issue_marker:-"auto_review%3A"}"
 issue_query="${issue_query:-"https://progress.opensuse.org/projects/openqav3/issues.json?limit=200&subproject_id=*&subject=~${issue_marker}"}"
+force_result_tracker="${force_result_tracker:-"openqa-force-result"}"
 reason_min_length="${reason_min_length:-"8"}"
 grep_timeout="${grep_timeout:-5}"
 email_unreviewed=${email_unreviewed:-false}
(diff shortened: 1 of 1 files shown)


openqa-label-known-issues
```diff
@@ -81,8 +82,12 @@ label_on_issues_from_issue_tracker() {
         force_result=''
         label="poo#$issue $subject"
         if [[ ${#search} -ge $min_search_term ]]; then
-            if [[ $tracker == 'openqa-force-result' && $after =~ :force_result:([a-z_]+) ]]; then
-                force_result=${BASH_REMATCH[1]}
+            if [[ $after =~ :force_result:([a-z_]+) ]]; then
+                if [[ $tracker == "$force_result_tracker" ]]; then
+                    force_result=${BASH_REMATCH[1]}
+                else
+                    label="$label (Ignoring force result for ticket which is not in $force_result_tracker)"
```
perlpunk, line 89: Maybe add the word "tracker" explicitly because most people don't know

Martchus, line 89:
```suggestion
                    label="$label (ignoring force result for ticket which is not in tracker \"$force_result_tracker\")"
```

okurz (changes requested):
Neat idea. Can you reference examples at best from o3 showing the new case?

Martchus (approved):

kalikiana:
> Neat idea. Can you reference examples at best from o3 showing the new case?

Screenshot 2023-10-02 at 15 39 04

See https://openqa.opensuse.org/tests/3583961#comments

perlpunk (approved):

okurz (approved):

PR #266, opened 2023-10-17
Consistently use the term unreviewed issue in all contexts

kalikiana:
We should not talk about unknown and unreviewed issues in different places to mean the same thing.

The docs also don't explain the actual email subject.

See: https://progress.opensuse.org/issues/132437

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ expensive when `(?s)` is used. Every `.*` can span over multiple lines and
 that involves a lot of backtracking and might not be needed. Use `[^\n]*` over
 `.*` where possible or use line spanning matches like `[\S\s]*`.
 
-### Unknown issues
+### Unreviewed issues
 
 If none of your configured searches above are matching, you can configure a
 notification for those failed jobs, for example to an email address or a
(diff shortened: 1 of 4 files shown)


Martchus:
And the "unreviewed" in the commit message lacks the "ed".

README.md
```diff
@@ -111,9 +111,9 @@ there is no carry-over and no automatic ticket assignment by auto-review.
 
 ### Combine auto-review and openqa-investigate
 
-A possible approach to combine handling known issues and unknown issues is to
+A possible approach to combine handling known issues and unreviewed issues is to
```
okurz, line 114: this is where it reads less nice than in before. In before we had "known and unknown" issues, not it's "known and unreviewed". Makes me think, should we use just "unknown" issues consistently everywhere and leave that aspect of "review" which is more like a SUSE internal convention what it means out of here?

okurz (changes requested):

README.md
kalikiana: SUSE internal? To my mind unreviewed just means nobody has looked at the issue. I don't see a difference between unknown and unreviewed, though, so like I said, feel free to convince others and I will change everything to use "unknown".

okurz (approved):

PR #268, opened 2023-10-20
Fix warnings about empty variables

perlpunk:
We saw warnings in the logs:

    /opt/os-autoinst-scripts/openqa-investigate: line 239: : command not found
    /opt/os-autoinst-scripts/openqa-investigate: line 241: : command not found

By testing if the variables have content this can be fixed. Not sure how I would make such warnings fatal in our tests.

Issue: https://progress.opensuse.org/issues/132272

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -236,9 +236,9 @@ identify-issue-type() {
             fi
         fi
     done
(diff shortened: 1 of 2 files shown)


Martchus (approved):

kalikiana (approved):
Good find!

PR #269, opened 2023-10-23
Add simple script to check linking reason for package

baierjan:
Check for comments in each package and look for reason to linking. Outputs the list of packages without a reason.

Reference: https://progress.opensuse.org/issues/128087

### DIFF
--- a/obs-check-devel-openqa-leap-extra-dependencies
+++ b/obs-check-devel-openqa-leap-extra-dependencies
@@ -1,0 +1,14 @@
+#!/bin/bash -e
+
+osc="osc --apiurl https://api.opensuse.org"
+
+for project in $($osc search --project -s 'devel:openQA:Leap:' | grep -o '^devel:openQA:Leap:.*'); do
+    for package in $($osc list "$project"); do
+        $osc api "/comments/package/$project/$package" | \
+            yq -pxml -oy '.comments.comment |= ([] + .) | .comments.comment[].+content' | \
+            grep -q "Reason for linking:" || \
+            { echo "No reason for $project/$package" >&2; problem=1; }
+    done
+done
+
+test -z "$problem"
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -19,6 +19,7 @@ main_requires:
   curl:
   grep:
   jq:
+  yq:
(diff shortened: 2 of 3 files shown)


obs-check-devel-openqa-leap-extra-dependencies
```diff
@@ -0,0 +1,15 @@
...
+    for package in $($osc list "$project"); do
+        $osc api "/comments/package/$project/$package" | \
+            yq -pxml -oy '.comments.comment |= ([] + .) | .comments.comment[].+content' | \
+            grep -q "Reason for linking:" || \
+            { echo "No reason for $project/$package" >&2; problem=1; }
+    done
+done
+
+[ -n "$problem" ] && exit 1
+exit 0
```
okurz, line 15: That's equivalent to just

```
test -z "$problem"
```

isn't it?

okurz (changes requested):

obs-check-devel-openqa-leap-extra-dependencies
baierjan: yeah, right, now it is 😄

obs-check-devel-openqa-leap-extra-dependencies
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
+osc="osc --apiurl https://api.opensuse.org"
+
+for project in $($osc search --project -s 'devel:openQA:Leap:' | grep -o '^devel:openQA:Leap:.*'); do
+    for package in $($osc list "$project"); do
+        $osc api "/comments/package/$project/$package" | \
+            yq -pxml -oy '.comments.comment |= ([] + .) | .comments.comment[].+content' | \
+            grep -q "Reason for linking:" || \
+            { echo "No reason for $project/$package" >&2; problem=1; }
```
kalikiana, line 10: Where is this going to be used? And do you have some example output to show that it works?

okurz: > Where is this going to be used?

Something like a script on schedule that informs us about such inefficient use of packages

Martchus (approved):

PR #270, opened 2023-11-13
Wait longer and sleep longer

perlpunk:
When making openqa-cli requests, we try to be more patient.

https://github.com/os-autoinst/openQA/pull/5359 Will make sure also connection timeouts will be retried.

Issue: https://progress.opensuse.org/issues/138545


okurz (commented):
In the commit message s/Issue: ihttps://progress.opensuse.org/issues/138545/Issue: https://progress.opensuse.org/issues/138545/, then consider the PR approved

kalikiana (approved):

Martchus (approved):

kraih:
I would have just changed the default timeout on the `openqa-cli` level. For a command line client 30 seconds seems like a good default.

perlpunk:
yep, see #5363

PR #272, opened 2023-11-22
Make investigation comment more prominent

asdil12:
https://progress.opensuse.org/issues/138299

image

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -259,21 +259,22 @@ post-investigate() {
     # cluster jobs might have the same OPENQA_INVESTIGATE_ORIGIN as the root retry job
     [[ $retry_name != "$origin_name:investigate:retry" ]] && echo "Job $retry_name ($id) is not a retry of $origin_name ($origin_job_id)" && return 0
 
-    comment="Investigate retry job **$retry_name**: $host_url/t$id"
+    comment="Investigate retry job *$retry_name*: $host_url/t$id"
     if is-ok "$retry_result"; then
-        comment+=" $retry_result, likely a sporadic failure"
+        comment+=" $retry_result."$'\n\n'
+        comment+="📋 **Likely a sporadic failure**."
     else
-        comment+=" failed"
+        comment+=" failed."$'\n\n'
         local product_issue=false
         local test_issue=false
         identify-issue-type "$origin_job_id" || return $?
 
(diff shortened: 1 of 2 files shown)


openqa-investigate
```diff
@@ -66,9 +66,8 @@ clone() {
 
     # output: { "$id": $clone_id }
     clone_id=$(echo "$out" | runjq -r ".\"$id\"")
-    url="$scheme://$host/t$clone_id"
     # Create markdown list entry
-    echo "* **$name**: $url"
+    echo "* *$name*: t#$clone_id"
```
okurz, line 70: can you explain this change?

okurz (changes requested):
The message itself looks nice. I just wonder why in your screenshot the aggregation message was added to the "bisect jobs" comment. I don't think it belongs there. Can you explain that?

openqa-investigate
asdil12: openQA will automatically expand `t#123` to a link to that test.

okurz: Ok. I suggest you put that into a separate explicit commit if it's not too much effort. I am asking that because for me it would take like 2-3m with `git gui` in an interactive rebase :)

asdil12:
> The message itself looks nice. I just wonder why in your screenshot the aggregation message was added to the "bisect jobs" comment. I don't think it belongs there. Can you explain that?

It was just a mock to see how it looks.

Martchus (approved):
Looks good, the old version definitely over-used bold formatting. You could also drop the italics and make those just normally formatted bullet points (but I don't really care).

openqa-investigate
```diff
@@ -66,9 +66,8 @@ clone() {
 
     # output: { "$id": $clone_id }
     clone_id=$(echo "$out" | runjq -r ".\"$id\"")
-    url="$scheme://$host/t$clone_id"
     # Create markdown list entry
-    echo "* *$name*: $url"
+    echo "* *$name*: t#$clone_id"
```
perlpunk, line 70: The job status bubbles created via javascript when loading comments don't recognize this syntax, e.g. compare
https://openqa.opensuse.org/tests/3734688#comments (good) vs.
https://openqa.opensuse.org/tests/3751441#comments (bad)

PR #274, opened 2023-12-08
Decrease priority of bisect jobs in accordance with investigation jobs

Martchus:
See https://progress.opensuse.org/issues/152281

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -80,6 +80,11 @@ def parse_args():
         help="The openQA test URL for which to trigger bisection investigation jobs",
     )
     parser.add_argument(
+        "--priority-add",
+        default=100,
+        help="Adds the specified value to the cloned job's priority value",
+    )
+    parser.add_argument(
         "--dry-run", action="store_true", help="Do not do any action on openQA"
     )
     args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

openqa-trigger-bisect-jobs
```diff
@@ -121,6 +126,19 @@ def openqa_comment(job, host, comment, dry_run):
     ]
     return call(args, dry_run)
 
+def openqa_set_job_prio(job_id, host, prio, dry_run):
```
mimi1vx, line 129: I don't think this whole part is needed

okurz:
> It creates new openqa api call ?
> 
> prio shuld be set during clone call by setting `_PRIORITY=<needed_prio>` without need second api call to set different priority

that works? Well, then that sounds good :)

openqa-trigger-bisect-jobs
```diff
@@ -292,6 +311,7 @@ def main(args):
         for job_id in sorted(created_job_ids):
             log.info(f"Created {job_id}")
             created += f"* **{test_name}**: {base_url}/t{job_id}\n"
+            openqa_set_job_prio(job_id, args.url, prio, args.dry_run)
```
mimi1vx, line 314: really this isn't needed

mimi1vx (changes requested):

openqa-trigger-bisect-jobs
Martchus: Is it really possible to set the priority with `openqa-clone-job`? It doesn't seem to be possible according to `--help`.

perlpunk:
I just tried
```
openqa-clone-job --host http://localhost:9526 --from https://openqa.opensuse.org/tests/3792472 _PRIORITY=124
```
and the prio of the created job was 50...

Martchus:
I don't think you're doing anything wrong. To my knowledge this is simply not implemented in `openqa-clone-job` (or the underlying API route). Note that this would have worked as scheduling parameter but that's a whole different route.

mimi1vx:
ah, @Martchus  is right ... it works in normal scheduling but not for clone-job :( Which is pretty inconsistent..

other parameters like `_GROUP` works

PR #276, opened 2023-12-14
Set prio of all cloned jobs in openqa-investigate

Martchus:
See https://progress.opensuse.org/issues/152281#note-5

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -68,7 +68,13 @@ clone() {
     clone_id=$(echo "$out" | runjq -r ".\"$id\"")
     # Create markdown list entry
     echo "* *$name*: t#$clone_id"
-    "${client_call[@]}" --json --data "{\"priority\": $((base_prio + prio_add))}" -X PUT jobs/"$clone_id" >/dev/null
+
+    # set priority of cloned jobs
+    # shellcheck disable=SC2207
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

openqa-investigate
```diff
@@ -68,7 +68,13 @@ clone() {
     clone_id=$(echo "$out" | runjq -r ".\"$id\"")
     # Create markdown list entry
     echo "* *$name*: t#$clone_id"
-    "${client_call[@]}" --json --data "{\"priority\": $((base_prio + prio_add))}" -X PUT jobs/"$clone_id" >/dev/null
+
+    # set priority of cloned jobs
+    # shellcheck disable=SC2207
+    clone_ids=($(echo "$out" | runjq -r 'values[]'))
+    for id in "${clone_ids[@]}"; do
+        echo "${client_call[@]}" --json --data "{\"priority\": $((base_prio + prio_add))}" -X PUT jobs/"$id" >/dev/null
```
nicksinger, line 76: did you forget to remove the echo before the command call?

nicksinger (changes requested):
see inline comment

openqa-investigate
kalikiana: There is `dry_run` which you may want to use here, too

okurz (approved):

PR #281, opened 2024-01-18
Replace spurt with Mojo::File::spew

b10n1k:
Pipeline shows a warning on xt/01-make-update-deps.t for deprecate use of Mojo::File::spurt. Fix it replacing with Mojo::File::spew


perlpunk (changes requested):
For changes under external/ please always change in the corresponding git subrepo first:
https://github.com/os-autoinst/os-autoinst-common
And then pull the changes in with `git subrepo pull`.
We should do a pull right now anyway, there have been changes.

okurz (changes requested):
Files in external/ should not be changed here but in https://github.com/os-autoinst/os-autoinst-common and then updated using git-subrepo. We might already have those changes prepared but not updated yet

PR #284, opened 2024-01-22
Enhance CI with whitespace checks on bash scripts

b10n1k:
Implement `tab_checker` to ensure non-tab intentation usage on the bash scripts. The scripts runs only against the committed files of the PR using github REST API.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,3 +26,6 @@ jobs:
         run: |
          sudo apt-get install yamllint shellcheck
          make checkstyle
+         tools/tab_checker "$PR_NUMBER"
+        env:
+	 PR_NUMBER: ${{ github.event.number }}
--- a/tools/tab_checker
+++ b/tools/tab_checker
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+pr_id=$1
+files=$(curl https://api.github.com/repos/os-autoinst/scripts/pulls/${pr_id}/files | jq -c '.[].filename' | sed 's/\"//g')
+checker() {
+for n in "$files"; do
+    if head -1 "$n" | grep -q '^#!'; then
+       if [[ $(sed -n '/\t/p' "$n" | wc -l) > 0 ]]; then
+          echo "$n is using TAB"
+       	  return 1
+       fi
+    return 0
+fi
+done
+}
+
+checker "$@"


tools/tab_checker
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+pr_id=$1
+files=$(curl https://api.github.com/repos/os-autoinst/scripts/pulls/${pr_id}/files | jq -c '.[].filename' | sed 's/\"//g')
+checker() {
+for n in "$files"; do
+    if head -1 "$n" | grep -q '^#!'; then
+       if [[ $(sed -n '/\t/p' "$n" | wc -l) > 0 ]]; then
```
kalikiana, line 8: If we're using this solution I'd like an in-place version that updates the files, either always or as an option

.github/workflows/ci.yml
```diff
@@ -26,3 +26,6 @@ jobs:
         run: |
          sudo apt-get install yamllint shellcheck
          make checkstyle
+         tools/tab_checker "$PR_NUMBER"
```
okurz, line 29: This should simply be part of `make checkstyle`

tools/tab_checker
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
```
okurz, line 2:
```suggestion
set -euo pipefail
```

is a good common practice

okurz (changes requested):
I see multiple issues, first is that likely we shouldn't do our own scripting but rely on upstream tooling as discussed in https://progress.opensuse.org/issues/153793 . Second I assume os-autoinst-common is a better place than scripts as we would benefit from inheriting this script into all common "os-autoinst" repos if we would go that route at all. Then there general design is a good idea but I would not make relying on github mandatory.

perlpunk:
> The scripts runs only against the committed files of the PR using github REST API.

I would be interested in the reason of doing this instead of all files.

tools/tab_checker
```diff
@@ -0,0 +1,17 @@
...
+pr_id=$1
+files=$(curl https://api.github.com/repos/os-autoinst/scripts/pulls/${pr_id}/files | jq -c '.[].filename' | sed 's/\"//g')
```
baierjan, line 4: You probably want to silent the curl and get rid of the `sed`, because that will delete also quotes in the middle.
```suggestion
files=$(curl -s https://api.github.com/repos/os-autoinst/scripts/pulls/${pr_id}/files | jq -r -c '.[].filename')
```

tools/tab_checker
```diff
@@ -0,0 +1,17 @@
...
+checker() {
+for n in "$files"; do
+    if head -1 "$n" | grep -q '^#!'; then
```
baierjan, line 7: I am not sure if this is enough for the "on bash scripts" part. This will also match other scripts. Maybe use `file`?

tools/tab_checker
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+pr_id=$1
+files=$(curl https://api.github.com/repos/os-autoinst/scripts/pulls/${pr_id}/files | jq -c '.[].filename' | sed 's/\"//g')
+checker() {
+for n in "$files"; do
```
baierjan, line 6: This will not work correctly. The quotes will prevent argument splitting. It would be better to use arrays.

baierjan (changes requested):

tools/tab_checker
perlpunk: right. In https://github.com/os-autoinst/openQA/blob/master/tools/perlfiles#L3 we are using `file --mime-type`

b10n1k:
> Please mention the ticket in the commit message / PR description: https://progress.opensuse.org/issues/153793
> 
> What about the idea mentioned in the ticket about using https://github.com/mvdan/sh for formatting? Was this discussed somewhere?

I also like it. I just read your comments in the tickets and i will try to provide another PR today with that tool. I am just not sure how this should behave exactly in this case.

b10n1k:
closed in favor of https://github.com/os-autoinst/scripts/pull/285

PR #285, opened 2024-01-23
Ci whitespace check shfmt

b10n1k:
Alternative approach using shfrm for bash scripts static analysis. 
First approach/related PR https://github.com/os-autoinst/scripts/pull/284
 https://progress.opensuse.org/issues/153793

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,9 +3,6 @@ indent_style = space
 indent_size = 4
 binary_next_line   = true
 
-[Makefile]
-indent_style = tab
-
 [.bpan/**]
 ignore = true
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+[*]
+indent_style = space
+indent_size = 4
+binary_next_line   = true
+
+[Makefile]
+indent_style = tab
+
+[.bpan/**]
+ignore = true
+
+[*~]
+ignore = true
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ test-online:
 checkstyle: test-shellcheck test-yaml
 
 test-shellcheck:
+	@which shfmt >/dev/null 2>&1 || echo "Command 'shfmt' not found, can not execute shell script formating checks"
 	shfmt -d .
 	@which shellcheck >/dev/null 2>&1 || echo "Command 'shellcheck' not found, can not execute shell script checks"
 	shellcheck -x $$(file --mime-type * | sed -n 's/^\(.*\):.*text\/x-shellscript.*$$/\1/p')
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ test-online:
 checkstyle: test-shellcheck test-yaml
 
 test-shellcheck:
+	shfmt -d .
 	@which shellcheck >/dev/null 2>&1 || echo "Command 'shellcheck' not found, can not execute shell script checks"
 	shellcheck -x $$(file --mime-type * | sed -n 's/^\(.*\):.*text\/x-shellscript.*$$/\1/p')
 
--- a/openqa-label-known-issues-and-investigate-hook
+++ b/openqa-label-known-issues-and-investigate-hook
@@ -5,7 +5,8 @@ set -eo pipefail
 # "hook script" intended to be called by openQA instances taking a job ID as
 # parameter and forwarding a complete job URL to "openqa-label-known-issues"
 # on stdin and all left unknowns to "openqa-investigate"
-dir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)
+dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+
 PATH=$dir:$PATH
 
 # shellcheck source=/dev/null
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,5 +24,5 @@ jobs:
       - uses: actions/checkout@v2
       - name: Static checks
         run: |
-         sudo apt-get install yamllint shellcheck
+         sudo apt-get install yamllint shellcheck shfmt
          make checkstyle
--- a/_common
+++ b/_common
@@ -43,7 +43,7 @@ runcli() {
     rc=$?
     set -e
     if [[ -s "$errorlog" ]]; then
-        read -r errstr < "$errorlog"
+        read -r errstr <"$errorlog"
         warn "$1 ($(caller)): (${args[*]}) stderr: >>>$errstr<<<"
     fi
     rm "$errorlog"
@@ -144,7 +144,7 @@ handle_unreviewed() {
 
     if "$email_unreviewed" && [[ "$group_id" != 'null' ]]; then
         group_data=$(openqa-cli "${client_args[@]}" "job_groups/$group_id")
-        group_description=$(echo "$group_data" | runjq -r '.[0].description' ) || true
+        group_description=$(echo "$group_data" | runjq -r '.[0].description') || true
         group_mailto=$(echo "$group_description" | perl -n -wE'm/.*MAILTO: (\S+).*/ and say $1' | head -1)
         group_mailto=${group_mailto:-$notification_address}
         # Avoid sending notifications for jobs that were restarted
@@ -231,7 +231,7 @@ get_image() {
 find_latest_published_tumbleweed_image() {
     local latest_builds build image=null assetname
     local tw_group_id=$1 arch=$2 machine=$3 type=$4
-    read -r -a latest_builds <<< "$(latest_published_tw_builds "$tw_group_id")"
+    read -r -a latest_builds <<<"$(latest_published_tw_builds "$tw_group_id")"
     if [[ "${#latest_builds[@]}" -eq 0 ]]; then
         warn "Unable to find latest published Tumbleweed builds."
         exit 1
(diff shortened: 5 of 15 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
+[*]
+indent_style = space
+indent_size = 4
```
perlpunk, line 3: Should we vote on the indentation? I could live with both 4 and 2, with a preference for 4 (eyes are getting worse when ageing :)

openqa-label-known-issues-and-investigate-hook
```diff
@@ -5,7 +5,10 @@ set -eo pipefail
 # "hook script" intended to be called by openQA instances taking a job ID as
 # parameter and forwarding a complete job URL to "openqa-label-known-issues"
 # on stdin and all left unknowns to "openqa-investigate"
-dir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)
+dir=$(
+    cd "$(dirname "${BASH_SOURCE[0]}")"
+    pwd
+)
```
perlpunk, line 11: I think this is the only change I really don't like, but apparently one cannot ignore certain lines

.editorconfig
Martchus: I think 4 makes most sense because it is in-line with what is use in Perl code as well and probably also what we already use in most shell files in most places.

openqa-label-known-issues-and-investigate-hook
Martchus: Yes, but I can also live with it.

Martchus (approved):

perlpunk (dismissed):
`s/shfrm/shfmt/` in git commit, title and description :)

perlpunk:
Also please mention the corresponding ticket in the commit message and description.
It might always not seem important, but it's good to get into the habit, as it can help later software archeologists to find out the broader reasons of why something was done.

b10n1k:
> Also please mention the corresponding ticket in the commit message and description. It might not always seem important, but it's good to get into the habit, as it can help later software archeologists to find out the broader reasons of why something was done.

No problem to add it on the PR but i hesitate to do so on the commit. links tend to be deprecated and in many cases others do not have access.

b10n1k:
@perlpunk Can you resolve the discussions?

.editorconfig
baierjan: having `set shiftwidth=4 tabstop=4 softtabstop=4 shiftround expandtab` in my vimrc I strongly prefer 4 😄

baierjan: Also note: we use 2 spaces in yaml files (so that should be probably also included in this configuration somehow).

openqa-label-known-issues-and-investigate-hook
baierjan: Maybe this would be better?
```suggestion
dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
```

baierjan (changes requested):
I guess the second part from https://github.com/os-autoinst/scripts/pull/284#pullrequestreview-1837403769 still applies here:
> Second I assume os-autoinst-common is a better place than scripts as we would benefit from inheriting this script into all common "os-autoinst" repos if we would go that route at all.

What do you think? I would also like to know the answer about yaml files.

perlpunk:
> No problem to add it on the PR but i hesitate to do so on the commit. links tend to be deprecated and in many cases others do not have access.

It's of course possible that links are not resolvable anymore or not accessible for some. However:
- That doesn't make it less useful to have them in the commit message in the first place
- If links to progress.o.o ever stop to work for some reason because the urls change, we might install redirects because losing the links would be a major disruption anyway
- We ask everyone to add a ticket link in the commit message
- Personally I can say it has helped me in the past to find a corresponding ticket link when going through the git history

Unless we changed the rule I would like to insist on adding it.

baierjan:
> in many cases others do not have access.

Our tickets in progress are usually publicly accessible (even without login) 😉

perlpunk:
> I guess the second part from [#284 (review)](https://github.com/os-autoinst/scripts/pull/284#pullrequestreview-1837403769) still applies here:
> 
> > Second I assume os-autoinst-common is a better place than scripts as we would benefit from inheriting this script into all common "os-autoinst" repos if we would go that route at all.

So this would be the .editorconfig. I think this would look the same for every repo, so yes, I guess it makes sense to add it there.

> What do you think?
>  I would also like to know the answer about yaml files.

What do you mean here? We merged https://github.com/os-autoinst/os-autoinst-common/pull/34 and will at some point link to the common .yamllint.

baierjan:
> > I would also like to know the answer about yaml files.
> 
> What do you mean here? We merged [os-autoinst/os-autoinst-common#34](https://github.com/os-autoinst/os-autoinst-common/pull/34) and will at some point link to the common .yamllint.

My concern is whether the `shfmt` tool will also autoformat yaml files. As I saw _[Makefile]_ section in the config, I assumed it can do more than shell scripts. You can of course disregard this question if that is not the case.

.editorconfig
```diff
@@ -0,0 +1,13 @@
...
+binary_next_line   = true
+
+[Makefile]
+indent_style = tab
```
baierjan, line 7: Maybe this is not needed because `shfmt` does not do anything to Makefiles?
```suggestion
```

openqa-label-known-issues-and-investigate-hook
perlpunk: That's worth a try

.editorconfig
b10n1k: This might be a lit misleading. But there are two different lists of files touched depending the arguments. Specifically `shfmt -l .` vs `shfmt -l $(git ls-files)`. Despite the former doesnt seem to touch Makefile, i though that it doesnt hurt to leave it in the `.editorconfig` and avoid any mistake.

b10n1k:
> > No problem to add it on the PR but i hesitate to do so on the commit. links tend to be deprecated and in many cases others do not have access.
> 
> It's of course possible that links are not resolvable anymore or not accessible for some. However:
> 
> * That doesn't make it less useful to have them in the commit message in the first place
> * If links to progress.o.o ever stop to work for some reason because the urls change, we might install redirects because losing the links would be a major disruption anyway
> * We ask everyone to add a ticket link in the commit message
> * Personally I can say it has helped me in the past to find a corresponding ticket link when going through the git history
> 
> Unless we changed the rule I would like to insist on adding it.

Done on https://github.com/os-autoinst/scripts/pull/285/commits/a9b1586ddd6476746ef93825011739f05d58ee0a

.editorconfig
perlpunk: When you pass filenames to shfmt, it will try to format everything it can, even yaml files. So that should be used only if you want to format a certain file.

Makefile
```diff
@@ -34,6 +34,7 @@ test-online:
 checkstyle: test-shellcheck test-yaml
 
 test-shellcheck:
+	shfmt -d .
```
baierjan, line 37: I just spotted this, it is only a suggestion for consistency, feel free to ignore that.

```suggestion
    @which shfmt >/dev/null 2>&1 || echo "Command 'shfmt' not found, can not execute shell script formating checks"
	shfmt -d .
```

.editorconfig
baierjan: In that case we just need to be careful to not execute it on yaml files (or add a definition which will be in sync with the yamllint, but that is probably not ideal). I am OK with that.

perlpunk: I guess my answer was misleading.
shfmt cannot understand and format YAML. It formats it, but then it will result in something different, likely even invalid YAML.
For formatting YAML I will introduce yamltidy config files (already did that in os-autoinst).

Makefile
perlpunk: We should also probably add the package to our dependencies: https://github.com/os-autoinst/scripts/blob/master/dependencies.yaml
Maybe add a new target `develop`

okurz (changes requested):
Please see the two open and still valid discussions

.editorconfig
baierjan: In that case we are back in the original question, can `shfmt` format makefiles (and by format I mean will the makefile still be a valid makefile)? If not, we should definitely drop this section from the configuration to avoid any confusion. The documentation of that tool does not suggest it understands anything besides shell scripts.

perlpunk: Yes, I think it should be dropped, if it can't format Makefiles correctly

perlpunk (commented):
@b10n1k 
Regarding the os-autoinst-common repo:
I like to keep things simple and step by step, so I suggest to have this PR merged first with the .editorconfig and all the formatting changes.
Also it would be good if you can have a look at os-autoinst and openQA as well.
And after that we can take the .editorconfig and put it in the common repo.

perlpunk:
I will make a PR about updating dependencies.yaml myself.

b10n1k:
I addressed:
 - https://github.com/os-autoinst/scripts/pull/285#discussion_r1463340333
 - https://github.com/os-autoinst/scripts/pull/285#discussion_r1463811447
 - https://github.com/os-autoinst/scripts/pull/285#discussion_r1463294922

I left out https://github.com/os-autoinst/scripts/pull/285#issuecomment-1909784485 which @perlpunk will provide a PR

perlpunk (approved):

baierjan (approved):
As long as the `.editorconfig` makes it to the os-autoinst-common I am fine with the changes.

kalikiana (approved):

PR #286, opened 2024-01-24
Make last_good_test+last_good_build jobs run on the same worker

b10n1k:
Assign the WORKER_CLASS used in the settings if this is present in the vars.json. https://progress.opensuse.org/issues/152853

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -58,6 +58,14 @@ clone() {
         [[ $casedir == null ]] && casedir=''
         repo=${casedir:-'https://github.com/os-autoinst/os-autoinst-distri-opensuse.git'}
         clone_settings+=("CASEDIR=${repo%#*}#${refspec}")
+        if [[ $name_suffix =~ (last_good_tests_and_build) ]]; then
+            worker_vars_settings=$(echo "$vars_json" | runjq -r '.WORKER_CLASS') || return $?
+            if [[ $worker_vars_settings != null ]]; then
+                clone_settings+=("WORKER_CLASS=${worker_vars_settings}")
+            else
+                name+="(unidentified worker class in vars.json)"
+            fi
+        fi
     fi
     [[ -n ${*:5} ]] && clone_settings+=("${@:5}")
     # clear "PUBLISH_" settings to avoid overriding production assets
--- a/test/02-investigate.t
+++ b/test/02-investigate.t
@@ -3,7 +3,7 @@
 source test/init
 bpan:source bashplus +err +fs +sym
 
-plan tests 79
+plan tests 81
 
 host=localhost
 url=https://localhost
@@ -17,6 +17,16 @@ openqa-cli() {
 }
 client_call=(openqa-cli)
 
+fetch-vars-json() {
+    local tid=$1
+    if [[ $tid -eq 10020 ]]; then
+        echo '{"TEST_GIT_URL": "https://github.com/os-autoinst/os-autoinst-distri-openQA.git"}'
+    elif [[ $tid -eq 10022 ]]; then
+        echo '{"WORKER_CLASS": "foo,duh,bar"}'
+    else
+        echo '{}'
+    fi
+}
 rc=0
 out=$(clone 41 42  2>&1 > /dev/null) || rc=$?
 is "$rc" 1 'fails when unable to query job data'
@@ -135,15 +145,6 @@ _clone_call() {
     echo "$@" >&2
 }
 
-fetch-vars-json() {
-    local tid=$1
-    if [[ $tid -eq 10020 ]]; then
-        echo '{"TEST_GIT_URL": "https://github.com/os-autoinst/os-autoinst-distri-openQA.git"}'
-    else
-        echo '{}'
-    fi
-}
-
 clone_call=_clone_call
 try clone 10023 10024
 is "$rc" 0 "Successful clone"
@@ -162,6 +163,12 @@ like "$got" 'CASEDIR=.*/os-autoinst-distri-openQA.git#refspec' "job is cloned wi
 try clone 10023 10030 foo refspec
 is "$rc" 0 "Successful clone"
 like "$got" 'CASEDIR=.*/os-autoinst-testrepo.git#refspec' "job is cloned and CASEDIR is set correctly"
+
+try clone 10022 10030 last_good_tests_and_build:123 refspec
+has "$got" 'WORKER_CLASS=foo,duh,bar' "job assigns the same WORKER_CLASS"
+
+try clone 10021 10030 last_good_tests_and_build:123 refspec
+has "$got" 'unidentified worker class in vars.json' "info shown in the name is WORKER_CLASS is empty"
 
 clone_call=echo
 try investigate 10027


openqa-investigate
```diff
@@ -49,9 +50,18 @@ clone() {
...
+    worker_class=$(echo "$vars_json" | runjq -r '.WORKER_CLASS') || return $?
+    used_worker=$(echo "$clone_job_data" | runjq -r '.job.settings.WORKER_CLASS') || return $?
     clone_settings=("TEST+=:investigate$name_suffix" '_TRIGGER_JOB_DONE_HOOK=1' '_GROUP_ID=0' 'BUILD=')
+    mapfile -td ',' workers < <(echo "${worker_class}")
+    if [[  "${workers[@]}" =~ "$used_worker" ]]; then
+        if [[ $worker_class != null ]] && [[ $name_suffix =~ (last_good_test|last_good_build) ]]; then
+            clone_settings+=("WORKER_CLASS=${used_worker}")
+        fi
+    else
+        echoerr "WORKER_CLASS not found in vars.json"
```
okurz, line 62: This "error" would end up in GRU minion logs on openQA executing this script but instead such message should go into the openQA comment informing openQA investigators about this caveat

openqa-investigate
```diff
@@ -34,6 +34,7 @@ clone() {
     name_suffix=${3+":$3"}
     refspec=${4+$4}
     local clone_job_data
+    vars_json=$(fetch-vars-json "$origin")
```
okurz, line 37: I don't think you should pull this out of the `if refspec` as that would mean that every "retry" job would now also need to parse the origin's vars.json file which should be avoided

openqa-investigate
```diff
@@ -49,9 +50,18 @@ clone() {
...
+    used_worker=$(echo "$clone_job_data" | runjq -r '.job.settings.WORKER_CLASS') || return $?
     clone_settings=("TEST+=:investigate$name_suffix" '_TRIGGER_JOB_DONE_HOOK=1' '_GROUP_ID=0' 'BUILD=')
+    mapfile -td ',' workers < <(echo "${worker_class}")
+    if [[  "${workers[@]}" =~ "$used_worker" ]]; then
+        if [[ $worker_class != null ]] && [[ $name_suffix =~ (last_good_test|last_good_build) ]]; then
+            clone_settings+=("WORKER_CLASS=${used_worker}")
+        fi
+    else
+        echoerr "WORKER_CLASS not found in vars.json"
+    fi
```
okurz, line 63: This "error" would end up in GRU minion logs on openQA executing this script but instead such message should go into the openQA comment informing openQA investigators about this caveat.
How about something like

```suggestion
    if [[  "${workers[@]}" =~ "$used_worker" ]]; then
        if [[ $worker_class != null ]] && [[ $name_suffix =~ (last_good_test|last_good_build) ]]; then
            clone_settings+=("WORKER_CLASS=${used_worker}")
        fi
        name+="(worker:$used_worker)"
    else
        name+="(unidentified worker class)"
    fi
```

```
name+=
```

okurz (changes requested):

openqa-investigate
b10n1k: `name+="(worker:$used_worker)"` should be in the nested if. we set the WORKER_CLASS only for last_good_test+last_good_build. or why did you propose this as such?

openqa-investigate
```diff
@@ -51,13 +51,24 @@ clone() {
...
-	vars_json=$(fetch-vars-json "$origin")
+        vars_json=$(fetch-vars-json "$origin")
         testgiturl=$(echo "$vars_json" | runjq -r '.TEST_GIT_URL')
         casedir=$(echo "$clone_job_data" | runjq -r '.job.settings.CASEDIR') || return $?
         [[ $testgiturl != null ]] && casedir=$testgiturl
         [[ $casedir == null ]] && casedir=''
         repo=${casedir:-'https://github.com/os-autoinst/os-autoinst-distri-opensuse.git'}
         clone_settings+=("CASEDIR=${repo%#*}#${refspec}")
+        worker_class=$(echo "$vars_json" | runjq -r '.WORKER_CLASS') || return $?
+        used_worker=$(echo "$clone_job_data" | runjq -r '.job.settings.WORKER_CLASS') || return $?
```
perlpunk, line 62: The variable names seem to me as if they should be swapped.
The actual *used* worker is the one in the vars.json, the requested worker class is in the job settings, right?

openqa-investigate
```diff
@@ -51,13 +51,24 @@ clone() {
...
+        mapfile -td ',' workers < <(echo "${worker_class}")
+        if [[ "${workers[@]}" =~ "$used_worker" ]]; then
+            if [[ $worker_class != null ]] && [[ $name_suffix =~ (last_good_test|last_good_build) ]]; then
```
perlpunk, line 65: If we only need this logic for `last_good_test|last_good_build`, then this `if` should probably be the outer one

openqa-investigate
okurz: yes, correct. I was confused about the two-level if there. So try to set it within then

openqa-investigate
okurz: There might have been a confusion. I also discussed with @b10n1k and I clarifed that *only* the combination job called "last_good_test+last_good_build" should be forced to run on the same worker as the original job. The jobs "last_good_test" *as well as* "last_good_build" should not be modified at all

okurz (changes requested):

perlpunk (commented):
Maybe the values in the tests could be a bit more realistic to see what it is doing.

I would like to clarify:
For example this job https://openqa.opensuse.org/tests/3895232#settings has
```
WORKER_CLASS=qemu_x86_64
```
in the settings and
```
"WORKER_CLASS" : "openqaworker21,qemu_x86_64,qemu_i586,heavyload,vagrant,tap,wsl2,prg,prg2,prg2-j12"
```
in vars.json.
What would be the WORKER_CLASS for the cloned job then?

test/02-investigate.t
```diff
@@ -159,10 +162,14 @@ try clone 10020 10024 foo refspec
...
 like "$got" 'CASEDIR=.*/os-autoinst-distri-openQA.git#refspec' "job is cloned with sets CASEDIR from TEST_GIT_URL"
 
-try clone 10023 10030 foo refspec
+try clone 10024 10030 foo refspec
 is "$rc" 0 "Successful clone"
 like "$got" 'CASEDIR=.*/os-autoinst-testrepo.git#refspec' "job is cloned and CASEDIR is set correctly"
 
+try clone 10022 10030 last_good_tests_and_build:123 refspec
+has "$got" 'WORKER_CLASS=duh' "job assigns the same WORKER_CLASS"
+has "$got" "last_good_tests_and_build:123(worker:duh)*" "worker appears in last_good_tests name"
```
okurz, line 171:
```suggestion
has "$got" "last_good_tests_and_build:123(worker:duh)*" "worker appears in last_good_tests_and_build name"
```

test/02-investigate.t
```diff
@@ -159,10 +162,14 @@ try clone 10020 10024 foo refspec
...
+
```
okurz, line 172: can you add a counter test for "unidentified worker" please?

okurz (changes requested):

perlpunk:
When I run this branch on this test https://openqa.opensuse.org/tests/3897000 I get:
```
% force=true dry_run=1 ./openqa-investigate https://openqa.opensuse.org/tests/3897000
openqa-cli api ... --host https://openqa.opensuse.org --retries=3 -X PUT jobs/3897000/comments/ text=Automatic investigation jobs for job 3897000:

...
openqa-clone-job ... https://openqa.opensuse.org/tests/3894274
 TEST+=:investigate:last_good_tests_and_build:4d511d4f421c2c55a2fdb8bab90d2abe7c43da44+20240125
 _TRIGGER_JOB_DONE_HOOK=1
 _GROUP_ID=0 BUILD=
 CASEDIR=https://github.com/os-autoinst/os-autoinst-distri-opensuse.git#4d511d4f421c2c55a2fdb8bab90d2abe7c43da44
 WORKER_CLASS=qemu_x86_64
 OPENQA_INVESTIGATE_ORIGIN=https://openqa.opensuse.org/t3897000
* *lxde:investigate:last_good_tests_and_build:4d511d4f421c2c55a2fdb8bab90d2abe7c43da44+20240125(worker:qemu_x86_64)*: t#null

💡[*Detailed explanation of this comment*](https://github.com/os-autoinst/scripts#More-details-and-examples-about-openqa-investigate-comments)
```
I think `WORKER_CLASS=qemu_x86_64` and `(worker:qemu_x86_64)` is not what we want.
The worker class would be `qemu_x86_64` anyway, if we don't change it.
We want  `openqaworker21`.

Also there is the other use case where the worker is not identifiable (not every worker will have its name in the WORKER_CLASS). It's not clear to me how that could be found out. Maybe look up WORKER_HOST and use the subdomain?

perlpunk (changes requested):
Several open questions regarding the algorithm

okurz:
> Maybe look up WORKER_HOST and use the subdomain?
I wouldn't go that far. If the name is not in the worker class just go to the "unidentified worker" branch

perlpunk:
> > Maybe look up WORKER_HOST and use the subdomain?
> I wouldn't go that far. If the name is not in the worker class just go to the "unidentified worker" branch

How do you know the name is not there if you don't know the name?

okurz:
> > > Maybe look up WORKER_HOST and use the subdomain?
> > > I wouldn't go that far. If the name is not in the worker class just go to the "unidentified worker" branch
> 
> How do you know the name is not there if you don't know the name?

Let me stay with your example: In https://openqa.opensuse.org/tests/3897000/file/vars.json I see `WORKER_CLASS	"openqaworker21,qemu_x86_64,qemu_i586,heavyload,vagrant,tap,wsl2,prg,prg2,prg2-j12"` so if we would just use that then we would schedule a job on the very same worker but not just WORKER_CLASS from job settings. That part is not yet there at all in this PR. So it could work like this in pseudo code:
```
new_worker_class = XOR(settings.WORKER_CLASS, vars.WORKER_CLASS)
if new_worker_class has content
then trigger job with new_worker_class
else "unidentified worker"
```

perlpunk:
> ```
> new_worker_class = XOR(settings.WORKER_CLASS, vars.WORKER_CLASS)
> if new_worker_class has content
> then trigger job with new_worker_class
> else "unidentified worker"
> ```
> 
> so I guess the change needs to be somewhere around https://github.com/os-autoinst/scripts/pull/286/files#diff-f73cf39a07f6cf8cdb453862496919d06df16d07e58b274e68ea148dd1f7dae5R113

How about https://openqa.opensuse.org/tests/3893967#settings then:
```
# settings
WORKER_CLASS: heavyload,qemu_aarch64
# vars.json
WORKER_CLASS : qemu_aarch64,qemu_aarch64_lse,qemu_aarch64_max,qemu_aarch64_max_mte_on,heavyload,m6g

# XOR(settings.WORKER_CLASS, vars.WORKER_CLASS)
qemu_aarch64_lse,qemu_aarch64_max,qemu_aarch64_max_mte_on,m6g
```
Or what about jobs that were already having a worker name in WORKER_CLASS settings for some reason. Then via the XOR it would get lost.
Also, using the XOR for determining if there is a difference would probably be true in many cases.
But if we use that, I would still use the WORKER_CLASS from vars.json then to trigger the new job, and not the result of the XOR.

okurz:
> […] But if we use that, I would still use the WORKER_CLASS from vars.json then to trigger the new job, and not the result of the XOR.

True. We should just trigger with the verbatim WORKER_CLASS from vars.json if readable.

openqa-investigate
```diff
@@ -58,6 +58,16 @@ clone() {
         [[ $casedir == null ]] && casedir=''
         repo=${casedir:-'https://github.com/os-autoinst/os-autoinst-distri-opensuse.git'}
         clone_settings+=("CASEDIR=${repo%#*}#${refspec}")
+        if [[ $name_suffix =~ (last_good_tests_and_build) ]]; then
+            local worker_settings
+            get_worker_settings worker_settings
+            if [[ "$worker_settings" != null ]]; then
```
Martchus, line 64:
```suggestion
            if [[ $worker_settings != null ]]; then
```

Martchus (commented):
I suppose it looks generally good.

openqa-investigate
```diff
@@ -83,6 +93,34 @@ clone() {
...
+            # check if worker is in the vars.json
+            if [[ "${workers[*]}" =~ $worker_item ]]; then
+                worker+=" ${worker_item}" && ret=$?
+            else
+                worker+=" unidentified worker ${worker_item}" && ret=$?
+            fi
+        done
+    else
+        worker="${workerclass_in_settings}" && ret=$?
+    fi
```
okurz, line 120: As discussed just simplify to read the WORKER_CLASS from vars.json from the original job, don't care about the settings table.

okurz (approved):

perlpunk (approved):

PR #288, opened 2024-01-25
Add devel package for linters

perlpunk:
Related ticket: https://progress.opensuse.org/issues/153793

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -8,11 +8,12 @@
 #######################################################
 ---
 targets:
-  cpanfile: [main, test]
-  spec:     [main, test]
+  cpanfile: [main, test, devel]
+  spec:     [main, test, devel]
   cpanfile-targets:
     # target: cpanfile target type (default main)
     test: test
+    devel: develop
 main_requires:
   bash:
   coreutils:  # date, ln, tail
@@ -37,3 +38,6 @@ main_requires:
 test_requires:
   perl(Test::Warnings):
   perl(Test::Most):
+devel_requires:
+  python3-yamllint:
+  shfmt:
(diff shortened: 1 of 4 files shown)


dependencies.yaml
```diff
@@ -8,11 +8,12 @@
...
 ---
 targets:
-  cpanfile: [main, test]
-  spec:     [main, test]
+  cpanfile: [main, test, devel]
+  spec:     [main, test, devel]
   cpanfile-targets:
     # target: cpanfile target type (default main)
     test: test
+    devel: develop
```
b10n1k, line 16: Suggestion: I would name it `devel`

baierjan (approved):

dist/rpm/os-autoinst-scripts-deps.spec
```diff
@@ -36,6 +38,11 @@ Suggests:       postgresql
 - auto-review - Automatically detect known issues in openQA jobs, label openQA jobs with ticket references and optionally retrigger
 - openqa-investigate - Automatic investigation jobs with failure analysis in openQA
 
+%package devel
+Summary:        Development package for os-autoinst-scripts-deps
+Group:          Development/Tools/Other
+Requires:       %devel_requires
+
```
perlpunk, line 45: I'm not sure if this enough to create a new package in OBS

baierjan: Good point. I believe so. This should create a new subpackage, like in openQA, but you are likely missing a [`%description`](https://github.com/os-autoinst/openQA/blob/6ca685738c5db7ccdd9394dec8d4841a293c65ec/dist/rpm/openQA.spec#L170) and [`%files`](https://github.com/os-autoinst/openQA/blob/6ca685738c5db7ccdd9394dec8d4841a293c65ec/dist/rpm/openQA.spec#L642) sections.

baierjan (changes requested):

dist/rpm/os-autoinst-scripts-deps.spec
perlpunk: ah right, the description.
I added both now. `%files` is just empty, but it's probably needed.

okurz: yes, an empty section is needed. Current version looks good now

okurz (approved):

baierjan (approved):

PR #289, opened 2024-02-08
Draft script for fetching origins of required packages

baierjan:
Fetch requirements from package spec file, look for package names for
such requirements, search all relevant instances where the package is
maintained and fetch version info from spec file and from Tumbleweed
repo.

Reference: https://progress.opensuse.org/issues/154723

Note: The only related change is in the last commit.

### DIFF
--- a/obs-check-package-origin
+++ b/obs-check-package-origin
@@ -1,0 +1,165 @@
+#!/bin/bash -e
+
+# Motivation: https://progress.opensuse.org/issues/154723
+# 1. fetch requirements from package spec file,
+# 2. look for package names for such requirements
+# 3. search all relevant instances where the package is maintained
+# 4. fetch version info from spec file and from Tumbleweed repo
+
+osc="${osc:-"osc --apiurl https://api.opensuse.org"}"
+
+BLUE="\e[94m"
+RED="\e[91m"
+GRAY="\e[90m"
+ENDCOLOR="\e[0m"
+
+debug() {
+    test -n "$DEBUG" && echo -e "${GRAY}$1${ENDCOLOR}" >&2
+}
+
+info() {
+    echo -e "${BLUE}$1${ENDCOLOR}" >&2
+}
+
+error() {
+    echo -e "${RED}$1${ENDCOLOR}" >&2
+}
+
+init_check() {
+    for cmd in which yq rpmspec osc zypper; do
+        which "$cmd" &> /dev/null || {
+            error "$cmd command is missing"
+            exit 100
+        }
+    done
+}
+
+parse_zypper() {
+    yq -pxml '.stream.search-result.solvable-list | (.solvable.+@name?,.solvable[].+@name)' \
+        | grep -v null \
+        | sort -u \
+        | tr '\n' ' '
+}
+
+get_package_version() {
+    local spec
+    spec="$(basename "$1").spec"
+    local tmpfile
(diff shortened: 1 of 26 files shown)


baierjan:
> Ok. The script generally looks good enough. I am still running it and it takes some time so I don't want to give my approval just yet :)

Yes, it takes a while. Also if you run it with `ZYPPER=1` (and source repos enabled) it is even longer. You are basically making several osc and zypper calls per each entry in requirements and the queries into OBS are not fast (and I did not bother to cache anything in this draft).

> Just because of the name "obs-check-package-origin" that reminded me of similarly named tool which you might want to look into if not already done:
> 
> * https://github.com/openSUSE/openSUSE-release-tools/blob/master/check_source.py
> * https://github.com/openSUSE/openSUSE-release-tools/blob/master/osc-origin.py
> 
> Would be great if you can put your addition in relation to those and describe differences

Yes, good point. Some if that would be probably doing similar things. The osc-origin plugin is probably the most prominent, however I am not sure if that is usable without checking out the package first. The advantage in my approach is, that we are basically doing https://en.opensuse.org/openSUSE:Maintenance_update_process#Check_if_a_Leap_package_is_inherited_from_SLE and that can be done just with plain osc without checking-out anything upfront. If I am not mistaken, the openSUSE-release-tools is wrapping around already created submission requests.

The script is far from perfect and the very core is just `osc sm <package>` in a loop. For further usage in any other scripts, one would probably need to use something better than a shell script.

okurz (approved):

obs-check-package-origin
```diff
@@ -0,0 +1,154 @@
...
+info() {
+    echo -e "${BLUE}$1${ENDCOLOR}" >&2
+}
+
+error() {
+    echo -e "${RED}$1${ENDCOLOR}" >&2
+}
+
+init_check() {
+    which yq &> /dev/null || {
```
foursixnine, line 29:
```suggestion
    command yq &> /dev/null || {
```

This was failing on a TW container until I changed it

baierjan: Hm, so `which` is not preinstalled in a TW container, good to know. Unfortunately `command` is not an equivalent of `which` here.

okurz (approved):

obs-check-package-origin
```diff
@@ -0,0 +1,156 @@
...
+        | while read -r; do
+            package="$(get_package_name "$REPLY")"
+            test -n "$package" || {
+                error "Unable to get package name for '$REPLY'"
+                continue
+            }
+            echo "$package"
+        done
+    rm "$tmpfile"
+}
```
foursixnine, line 87: @baierjan maybe you want to look at how I did it at https://progress.opensuse.org/issues/154747#note-3

Not entirely sure if buildrequires will also pull requires (the distri doesn't have buildrequires for instance)

obs-check-package-origin
foursixnine: > Hm, so `which` is not preinstalled in a TW container, good to know. Unfortunately `command` is not an equivalent of `which` here.

why so? but either way at least is known now

baierjan: If you are brave enough you can try the difference between `which vim` and `command vim`; but even if `command -v` is used, it behaves differently (the difference is in respect of bash functions and aliases, which will be found by command but would not work without interactive shell).

Martchus (approved):

PR #290, opened 2024-02-12
Add script for scheduling MM ping test 

baierjan:
Reference: https://progress.opensuse.org/issues/154624

Note: last commit is the important one; the rest is just `shfmt` like in #289

### DIFF
--- a/openqa-schedule-mm-ping-test
+++ b/openqa-schedule-mm-ping-test
@@ -1,0 +1,59 @@
+#!/bin/bash
+set -eux -o pipefail
+
+openqa_url=${openqa_url:-https://openqa.opensuse.org}
+distri=${distri:-opensuse}
+flavor=${flavor:-DVD}
+arch=${arch:-x86_64}
+version=${version:-Tumbleweed}
+test_name=${test_name:-ping_client}
+
+tmpfile=$(mktemp)
+trap 'rm -f "$tmpfile"' EXIT
+
+cat > "$tmpfile" << EOF
+---
+products:
+  mm-ping-test:
+    distri: $distri
+    flavor: $flavor
+    arch: $arch
+    version: $version
+
+machines:
+  64bit:
+    backend: qemu
+    settings:
+      WORKER_CLASS: qemu_x86_64,tap
+
+.common: &common
+  BOOT_HDD_IMAGE: '1'
+  DESKTOP: textmode
+  IS_MM_SERVER: '1'
+  NICTYPE: tap
+  QEMU_DISABLE_SNAPSHOTS: '1'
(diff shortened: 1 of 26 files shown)


openqa-schedule-mm-ping-test
```diff
@@ -0,0 +1,57 @@
...
+set -eux -o pipefail
+
+openqa_url=${openqa_url:-https://openqa.opensuse.org}
+distri=${distri:-opensuse}
+flavor=${flavor:-DVD}
+arch=${arch:-x86_64}
+version=${version:-Tumbleweed}
+test_name=${test_name:-ping_client}
+
+tmpfile=$(mktemp)
```
okurz, line 11: I suggest to use a cleanup jook, e.g.

```
trap 'rm $tmpfile' EXIT
```

And then without the rm-call after openqa-cli. No guarantee for first syntax though :)

openqa-schedule-mm-ping-test
```diff
@@ -0,0 +1,57 @@
...
+      HOSTNAME: client
+      PARALLEL_WITH: ping_server
+EOF
+
+openqa-cli schedule \
+    --monitor \
+    --host "$openqa_url" \
+    --param-file SCENARIO_DEFINITIONS_YAML="$tmpfile" \
+    DISTRI="$distri" VERSION="$version" FLAVOR="$flavor" ARCH="$arch" \
+    BUILD="$(date +%+4Y%m%d-%H%M)" _GROUP_ID=0 \
```
okurz, line 56: How about `date -Im`

okurz (changes requested):

openqa-schedule-mm-ping-test
baierjan: Yeah, that's a good idea

openqa-schedule-mm-ping-test
baierjan: I was trying to not use `:` in the build as that is sometimes used as a delimiter. But I am ok with any sensible format, if you prefer ISO 8601 a will make that change.

okurz: Yes, I like ISO 8601 :)

okurz (approved):
I like the here-doc approach :)

openqa-schedule-mm-ping-test
Martchus: Note that `date +%+4Y%m%d-%H%M` is *almost* an ISO 8601 timestamp. The only validation I see is the dash (the date and time need to be separated using a `T` instead). An ISO 8601 timestamp does not need to contain any colons as you can omit most separators (see https://en.wikipedia.org/wiki/ISO_8601).

Martchus (approved):

PR #292, opened 2024-02-21
Add GitHub action for cloning an openQA job mentioned in a PR description

Martchus:
This GitHub action allows to clone a job when creating a PR by mentioning it in the PR description via `@openqa: Clone https://…`.

Related ticket: https://progress.opensuse.org/issues/130934


openqa-clone-and-monitor-job-from-pr
```diff
@@ -0,0 +1,56 @@
...
+        push @urls, $2 if index($2, $expected_url) == 0;
+    }
+    return \@urls if @urls;
+    print 'No test cloned; the PR description does not contain ';    # uncoverable statement
+    print "a command like '\@openqa: Clone $expected_url/tests/<JOB_ID>'.\n";    # uncoverable statement
+    exit 0;
+}
+
+sub _run_cmd ($name, $args, $capture_output = 0) {
+    my $res = $capture_output ? `$name $args` : system "$name $args";
```
perlpunk, line 34: We should probable make sure that the arguments cannot be used to run arbitrary commands, not sure if this would be possible with the current inputs. But better be safe than sorry.
If this is used for pull_request_target, it could be possible to leak our API credentials.

Martchus: We *could* but we needed to add `OpenQA::` to the search path. And we needed to make sure not to break the API in the future.

perlpunk: hm, now I'm wondering why the openQA modules aren't installed in the usual place for perl modules, where it would be found by default.
I understand that for the os-autoinst modules that could be a bad idea, but OpenQA is a clear namespace.

perlpunk:
```
open(my $fh, "-|", $cmd, @args);
while (<$fh>) { ... }
```

perlpunk: And to clarify what I meant and what we discussed in the call: If a user manages to manipulate any of the input variables to contain anything then the secrets can be leaked: `'; curl-command-to-post-secrets-to-some-server; '`

openqa-clone-and-monitor-job-from-pr
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env perl
+
+# Note that this script is intended for execution via the GitHub action defined in
+# actions/clone-job/action.yaml.
+
+package openqa_clone_and_monitor_job_from_pr;
```
foursixnine, line 6: This is not needed, as this is a script. An idea on a separate level, why not make it an openQA cli command instead of a wrapper for it?

foursixnine (changes requested):

openqa-clone-and-monitor-job-from-pr
```diff
@@ -0,0 +1,64 @@
...
+my $expected_url = $ENV{OPENQA_HOST} // 'https://openqa.opensuse.org';
+my $group_id = $ENV{OPENQA_SCHEDULE_GROUP_ID} // 118;
+my $pr_body = $ENV{GH_PR_BODY} or die 'GH_PR_BODY must be set';
+my $gh_repo = $ENV{GH_REPO} or die 'GH_REPO must be set';
+my $gh_ref = $ENV{GH_REF} or die 'GH_REF must be set';
+my $gh_srv = $ENV{GITHUB_SERVER_URL} or die 'GITHUB_SERVER_URL must be set';
+my $api_key = $ENV{OPENQA_API_KEY} or die 'OPENQA_API_KEY must be set';
+my $api_secret = $ENV{OPENQA_API_SECRET} or die 'OPENQA_API_SECRET must be set';
+my @secrets = ('--apikey', $api_key, '--apisecret', $api_secret);
+my @vars = ("BUILD=$gh_repo.git#$gh_ref", "_GROUP_ID=$group_id", "CASEDIR=$gh_srv/$gh_repo.git#$gh_ref");
```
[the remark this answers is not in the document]
Martchus: Not really except for manual testing. It is supposed to be used via a GitHub action. But I'll create POD documentation nevertheless.

openqa-clone-and-monitor-job-from-pr
foursixnine: > An idea on a separate level, why not make it an openQA cli command instead of a wrapper for it?

obviously, out of the scope of this PR i'd say

Martchus: Oh, and the openQA repo is already big enough. So I really like to add code that is rather specific to this additional scripts repo (like we did with many of the other scripts contained here). But that also means I cannot really utilize the modules from the openQA repo. Even if the modules would be installed in the normal Perl path (or we adjusted the search path) this would be problematic because when someone would change things on the openQA-side it might break things here (and openQA is not really developed as an API-stable Perl module).

openqa-clone-and-monitor-job-from-pr
```diff
@@ -0,0 +1,106 @@
...
+
+sub _parse_urls ($text) {
+    my @urls;
+    while ($text =~ /(\@openqa:?\s+Clone\s+(https?:[^\s]+))/ig) {
+        push @urls, $2 if index($2, $expected_url) == 0;
+    }
+    return \@urls if @urls;
+    print 'No test cloned; the PR description does not contain ';    # uncoverable statement
+    print "a command like '\@openqa: Clone $expected_url/tests/<JOB_ID>'.\n";    # uncoverable statement
```
foursixnine, line 71: will this be shown in the UI? as a comment on the GH pull request?

openqa-clone-and-monitor-job-from-pr
```diff
@@ -0,0 +1,106 @@
...
+
+It will then clone and monitor an openQA job when a PR mentioning one via e.g.
+C<@openqa: Clone https://openqa.opensuse.org/tests/123456> is created. By
+default it will clone the job on o3 into the "Development / GitHub" group. To
+use a different openQA instance and group, set the environment variables
+C<OPENQA_HOST> and C<OPENQA_SCHEDULE_GROUP_ID> accordingly.
+
+For local testing you may also invoke the script manually. Have a look at the
+handling of environment variables at the beginning of the script code as you
+need to set certain additional environment variables that are normally supplied
```
foursixnine, line 41:
```suggestion
For local testing you may also invoke the script manually. Have a look at the
handling of environment variables at the beginning of the script code as you
need to set certain additional environment variables, the `test/05-clone-and-monitor.t`
can also clarify this script's behavior from user perspective.
```

Since this is only meant for GH, I think it's ok to redirect the reader to how the test is done.

openqa-clone-and-monitor-job-from-pr
Martchus: So far it'll be only shown in the log of the GitHub action. We could extend it later to also write a comment or use the statuses API.

Martchus:
Marking as ready again as I was able to test it and it works. Of course I only tested it so far on my fork but I guess that should suffice for now (see https://github.com/Martchus/os-autoinst-distri-opensuse/pull/2 for an example; I cancelled the cloned job manually). Later (or next week) I'll provide a PR to use it in the openSUSE distribution.

---

In addition to that, I suppose it would also make sense to treat `OPENQA_HOST` not as secret. Otherwise the log looks like this:

```
 Cloned '***/tests/3957590' into:
 - ***/tests/3959032
```

Or was there a reason why we even treat the hostname as a secret?

PR #293, opened 2024-02-21
Identify reproducible infrastructure issues

perlpunk:
I refactored the code a bit and added individual functions for all the api calls

Example: https://openqa.opensuse.org/tests/3951980#comment-549838

Issue: https://progress.opensuse.org/issues/151399

### DIFF
--- a/_common
+++ b/_common
@@ -260,3 +260,36 @@ fetch-vars-json() {
     local tid=$1
     runcurl "${curl_args[@]}" -sS "$host_url/tests/$tid"/file/vars.json || return $?
 }
+
+client-get-job() {
+    local tid=$1
+    openqa-cli "${client_args[@]}" --json jobs/"$tid" || return $?
+}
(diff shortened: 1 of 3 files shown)


okurz (changes requested):

okurz:
Ok, https://openqa.opensuse.org/tests/3951980#comment-549838 is a good example why we should rethink the wording. Here the problem is some expired Windows password or something so calling it "infrastructure" might be misleading. It wouldn't be good if on such issues we as admins are informed. It's in the scope of test maintainers. How about instead of "infrastructure" we say "test environment (settings, external resources, infrastructure)"?

perlpunk:
> How about instead of "infrastructure" we say "test environment (settings, external resources, infrastructure)"?

done

okurz (approved):

Martchus (approved):

PR #298, opened 2024-03-07
Add script for finding out testapi usage

baierjan:
Reference: https://progress.opensuse.org/issues/138032

### DIFF
--- a/os-autoinst-testapi-usage
+++ b/os-autoinst-testapi-usage
@@ -1,0 +1,39 @@
+#!/bin/bash -e
+
+get_function_list() {
+    grep -P '^sub \w+\s*[({:]' "$testapi_path" | cut -f2 -d' ' | grep -v ^_ | sort
+}
+
+rgg() {
+    # needs ripgrep with pcre2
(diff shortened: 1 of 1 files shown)


os-autoinst-testapi-usage
```diff
@@ -0,0 +1,39 @@
...
+
+[ -r "$testapi_path" ] || {
+    echo "$testapi_path not found" >&2
+    exit 1
+}
+grep -q 'package testapi' "$testapi_path" || echo "Warning: provided file does not look like testapi.pm" >&2
+
+while (($#)); do
+    get_function_list | while read -r; do
+        echo "$1 - ${REPLY} : $(rgg "${REPLY}" "$1" | tail -8 | head -n1)"
```
perlpunk, line 36: Where's REPLY coming from?

baierjan: That's the default for `read`.

From `help read`:
> If no NAMEs are supplied, the line read is stored in the REPLY variable.

okurz (approved):

PR #299, opened 2024-03-11
Allow investigation to schedule modules regardless PUBLISH_

b10n1k:
Due to conditions in main_common PUBLISH_* variable cant be empty. Set to `None` to allow the scheduling.

https://progress.opensuse.org/issues/156394

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -70,7 +70,7 @@ clone() {
     [[ -n ${*:5} ]] && clone_settings+=("${@:5}")
     # clear "PUBLISH_" settings to avoid overriding production assets
     # shellcheck disable=SC2207
-    clone_settings+=($(echo "$clone_job_data" | runjq -r '.job.settings | keys[] | select (startswith("PUBLISH_")) | . + "="')) || return $?
+    clone_settings+=($(echo "$clone_job_data" | runjq -r '.job.settings | keys[] | select (startswith("PUBLISH_")) | . + "=none"')) || return $?
     clone_settings+=("OPENQA_INVESTIGATE_ORIGIN=$host_url/t$origin")
     out=$($clone_call "$host_url/tests/$id" "${clone_settings[@]}")
     if [[ $dry_run = 1 ]]; then


okurz (changes requested):
According to https://progress.opensuse.org/issues/156394#note-14
Jobs would still fail in the same way in before. And how does os-autoinst and openQA handle the jobs trying to publish a non-existant asset "none"?

b10n1k:
> According to https://progress.opensuse.org/issues/156394#note-14 Jobs would still fail in the same way in before. And how does os-autoinst and openQA handle the jobs trying to publish a non-existant asset "none"?

I am looking at the second part now. Which maybe it would be good idea to do first

perlpunk:
> According to https://progress.opensuse.org/issues/156394#note-14 Jobs would still fail in the same way in before.

That's actually a good thing. With that I meant that it ran the whole schedule, so the investigation job could reproduce the original failure.

But I guess we should set this to not-ready until changes are done on the openQA side.

openqa-investigate
```diff
@@ -70,7 +70,7 @@ clone() {
     [[ -n ${*:5} ]] && clone_settings+=("${@:5}")
     # clear "PUBLISH_" settings to avoid overriding production assets
     # shellcheck disable=SC2207
-    clone_settings+=($(echo "$clone_job_data" | runjq -r '.job.settings | keys[] | select (startswith("PUBLISH_")) | . + "="')) || return $?
+    clone_settings+=($(echo "$clone_job_data" | runjq -r '.job.settings | keys[] | select (startswith("PUBLISH_")) | . + "=None"')) || return $?
```
Martchus, line 73: "none" is lower-case in English (maybe not in Python but I guess that doesn't matter here):
```suggestion
    clone_settings+=($(echo "$clone_job_data" | runjq -r '.job.settings | keys[] | select (startswith("PUBLISH_")) | . + "=none"')) || return $?
```

kalikiana: Would you expect to be able to publish a disk called `None`? Maybe it should just be caseless anyway - ofc that needs to be implemented in os-autoinst, not here

Martchus: Yes, the matching in os-autoinst should probably be case-insensitive.

okurz (approved):

openqa-investigate
b10n1k: https://github.com/os-autoinst/os-autoinst/pull/2470

okurz:
@b10n1k can you reply to https://github.com/os-autoinst/scripts/pull/299#issuecomment-1987993165 please?

openqa-investigate
okurz: If you reference an issue like this:
* #os-autoinst/os-autoinst/pull/2470

then one sees the current status right away.

b10n1k:
> > According to https://progress.opensuse.org/issues/156394#note-14 Jobs would still fail in the same way in before.
> 
> That's actually a good thing. With that I meant that it ran the whole schedule, so the investigation job could reproduce the original failure.
> 
> But I guess we should set this to not-ready until changes are done on the openQA side.

Sorry missing to reply earlier. However we were looking on this with @kalikiana . I remember running a job (with an older commit of os-autoinst https://github.com/os-autoinst/os-autoinst/pull/2470) and observe to observe the behavior. the job finishes without uploading hdd and without any further problem. If there is something else to be done in openQA i havent seen it. openQA seems to handle the rest properly without any change.

openqa-investigate
b10n1k: how did you do that? :)

Martchus (approved):

openqa-investigate
okurz: In markdown list format with `#<id>`

perlpunk (approved):

PR #302, opened 2024-03-18
Retry osc command-line invocations

kalikiana:
Try again if OBS is unavailable temporarily.

See: https://progress.opensuse.org/issues/157018

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -33,6 +33,7 @@ main_requires:
   perl(Mojo::File):
   perl(Text::Markdown):
   perl(YAML::PP):
+  retry:
   html-xml-utils:
   xmlstarlet:
 test_requires:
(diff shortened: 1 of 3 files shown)


baierjan:
I guess  you also need to call `make update-deps` and include the resulting changes in the spec file to this commit.

```
Failed test 'No changed files after 'make update-deps''
#   at xt/01-make-update-deps.t line 25.
#  M dist/rpm/os-autoinst-scripts-deps.spec
```

perlpunk (approved):

baierjan:
Unfortunately, this will still need a fix or a revert:
```
+ retry -e osc service wait devel:openQA openQA
ok
+ retry -e osc co --server-side-source-service-files devel:openQA/openQA
retry: unrecognized option '--server-side-source-service-files'
usage: /usr/bin/retry [options] [cmd...]
options:
    -h,--help                   Show this help
    -r,--retries=RETRIES        How many retries to do on command failure after
                                the initial try. Defaults to 3.
    -s,--sleep=SLEEP            How many seconds to sleep between retries.
                                Defaults to 3 seconds.
    -e,--exponential[=FACTOR]   Enable simple exponential back-off algorithm.
                                Disabled by default, factor defaults to 2
                                (binary exponential back-off).
```

Maybe using `retry -e -- osc` as a prefix will fix it?

PR #306, opened 2024-03-19
Prevent useless retries on osc cat always failing in pipe

okurz:
https://github.com/os-autoinst/scripts/pull/305 introduced a problem
causing "osc cat" to always retry as it fails within a piped call.
Apparently there is a SIGPIPE due to the head in the command:

```
osc cat … | grep '…' | head -n1
```

which can be seen with `echo "${PIPESTATUS[@]}"`. The
`grep … | head -n1`
can be simplified to `grep -m1` which gets rid of the sigpipe of
grep but keeps the failure of osc cat. But we can use process
substitution to rescue :)

This also allows us to enforce the proper handling of pipefails
everywhere else.

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -euo pipefail
 
 export LC_ALL=C
 src_project="${src_project:-"devel:openQA"}"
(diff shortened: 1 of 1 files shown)


kalikiana (approved):

perlpunk:
For reference, the problematic `cat` is:
```
+++ retry -e -- osc cat openSUSE:Factory/os-autoinst/os-autoinst.changes
+++ grep 'Update to version'
+++ head -n1
Retrying up to 3 more times after sleeping 3s …
Retrying up to 2 more times after sleeping 6s …
Retrying up to 1 more times after sleeping 12s …
++ line='- Update to version 4.6.1710500193.0c8ac9b:'
```

perlpunk (approved):

PR #308, opened 2024-03-19
Revert "Prevent useless retries on osc cat always failing in pipe"

perlpunk:
Reverts os-autoinst/scripts#306


Martchus:
I guess we can avoid this in favor of #307.

okurz:
https://github.com/os-autoinst/scripts/pull/307 is already merged

PR #310, opened 2024-03-20
Prevent leaked background processes of "osc cat"

okurz:
My previous change from using a pipe to process substitution helped to
not accumulate unnecessary waiting time due to retries of "osc cat" but
it caused the "osc cat" to still continue in the background leading to
intermingled output. So instead now this commit goes back to using a
pipe operator but combined with "cat >/dev/null" to swallow all
superfluous output to make "osc cat" happy and not cause retriggers.

Tested locally with

```
project="devel:openQA"; package=openQA; file=openQA.changes; osc="retry -s0 -e -- osc"; $osc cat "$project/$package/$file" | { grep -m1 'Update to version'; cat >/dev/null; }
```

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -89,7 +89,12 @@ last_revision() {
         file=_service:$service:$file
     fi
     local line
-    line=$(grep -m1 'Update to version' <($osc cat "$project/$package/$file"))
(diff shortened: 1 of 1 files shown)


kalikiana (approved):

okurz:
> Looks like the style check wants you to go multi-line.

Done

PR #311, opened 2024-03-20
os-autoinst-obs-auto-submit: Fix case of no previous pending request

okurz:
This is a follow-up to the recent change to abort the script on
unexpected pipefails. In case there is no pending submit request we must
ignore the exit status as the check for the value validity is done
afterwards anyway.


okurz:
Sorry, wrong branch -> https://github.com/os-autoinst/scripts/pull/312

PR #313, opened 2024-03-28
Support passing job settings when cloning job via PR description

Martchus:
See https://progress.opensuse.org/issues/130943

### DIFF
--- a/openqa-clone-and-monitor-job-from-pr
+++ b/openqa-clone-and-monitor-job-from-pr
@@ -34,8 +34,8 @@ jobs:
 =end text
 
 It will then clone and monitor an openQA job when a PR mentioning one via e.g.
-C<@openqa: Clone https://openqa.opensuse.org/tests/123456> is created. By
-default it will clone the job on o3 into the "Development / GitHub" group. To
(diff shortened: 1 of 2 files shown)


kalikiana (approved):

perlpunk:
Considering every github user can schedule jobs - are there any settings that shouldn't be overridable? e.g. `PUBLISH_HDD_1`

okurz:
> Considering every github user can schedule jobs - are there any settings that shouldn't be overridable? e.g. `PUBLISH_HDD_1`

Wait, how can every github user schedule jobs? For not trusted, unknown contributors github asks project owners to approve CI runs. Wouldn't that apply here?

Martchus:
I suppose it would apply here. Creating/maintaining a list of settings to exclude seems also not very helpful because one can also already change the test code (and that test code can then change settings anyway).

perlpunk:
True, those rules probably apply also to actions run when a comment is written.
Still, good that we talked about it IMHO :)

okurz (approved):
Let's hope you are right then :-)

PR #314, opened 2024-04-15
Add racktables machine reachability check

nicksinger:
Related ticket: https://progress.opensuse.org/issues/158907

### DIFF
--- a/racktables/find_unpingable_machines.sh
+++ b/racktables/find_unpingable_machines.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+python3 get_unused_machines.py | while read fqdn; do
+    ping -c1 -W1 "${fqdn}" &> /dev/null && echo "${fqdn} up"
+done
--- a/racktables/get_unused_machines.py
+++ b/racktables/get_unused_machines.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+from racktables import Racktables, RacktablesObject
+from getpass import getpass
+import os
+
+rt_url = os.environ.get("RT_URL", "https://racktables.suse.de")
+user = (
+    os.environ["RT_USERNAME"]
+    if "RT_USERNAME" in os.environ.keys()
+    else input("Username: ")
+)
+pwd = (
+    os.environ["RT_PASSWORD"]
+    if "RT_PASSWORD" in os.environ.keys()
+    else getpass("Password (masked): ")
+)
+
+rt = Racktables(rt_url, user, pwd)
+search_payload = {
+    "andor": "and",
+    "cft[]": "197",
+    "cfe": "{%24typeid_4}+and+not+{Decommissioned}",
+    "page": "depot",
+    "tab": "default",
+    "submit.x": "9",
+    "submit.y": "24",
+}
+results = rt.search(search_payload)
+for result_obj in results:
+    url_path = result_obj.find("a")["href"]
+    obj = RacktablesObject(rt)
+    obj.from_path(url_path)
+    try:
+        print(obj.fqdn, flush=True)
+    except:
+        print(obj.common_name, flush=True)
--- a/racktables/racktables.py
+++ b/racktables/racktables.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+from requests.auth import HTTPBasicAuth
+from bs4 import BeautifulSoup
+from os.path import join as join
+import requests
+import re
+
+
+class Racktables:
+    def __init__(self, url, username, password):
+        self.s = requests.Session()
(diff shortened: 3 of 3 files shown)


racktables/find_unpingable_machines.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+python3 get_unused_machines.py | while read fqdn; do
+    ping -c1 -W1 "${fqdn}" > /dev/null 2>&1 && echo "${fqdn} up"
```
baierjan, line 3: Or maybe even
```suggestion
    ping -c1 -W1 "${fqdn}" &> /dev/null && echo "${fqdn} up"
```
as a more readable way to redirect both to `/dev/null`?

baierjan (commented):
Looks OK.

racktables/find_unpingable_machines.sh
nicksinger: I always forget this syntax and I hate it (because I regularly background my programs without wanting it…) but I added it :)

racktables/get_unused_machines.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+from racktables import Racktables, RacktablesObject
+from getpass import getpass
+import os
+
+rt_url = os.environ["RT_URL"] if "RT_URL" in os.environ.keys() else "https://racktables.suse.de"
```
Martchus, line 6: If is possible I'd suggest to change these three lines like this:
```suggestion
rt_url = os.environ.get("RT_URL", "https://racktables.suse.de")
```

racktables/racktables.py
```diff
@@ -0,0 +1,40 @@
...
+    self.s = requests.Session()
+    self.s.verify = "/etc/ssl/certs/SUSE_Trust_Root.pem"
+    self.s.auth = HTTPBasicAuth(username, password)
+    self.url = url
+
+  def search(self, search_payload={}):
+    req = self.s.get(join(self.url, "index.php"), params="&".join("%s=%s" % (k,v) for k,v in search_payload.items()))
+    soup = BeautifulSoup(req.text, 'html.parser')
+    result_table = soup.find("table", {"class": "cooltable"})
+    result_objs = result_table.find_all("tr", lambda tag: tag != None) #racktables does not use table-heads so we have to filter the header out (it has absolutely no attributes)
```
Martchus, line 19:
```suggestion
    result_objs = result_table.find_all("tr", lambda tag: tag != None)  # Racktables does not use table-heads so we have to filter the header out (it has absolutely no attributes)
```

racktables/racktables.py
```diff
@@ -0,0 +1,40 @@
...
+    return result_objs
+
+class RacktablesObject(object):
```
asdil12, line 22:
```suggestion
class RacktablesObject:
```

This is not needed anymore in python3.

racktables/racktables.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+from requests.auth import HTTPBasicAuth
+from bs4 import BeautifulSoup
+from os.path import join as join
+import requests
+import re
+
+class Racktables():
```
asdil12, line 8:
```suggestion
class Racktables:
```

racktables/racktables.py
```diff
@@ -0,0 +1,40 @@
...
+    objectview_table = soup.find("table", {"class": "objectview"})
+    portlets = list(objectview_table.find_all("div", {"class": "portlet"}))
+    summary = list(filter(lambda x: x.find("h2").text == "summary", portlets))[0]
+    rows = list(summary.find_all("tr"))
+    for row in rows:
+      try:
+        name = row.find("th").text
+        value = row.find("td").text
+        sane_name = re.sub(r'[^a-z_]+', '', name.lower().replace(" ", "_"))
+        setattr(self, sane_name, value)
```
asdil12, line 38: Brave, but I don't mind ;)

asdil12:
Could you maybe indent the python scripts also by 4. (Or using tabs - I would totally approve that as well :P)

nicksinger:
So all the changes you requested are in https://github.com/os-autoinst/scripts/commit/b1a6ef5f0b4900804aa28c9c4b0537c5afb098c6 but the current version of both `.py`-files is now formatted with default `black`-settings.

racktables/get_unused_machines.py
nicksinger: I'd love to have used this pattern for the other two values as well…

racktables/racktables.py
nicksinger: image

Martchus (approved):

okurz (approved):

PR #316, opened 2024-04-22
Report racktables results and issues for the user

nicksinger:


Martchus (approved):
I guess it is fine although it would probably be useful to log also the response and possibly even the route the script tried to query.

racktables/find_pingable_machines.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 python3 get_unused_machines.py | while read fqdn; do
     ping -c1 -W1 "${fqdn}" &> /dev/null && echo "${fqdn} up"
-done
+done && exit 1 || exit 0
```
okurz, line 4: If you want to turn around the exit code you should be able to do that with

```
! python3 … ; do
…
done
```

nicksinger: your suggestion looks good. However while testing I realized that without a temporary variable to store the state, I will most likely not archive what I want with this change (fail if _any_ host is reachable, not just fail if the last one is reachable)

okurz: ok, nevermind. Just keep as is then I guess

racktables/racktables.py
```diff
@@ -18,6 +18,11 @@ def search(self, search_payload={}):
             join(self.url, "index.php"),
             params="&".join("%s=%s" % (k, v) for k, v in search_payload.items()),
         )
+        status = req.status_code
+        if status == 401:
+          raise Exception("Racktables returned 401 Unauthorized. Are your credentials correct?")
+        elif status >= 300:
+          raise Exception(f"Racktables returned statuscode {status} for a search request. Manual investigation needed.")
```
kalikiana, line 25: Might be worth including the search terms if we can? To make it easy to reproduce

nicksinger: I tried to incorporate your feedback. Are you fine with that?

racktables/find_pingable_machines.sh
nicksinger: got pretty ugly now. If you know of a better way please let me know :)

okurz (approved):

racktables/find_pingable_machines.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+found_machines=0
+while read fqdn; do
```
baierjan, line 3: shellcheck would probably suggests

```suggestion
while read -r fqdn; do
```

racktables/find_pingable_machines.sh
```diff
@@ -0,0 +1,6 @@
...
+    ping -c1 -W1 "${fqdn}" &> /dev/null && echo "${fqdn} up" && found_machines=$(($found_machines + 1))
```
baierjan, line 4: You are in "math mode" here
```suggestion
    ping -c1 -W1 "${fqdn}" &> /dev/null && echo "${fqdn} up" && found_machines=$((found_machines + 1))
```

racktables/find_pingable_machines.sh
```diff
@@ -0,0 +1,6 @@
...
+done <<< $(python get_unused_machines.py)
+exit $(($found_machines > 0))
```
baierjan, line 6:
```suggestion
exit $((found_machines > 0))
```

racktables/find_pingable_machines.sh
nicksinger: might make sense if machine names contain funny characters :)

Martchus (approved):

PR #318, opened 2024-05-17
Allow cloning jobs from GitHub comment bodies

Martchus:
* Allow to restrict allowed users to a team of a GitHub organization to
  avoid security concerns
* Allow querying the PR ref and clone URL via the GitHub API because
  those are not part of the event data of comment events
* Update status on GitHub when cloning a job via a PR comment
* See https://progress.opensuse.org/issues/130940

### DIFF
--- a/openqa-clone-and-monitor-job-from-pr
+++ b/openqa-clone-and-monitor-job-from-pr
@@ -53,22 +53,41 @@ package openqa_clone_and_monitor_job_from_pr;    # for testing
 
 use Mojo::Base -strict, -signatures;
 use Mojo::JSON qw(decode_json);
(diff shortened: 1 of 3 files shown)


Martchus:
This is how it would look like: https://github.com/Martchus/os-autoinst-distri-opensuse/pull/4
(Ignore the "default" check.)

The workflow-yaml one needs to add looks very similar to what we already have for cloning jobs from PR descriptions, see https://github.com/Martchus/os-autoinst-distri-opensuse/blob/master/.github/workflows/openqa-comment.yml.

asdil12 (approved):

PR #320, opened 2024-05-22
Add exp retries for poo req for ~1h to openqa-label-known-issues

asdil12:
Ticket: https://progress.opensuse.org/issues/160343


_common
```diff
@@ -68,22 +68,40 @@ runjq() {
     return $rc
 }
 
+exp_retry() {
+    local stop_exponent=$1
+    local exponent=$2
+    if [[ "$exponent" == "$stop_exponent" ]]; then
```
Martchus, line 74:
```suggestion
    if [[ $exponent == "$stop_exponent" ]]; then
```

_common
```diff
@@ -68,22 +68,42 @@ runjq() {
...
+        warn "Retries exceeded"
```
Martchus, line 75:
```suggestion
        warn "$stop_exponent retries exceeded"
```

Martchus (commented):
Now there's a little bit too many `&&` and `||` in one line but I *guess* it'll work.

perlpunk (commented):
Would it be possiblre to just use the `retry` command?

asdil12:
> Would it be possible to just use the `retry` command?

Wouldn't bee that easy as `runcurl` is a function - not a command.

perlpunk:
Hm, this PR wasn't approved by anyone

asdil12:
Ah - overlooked that I guess.

PR #321, opened 2024-05-23
Increase retry sleep & log at openqa-cli in openqa-label-known-issues

asdil12:
Ticket: https://progress.opensuse.org/issues/159840


openqa-label-known-issues
```diff
@@ -126,6 +126,7 @@ investigate_issue() {
     local id="${1##*/}"
     local reason
     local curl_output
+    echo "Requesting jobs/${id} via openqa-cli" >&2
```
perlpunk, line 129: I think we only want errors or warnings in the log.

kalikiana (approved):

Martchus (approved):

openqa-label-known-issues
```diff
@@ -126,6 +126,7 @@ investigate_issue() {
     local id="${1##*/}"
     local reason
     local curl_output
+    echo "Requesting jobs/${id} via openqa-cli"
```
perlpunk, line 129: Where will this message appear now?

asdil12: maybe in the gru log?

perlpunk: Well, after my first comment you removed the `>&2`. Before it would appear in the gru journal. But we only want errors or warnings in the log.
What was the purpose of removing the `>&2`? Where should it appear?
If the message is supposed to be helpful it should show up somewhere?

asdil12: I added it because of the suggestion 
* Consider improving the scripting, error reporting, etc. e.g. reveal the URL/ API route

perlpunk: Sure, but where does it appear? If we don't see it, we could also remove it again.
It could go to STDERR and land in the journal if you add `>&2`, but then it should only be written in case it fails.

perlpunk: Actually, since this is openqa-label-known-issues: its output is filtered in our hook script:
https://github.com/os-autoinst/scripts/blob/master/openqa-label-known-issues-and-investigate-hook#L32
So it will never appear anywhere.

asdil12: I guess at least you can see it, if called by hand.

PR #325, opened 2024-06-03
Document usage of `openqa-clone-and-monitor-job-from-pr` for PR comments

Martchus:
Relevant ticket: https://progress.opensuse.org/issues/130940


kalikiana (approved):
I would have made this an action, which is easier to use and gives us visibility by way of being discoverable in the marketplace.

Looks very nice either way.

okurz (approved):

PR #326, opened 2024-07-10
Add script openqa-search-maintenance-core-jobs

manfredi:


okurz:
please see the results from CI style checks: https://github.com/os-autoinst/scripts/actions/runs/9877119446/job/27277932785?pr=326

manfredi:
Now style should be fine.

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+
+# Search openQA jobs related to a given update inside job groups:
+# - Maintenance: Single Incidents / Core Incidents
+# - Maintenance: Aggregated updates / Core Maintenance Updates
+# Default search within the last 5 days for "Aggregated updates" builds
+# Usage: eg
+#   openqa-qam-sle -a SUSE:Maintenance:34576:337335
+#   openqa-qam-sle -a SUSE:Maintenance:34577:337331
+#   openqa-qam-sle -a SUSE:Maintenance:34564:337279
+#   openqa-qam-sle -a SUSE:Maintenance:34581:337626
```
Martchus, line 11: Maybe the usage function should print that. So it would make sense to turn this comment into a multi-line string and print it in the usage function. You can move the usage function to the very top so this can still be easily found when looking at the script code itself.

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+#     367 # 15-SP3
+#     439 # 15-SP4
+#     490 # 15-SP5
+#     546 # 15-SP6
+#     106 # 12-SP3
+#     282 # 12-SP5
+#     521 # 15-SP4-TERADATA
+#     191 # 12-SP3-TERADATA
+# )
+
```
Martchus, line 27: Please avoid out-commented code.

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+dict_group["15-SP1"]="233"
+dict_group["15-SP2"]="306"
+dict_group["15-SP3"]="367"
+dict_group["15-SP4"]="439"
+dict_group["15-SP5"]="490"
+dict_group["15-SP6"]="546"
+dict_group["12-SP3"]="106"
+dict_group["12-SP5"]="282"
+dict_group["15-SP4-TERADATA"]="521"
+dict_group["12-SP3-TERADATA"]="191"
```
Martchus, line 39: You can declare an associative array in bash like this: https://github.com/Martchus/PKGBUILDs/blob/master/devel/versions.sh

So there's no need for all the quotes and repetition of `dict_group`.

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+_title() {
+    echo -e "\033[01;36m$1\033[0m"
+}
+
+search_maintenance_single_incidents() {
+    _title "Maintenance: Single Incidents / Core Incidents"
+    local ReviewRequestID=$1
+
+    local ID
+    ID="$(echo "$ReviewRequestID" | awk -F':' '{printf "%s", $3}')" # The ID of the update
```
Martchus, line 62: This could be one line and there's no need for quoting the right hand side of an assignment (unless you want a string literal with spaces in it). I'd also prefer to lower-case local variables.

```suggestion
    local id=$(echo "$ReviewRequestID" | awk -F':' '{printf "%s", $3}') # The ID of the update
```

The same counts for the other occurrences.

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+    incident_settings=$(curl -sk -X GET "http://dashboard.qam.suse.de/api/incident_settings/$ID")
+
+    local build
+    build=$(echo "$incident_settings" | jq -r '.[0].settings.BUILD')
+
+    local versions
+    versions=$(echo "$incident_settings" | jq -r '.[].settings.VERSION' | sort | uniq)
+
+    local versions_teradata
+    versions_teradata=$(echo "$incident_settings" | jq -r '[.[]| { KEY:(.settings.VERSION + "|" +.settings.FLAVOR)}] | .[].KEY' | sort | uniq | grep TERADATA | sed -e 's/|.*-TERADATA$/-TERADATA/' | sort | uniq)
```
Martchus, line 75: Isn't it enough to run `sort | uniq` one time at the end?

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+        [[ $failed -eq 0 ]] && _ok "PASSED"
+        [[ $failed -gt 0 ]] && _ko "FAILED ($failed jobs)"
+        echo ""
+    done
+    echo "---"
+}
+
+search_maintenance_aggregated() {
+    _title "Maintenance: Aggregated updates / Core Maintenance Updates"
+    local ReviewRequestID=$1
```
Martchus, line 112: I'd stick to snake case in bash scripts, especially because your other variables are also using snake case already:
```suggestion
    local review_request_id=$1
```

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+            ;;
+    esac
+done
+
+if [[ -z "$ReviewRequestID" ]]; then
+    usage
+    exit 1
+fi
+
+days="${days:-5}"
```
Martchus, line 211:
```suggestion
days=${days:-5}
```

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+        [[ $failed -gt 0 ]] && _ko "FAILED ($failed jobs)"
+        echo ""
+    done
+    echo "---"
+}
+
+search_maintenance_aggregated() {
+    _title "Maintenance: Aggregated updates / Core Maintenance Updates"
+    local ReviewRequestID=$1
+    local days=${2:-5}  # Default last 5 days
```
Martchus, line 113: It would make sense to rename this local variable for better distinction with the global `days` variable.

Martchus (changes requested):
The name of the script is very generic and not meaning full at all to anybody not familiar with SUSE internal abbreviations. Maybe `openqa-search-maintenance-jobs` would be a better name.

If there's more SUSE internal to come it would make sense to put those scripts into a sub directory. (The root of this repo is already quite big and we sometimes have a hard time to find the right scripts.)

openqa-qam-sle
manfredi: These comes after apply the `shellcheck`, eg:
```bash
$ shellcheck openqa-qam-sle
In openqa-qam-sle line 62:
    local ID="$(echo "$ReviewRequestID" | awk -F':' '{printf "%s", $3}')" # The ID of the update
          ^-- SC2155 (warning): Declare and assign separately to avoid masking return values.

For more information:
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
```

openqa-qam-sle
manfredi: In general I prefer to quote the strings to avoid some sporadic side effects, like blank and so no.

openqa-qam-sle
manfredi: just a bookmark for next develops

openqa-qam-sle
manfredi: Could be, but I'll need to test it

openqa-qam-sle
manfredi: The `local` should do just that. It's actually the same var, so it makes more sense to have the same name in my point of view.

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+    local groupid="414" # Maintenance: Aggregated updates / Core Maintenance Updates
+
+    local ID
+    ID="$(echo "$ReviewRequestID" | awk -F':' '{printf "%s", $3}')" # The ID of the update
+    # local RR="$(echo "$ReviewRequestID" | awk -F':' '{printf "%s", $4}')" # The Release Request of the update
```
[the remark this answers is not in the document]
manfredi: just a placeholder for future developments

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+    versions=$(curl -sk -X GET "http://dashboard.qam.suse.de/api/incident_settings/$ID" | jq -r '.[].settings.VERSION' | sort | uniq)
+
+    for version in $versions; do
+        [[ -z "${dict_group[$version]}" ]] && continue
+
+        for day in $(seq 0 "$days"); do
+            local build
+            build="$(date +%Y%m%d -d "$day day ago")-1"
+
+            local URL="https://openqa.suse.de/api/v1/jobs/overview?distri=sle&version=${version}&build=${build}&groupid=${groupid}&state=done"
```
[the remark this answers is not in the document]
manfredi: For the moment it is a bash script.. But I agree with you

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+    local ReviewRequestID=$1
+    local days=${2:-5}  # Default last 5 days
+    local groupid="414" # Maintenance: Aggregated updates / Core Maintenance Updates
+
+    local ID
+    ID="$(echo "$ReviewRequestID" | awk -F':' '{printf "%s", $3}')" # The ID of the update
+    # local RR="$(echo "$ReviewRequestID" | awk -F':' '{printf "%s", $4}')" # The Release Request of the update
+
+    local versions
+    versions=$(curl -sk -X GET "http://dashboard.qam.suse.de/api/incident_settings/$ID" | jq -r '.[].settings.VERSION' | sort | uniq)
```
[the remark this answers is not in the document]
manfredi: Yes I agree, but the original idea was to have an independent function to reuse with no global variables to depend on. This also explains why I preferred passing parameters to functions rather than using global variables

openqa-qam-sle
manfredi: Done

openqa-qam-sle
manfredi: Removed

openqa-qam-sle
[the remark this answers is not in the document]
manfredi: Removed

openqa-qam-sle
```diff
@@ -0,0 +1,220 @@
...
+}
+
+search_build_checks() {
+    _title "Build checks"
+    local ReviewRequestID=$1
+    local index
+    index="https://qam.suse.de/testreports/${ReviewRequestID}/build_checks"
+    local logs
+    logs=$(curl -sLk -X GET "$index" | perl -lne '/a href="(.*?\.log)"/ && print "$1"')
+    if [[ -n "$logs" ]]; then
```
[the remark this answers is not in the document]
manfredi: Removed quote

okurz (changes requested):
There is an API route which allows to search for jobs by setting. With this it should be much easier to find according jobs. Please see https://progress.opensuse.org/issues/117655, https://progress.opensuse.org/issues/120118, 
https://progress.opensuse.org/issues/120115 .
With that it's possible to use examples like

```
openqa-cli api --o3  /job_settings/jobs key=*_TEST_ISSUES list_value=1234567
```

openqa-qam-sle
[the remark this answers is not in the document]
Martchus: I think for something like this it is totally fine to use a global variable.

openqa-qam-sle
Martchus: Ok, I guess if shellcheck demands it it should stay :-)

openqa-qam-sle
Martchus: You can at least avoid the repetition of `dict_group` and the quotes on the right hand side.

openqa-qam-sle
```diff
@@ -0,0 +1,208 @@
...
+
+if [[ -z "$ReviewRequestID" ]]; then
+    usage
+    exit 1
+fi
+
+days="${days:-5}"
+
+# ReviewRequestID=$1
+# days=${2:-5}
```
Martchus, line 202: Maybe I wasn't clear but I think all commented-out code should be removed.

manfredi:
Thanks Oliver! That is interesting, but my goal is the "reverse way": given a job id (that I already know) search which values are inside settings of type `_TEST_ISSUES` to compare with a given value.

openqa-qam-sle
manfredi: Removed

openqa-qam-sle
manfredi: Removed duplicate filters in pipe

openqa-qam-sle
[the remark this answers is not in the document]
manfredi: Replaced hard-coded url with readonly var

openqa-qam-sle
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env bash
+
+set -e
```
okurz, line 3: I recommend

```suggestion
set -euo pipefail
```

to prevent bad surprises

openqa-qam-sle
okurz, line 1: The name "openqa-qam-sle" is highly unintuitive and also we should consider openSUSE the default rather than SLE or even better make that distro agnostic. Maybe you can come up with a better name?

openqa-qam-sle
```diff
@@ -0,0 +1,208 @@
...
+
+usage() {
+    cat << EOF
+Usage: $(basename "$0") -a SUSE:Maintenance:II:RR [-d days]
+
+Search openQA jobs related to a given update inside job groups:
```
okurz, line 9: Do you mean "Search *for* openQA jobs"? Because in https://github.com/os-autoinst/scripts/pull/326#issuecomment-2222987460 you stated the opposite that you would start with a job-id but here you search by incident+RR which I explained is better done with the /job_settings/jobs api route

openqa-qam-sle
```diff
@@ -0,0 +1,208 @@
...
+
+    local build
+    build=$(echo "$incident_settings" | jq -r '.[0].settings.BUILD')
+
+    local versions
+    versions=$(echo "$incident_settings" | jq -r '.[].settings.VERSION' | sort | uniq)
+
+    local versions_teradata
+    versions_teradata=$(echo "$incident_settings" | jq -r '[.[]| { KEY:(.settings.VERSION + "|" +.settings.FLAVOR)}] | .[].KEY' | grep TERADATA | sed -e 's/|.*-TERADATA$/-TERADATA/' | sort | uniq)
+
```
okurz, line 74: Please treat anything related to that term confidential

openqa-qam-sle
```diff
@@ -0,0 +1,208 @@
...
+                    local running_url="${url_openqa}/api/v1/jobs/overview?distri=sle&version=${version}&build=${build}&groupid=${groupid}&state=scheduled&state=running"
+                    local running
+                    running=$(curl -sk -X GET "$running_url" | jq -r '. | length')
+                    [[ $running -gt 0 ]] && {
+                        _warn "RUNNING / SCHEDULED ($running jobs) Awaiting completion..."
+                        echo ""
+                        continue
+                    }
+
+                    local failed_url="${url_openqa}/api/v1/jobs/overview?distri=sle&version=${version}&build=${build}&groupid=${groupid}&result=failed&result=incomplete&result=timeout_exceeded"
```
okurz, line 156: I suggest you call that "not_ok" instead of "failed" as you include more than just "failed"

openqa-qam-sle
```diff
@@ -0,0 +1,208 @@
...
+                    local failed
+                    failed=$(curl -sk -X GET "$failed_url" | jq -r '. | length')
+                    [[ $failed -eq 0 ]] && _ok "PASSED"
+                    [[ $failed -gt 0 ]] && _ko "FAILED ($failed jobs)"
```
okurz, line 160:
```suggestion
                    [[ $failed -eq 0 ]] && _ok "PASSED" || _ko "FAILED ($failed jobs)"
```

okurz (changes requested):

openqa-qam-sle
Martchus: What about the one I mentioned in https://github.com/os-autoinst/scripts/pull/326#pullrequestreview-2171550263?

okurz: Yes, something like that.

openqa-qam-sle
manfredi: It seems that your suggestion does not pass the `shellcheck` control:
https://github.com/os-autoinst/scripts/actions/runs/9943776882/job/27468355251?pr=326
https://www.shellcheck.net/wiki/SC2015

openqa-qam-sle
manfredi: I keep in mind

openqa-qam-sle
manfredi: There are two points here:
1. I prefer not to have any dependencies with other tools, so the script can be run even if openqa-cli is not installed
2. is there an equivalent API URL that can be called with a simple curl command?

Said that, I am going to modify the description so that could explain better the goal of this script, that is basically to fill the section "regression tests" in the log of QA test report.

openqa-qam-sle
manfredi: Would be fine for me, but actually it searches only for a subset, that is `qam-sle`.

openqa-qam-sle
manfredi: In my case this setting `set -o pipefail` is not appropriate because the pipe is expected to finish with a non-zero exit code, when it finds no match. In fact in most cases the return code of the pipe will not be zero.

openqa-qam-sle
okurz: ah right, that :)

openqa-qam-sle
okurz: I know. To make that not confusing I was suggesting to use a different type

openqa-qam-sle
foursixnine: Well, `teradata` isn't necessarily confidential: https://github.com/search?q=repo%3Aos-autoinst%2Fos-autoinst-distri-opensuse%20teradata&type=code

okurz: yeah good point. I would still prefer to avoid it here.

openqa-qam-sle
okurz: > There are two points here:
> 
>     1. I prefer not to have any dependencies with other tools, so the script can be run even if openqa-cli is not installed

Sure, that's fine and certainly possible with the way *u implemented using only curl and openQA API routes
> 
>     2. is there an equivalent API URL that can be called with a simple curl command?

Well I think the API route /job_settings/jobs which can be called just with curl should suffice. If there is something missing I would be happy to learn about that.

> Said that, I am going to modify the description so that could explain better the goal of this script, that is basically to fill the section "regression tests" in the log of QA test report.

Good idea!

openqa-qam-sle
okurz: ok, nevermind. We have https://www.suse.com/de-de/success/teradata/ :)

openqa-qam-sle
okurz: IMHO the comment is redundant though

manfredi: removed `# Default last 5 days`

openqa-qam-sle
manfredi: Proposals / Tips for renaming the script:
-  `openqa-search-maintenance-core-jobs`
- `openqa-search-regression-tests-core`

okurz: `openqa-search-maintenance-core-jobs` is ok

manfredi: renamed to `openqa-search-maintenance-core-jobs`

openqa-search-maintenance-core-jobs
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env bash
+
+set -e
```
okurz, line 3: From my comment before: Consider ´set -euo pipefail`

openqa-search-maintenance-core-jobs
```diff
@@ -0,0 +1,209 @@
...
+    ["15-SP6"]=546
+    ["12-SP3"]=106
+    ["12-SP5"]=282
+    ["15-SP4-TERADATA"]=521
+    ["12-SP3-TERADATA"]=191
+)
+
+readonly url_dashboard_qam='http://dashboard.qam.suse.de'
+readonly url_openqa='https://openqa.suse.de'
+readonly url_qam='https://qam.suse.de'
```
okurz, line 41: https://github.com/os-autoinst/scripts/pull/326#discussion_r1673814934 still applies

okurz (changes requested):

openqa-search-maintenance-core-jobs
manfredi: these are declared as global vars and used inside the functions as is, without any override with local vars

openqa-search-maintenance-core-jobs
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env bash
+
+set -e
+
+usage() {
+    cat << EOF
+Usage: $(basename "$0") -a SUSE:Maintenance:II:RR [-d days]
+
+The goal of this script is basically to fill the section
+"regression tests" in the log of QA test report,
```
Martchus, line 10:
```suggestion
"regression tests" in the log of QA test reports,
```

openqa-search-maintenance-core-jobs
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env bash
+
+set -e
+
+usage() {
+    cat << EOF
+Usage: $(basename "$0") -a SUSE:Maintenance:II:RR [-d days]
+
+The goal of this script is basically to fill the section
```
Martchus, line 9:
```suggestion
The goal of this script is to fill the section
```

openqa-search-maintenance-core-jobs
```diff
@@ -0,0 +1,209 @@
...
+"regression tests" in the log of QA test report,
+for a given update inside job groups:
+- Maintenance: Single Incidents / Core Incidents
+- Maintenance: Aggregated updates / Core Maintenance Updates
+and search for regression tests related with the update.
```
Martchus, line 14: Maybe it makes sense to start a new sentence.

openqa-search-maintenance-core-jobs
```diff
@@ -0,0 +1,209 @@
...
+Default search within the last 5 days for "Aggregated updates" builds
```
Martchus, line 15:
```suggestion
It searches by default within the last 5 days for "Aggregated updates" builds.
```

openqa-search-maintenance-core-jobs
```diff
@@ -0,0 +1,209 @@
...
+
+Usage: eg
```
Martchus, line 17:
```suggestion
Usage examples:
```

I would also cut this list short. Maybe just one line without and one with the `-d` parameter.

openqa-search-maintenance-core-jobs
manfredi: Changed

openqa-search-maintenance-core-jobs
manfredi: Changed

openqa-search-maintenance-core-jobs
manfredi: Changed

openqa-search-maintenance-core-jobs
manfredi: Changed

openqa-qam-sle
manfredi: Changed description

openqa-search-maintenance-core-jobs
okurz: oh geez, again wrong conversation, my bad. I meant https://github.com/os-autoinst/scripts/pull/326#discussion_r1673818872

okurz: Just use variables with default assignments like in https://github.com/os-autoinst/scripts/blob/master/monitor-openqa_job#L15 so that anyone calling those scripts can override on the fly, e.g. `url_openqa=https://openqa.opensuse.org openqa-search-maintenance-core-jobs`

manfredi: Added default values

okurz (approved):

Martchus (approved):

PR #327, opened 2024-07-11
Print warning messages to stderr

perlpunk:
So that jq doesn't receive it as its input

Issue: https://progress.opensuse.org/issues/163769


okurz (approved):
Again I would really prefer to have the git commit message state the "so what" rather than "what" but I don't want to annoy you so ... approved?

PR #329, opened 2024-07-22
Use runjq wrapper to see potential errors in API outputs

baierjan:
Reference: https://progress.opensuse.org/issues/163622

### DIFF
--- a/openqa-schedule-mm-ping-test
+++ b/openqa-schedule-mm-ping-test
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -eux -o pipefail
 
+. "$(dirname "$0")"/_common
+
 openqa_url=${openqa_url:-https://openqa.opensuse.org}
 distri=${distri:-opensuse}
 flavor=${flavor:-DVD}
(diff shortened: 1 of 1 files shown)


baierjan:
> I think for this we need a change on the gitlab side as well. There we only download the script instead of cloning the repo, so there is no _common file

Yes, that's correct. That change is already done; I just reused the framework which was added by Nick. The pipeline is cloning this whole repo.

perlpunk:
> Yes, that's correct. That change is already done; I just reused the framework which was added by Nick. The pipeline is cloning this whole repo.

Ok, I didn't see any change in the script-ci repo...
I guess it's a bit confusing that the pipelin clones the repo and then still downloads the script with curl...

perlpunk (approved):

baierjan:
> Ok, I didn't see any change in the script-ci repo... I guess it's a bit confusing that the pipelin clones the repo and then still downloads the script with curl...

All changes are in the GUI (additional variables in the pipeline schedule); it could be, but I guess it was easier to support both modes in this way 🤷🏻

kalikiana (approved):

PR #332, opened 2024-07-23
Rewrite openqa-advanced-retrigger-jobs in python

okurz:


perlpunk (commented):
It would still be good to have an example. I remember that it was complicated to get the quoting right

okurz:
> It would still be good to have an example. I remember that it was complicated to get the quoting right

An example for what? Do you mean an advanced example call in the help text?

openqa-advanced-retrigger-jobs
```diff
@@ -1,16 +1,88 @@
...
+    additional_filters = f" and {args.additional_filters}" if args.additional_filters else ""
+    client_prefix = "echo" if args.dry_run else ""
+
+    query = (
+        f"select id from jobs where ({worker_string}result='{args.result}' "
+        f"and clone_id is null and t_finished >= '{args.failed_since}'{additional_filters});"
+    )
+
+    log.debug(f"Using SQL query: '{query}' on {args.host}")
+    ssh_command = f'ssh {args.host} "sudo -u geekotest psql --no-align --tuples-only --command=\\"{query}\\" openqa"'
```
asdil12, line 78: Better use an array instead of a space separated string, otherwise a shell needs to parse it.

perlpunk:
> An example for what? Do you mean an advanced example call in the help text?

yes

perlpunk:
Also how about moving scripts to `bin/` and adding an extension, `.py`. I don't have a problem with extensions, and it makes tooling like linting easier

okurz:
> Also how about moving scripts to `bin/` and adding an extension, `.py`. I don't have a problem with extensions, and it makes tooling like linting easier

An extension would mean that we would need a separate installation tooling and step to install as executable without extension and I would like to avoid that.

okurz:
> It would still be good to have an example. I remember that it was complicated to get the quoting right

Done

openqa-investigate
```diff
@@ -1,391 +1,178 @@
...
-        client_prefix="${client_prefix:-runcli}"
-        client_call=("$client_prefix" "${client_call[@]}")
-    fi
-    set -u
-    clone_call="${clone_call:-"$client_prefix openqa-clone-job --json-output --skip-chained-deps --max-depth 0 --parental-inheritance --within-instance"}"
-    investigate "$@"
-}
-
-caller 0 > /dev/null || main "$@"
+import requests
```
sqozz, line 1: You're missing a shebang here - might help github to detect the used language as well

nicksinger, line 1: You're missing a shebang here - might help github to detect the used language as well

openqa-investigate
```diff
@@ -1,391 +1,178 @@
...
+
+def runcli(args):
+    try:
+        result = subprocess.run(args, capture_output=True, text=True, check=True)
+        return result.stdout
+    except subprocess.CalledProcessError as e:
+        warn(f"Command {args} failed with error: {e.stderr}")
+        return None
+
+def runjq(input_data, jq_filter):
```
perlpunk, line 22: Functions like this should be put into a library, not in the script directly. For bash we already have the `_common` file. We don't want to reimplement this in every script.
And I still think scripts should go into a bin folder.
The top directory is a mess.
Also tests are failing now.

openqa-investigate
```diff
@@ -1,391 +1,178 @@
...
+    try:
+        result = subprocess.run(['jq', jq_filter], input=input_data, capture_output=True, text=True, check=True)
+        return result.stdout
+    except subprocess.CalledProcessError as e:
+        warn(f"jq error: {e.stderr}")
```
perlpunk, line 27: Will we actually get the original input here? That's what the bash function is doing currently. If jq fails we want to know what the input was.

openqa-investigate
okurz: Sorry, please ignore openqa-investigate. That shouldn't have been part of this PR. a draft of openqa-investigate based on AI support is in
* #334

openqa-investigate
okurz: -> https://github.com/os-autoinst/scripts/pull/332?notification_referrer_id=NT_kwDOABnW-LMxMTYzMjQ3OTAzMjoxNjkzNDMy&notifications_query=is%3Aunread#discussion_r1689552634

openqa-investigate
okurz: -> https://github.com/os-autoinst/scripts/pull/332?notification_referrer_id=NT_kwDOABnW-LMxMTYzMjQ3OTAzMjoxNjkzNDMy&notifications_query=is%3Aunread#discussion_r1689552634

openqa-advanced-retrigger-jobs
okurz: done

okurz:
Updated:
* Simple and advanced example
* Converted single string in `subprocess.run` with `shell=True` into list (or actually a set). For that I extracted the `post` function so that I can just print in case of "dry-run" instead of the "client prefix" which was previously used to "echo" instead in bash.

Tested manually with

```
while_inotifywait sh -c "python3.11 ./openqa-advanced-retrigger-jobs -vvvv --host openqa.suse.de --failed-since \"2024-07-23T13:00\" --result failed --additional-filters \"test not like '%:investigate:%' and id=14995768\" --comment \"my foo comment\" --dry-run && python3.11 ./openqa-advanced-retrigger-jobs -vvvv --host openqa.suse.de --failed-since \"2024-07-23T13:00\" --result failed --additional-filters \"test not like '%:investigate:%' and id=14995768\" --comment \"my foo comment\""
```

openqa-advanced-retrigger-jobs
```diff
@@ -1,16 +1,105 @@
...
+def main(args):
+    log.debug(args)
+    instance_string = f" and instance='{args.instance}'" if args.instance else ""
+    worker_string = f"assigned_worker_id in (select id from workers where (host='{args.worker}'{instance_string})) and " if args.worker else ""
+    additional_filters = f" and {args.additional_filters}" if args.additional_filters else ""
+
+    query = (
+        f"select id from jobs where ({worker_string}result='{args.result}' "
+        f"and clone_id is null and t_finished >= '{args.failed_since}'{additional_filters});"
+    )
```
Martchus, line 92: Normally I'd say this needs escaping but of course the previous script also didn't have that…

openqa-advanced-retrigger-jobs
```diff
@@ -1,16 +1,105 @@
...
+    }
+    logging_level = logging.DEBUG if args.verbose > 4 else verbose_to_log[args.verbose]
+    log.setLevel(logging_level)
+    return args
+
+def post(dry_run: bool, host: str, route: str, *args):
+    cmd=('openqa-cli', 'api', '--host', host, '-X', 'POST', route, *args)
+    if dry_run:
+        return print(f"dry run: {cmd}")
+    subprocess.run(cmd)
```
Martchus, line 81: To retain the behavior on errors of the old script:
```suggestion
    subprocess.run(cmd, check=True)
```

openqa-advanced-retrigger-jobs
```diff
@@ -1,16 +1,105 @@
...
+    )
+    args = parser.parse_args()
+    verbose_to_log = {
+        0: logging.CRITICAL,
+        1: logging.ERROR,
+        2: logging.WARN,
+        3: logging.INFO,
+        4: logging.DEBUG,
+    }
+    logging_level = logging.DEBUG if args.verbose > 4 else verbose_to_log[args.verbose]
```
Martchus, line 73: Maybe this is not very explicit but I like it (considering numerical values for log levels are [documented explicitly](https://docs.python.org/3/library/logging.html#logging-levels)):
```suggestion
    logging_level = (5 - min(args.verbose, 4)) * 10
```

openqa-advanced-retrigger-jobs
okurz: I don't know what additional escaping you mean. Why do you mean we *need* escaping?

nicksinger: I think he means something in the lines of https://www.psycopg.org/psycopg3/docs/basic/params.html#execute-arguments

Currently e.g. `args.result` could be used to insert arbitrary SQL statements

nicksinger: While malice is very unlikely (given you'd need access anyway) it could help if people accidentally pass arguments which cause the query to break

Martchus: And one can also (accidentally) break out of the whole `psql` invocation.

PR #335, opened 2024-08-05
Allow labeling checks even when autoinst-log.txt does not exist

b10n1k:
Explicit call label_on_issues_from_issue_tracker for a job assuming that
autoinst-log.txt is already lost. Action is taken  a lit premature before                                                                                                                                                                                                      
further checks but ensures the flow doesnt change. 
 https://progress.opensuse.org/issues/164296

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -33,10 +33,9 @@ trap 'test "$KEEP_REPORT_FILE" == "1" || rm "$out"' EXIT
 
 echoerr() { echo "$@" >&2; }
 
-handle_unreachable_or_no_log() {
-    local id=$1
-    local testurl=$2
-    local out=$3
+handle_unreachable() {
+    local testurl=$1
+    local out=$2
 
     if ! curl "${curl_args[@]}" -s --head "$testurl" -o /dev/null; then
         # the page might be gone, try the scheme+host we configured (might be different one though)
@@ -58,13 +57,7 @@ handle_unreachable_or_no_log() {
         curl "${curl_args[@]}" "$testurl"
         exit 2
     fi
-    # if the page is there but not even an autoinst-log.txt exists
-    # then the job might be too old and logs are already deleted.
-    # Checking timestamp
-    if [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
-        echoerr "'$testurl' does not have autoinst-log.txt but is rather old, ignoring"
-        return
-    fi
+
     if hxnormalize -x "$out" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
         "${client_call[@]}" -X POST jobs/"$id"/comments text='poo#62456 test incompletes after failing in GRU download task on "Inactivity timeout" with no logs'
         "${client_call[@]}" -X POST jobs/"$id"/restart
@@ -143,7 +136,17 @@ investigate_issue() {
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
-        handle_unreachable_or_no_log "$id" "$testurl" "$out"
+        handle_unreachable "$testurl" "$out"
+        # Assume autoinst-log.txt still not found but process job anyway due to reason var
+        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
+            label_on_issues_from_issue_tracker "$id"
+        # Checking timestamp
+        elif [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
+            # if the page is there but not even an autoinst-log.txt exists
+            # then the job might be too old and logs are already deleted.
+            echoerr "'$testurl' does not have autoinst-log.txt but is rather old, ignoring"
+            return
+        fi
     elif label_on_issues_from_issue_tracker "$id"; then
         return
 


openqa-label-known-issues
```diff
@@ -144,7 +144,8 @@ investigate_issue() {
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
         handle_unreachable_or_no_log "$id" "$testurl" "$out"
-    elif label_on_issues_from_issue_tracker "$id"; then
+    fi
+    if label_on_issues_from_issue_tracker "$id"; then
```
perlpunk, line 148: There is already handling in `handle_unreachable_or_no_log` that will additionally check if the job itself is deleted or not, so I think we shouldn't unconditionally go on here.

I don't have a good suggestion right now for this though.

b10n1k: But I didnt change anything in this logic. it is as it was before. And sure to take a look, but it doesnt sound direct connected to the problem in the issue's description. 

I will quote this comment in the ticket

b10n1k: tried to improve on https://github.com/os-autoinst/scripts/commit/271c4247ed9b7df701a1cab3f354013639799b58

openqa-label-known-issues
```diff
@@ -142,6 +142,10 @@ investigate_issue() {
     # see https://progress.opensuse.org/issues/69178
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
+        # Assume autoinst-log.txt still not found but process job anyway due to reason var
+        if [[ -n $reason ]] && [[ "$curl_output" = "404" ]] && [[ $reason != null ]]; then
```
Martchus, line 146: For the sake of consistency:
```suggestion
        if [[ -n $reason ]] && [[ $curl_output = 404 ]] && [[ $reason != null ]]; then
```

openqa-label-known-issues
```diff
@@ -142,6 +142,10 @@ investigate_issue() {
...
+            label_on_issues_from_issue_tracker "$id"
```
Martchus, line 147: I'm not really familiar with this script at this point but don't we need the early `return` here as well? And if we have it, we still somehow need to ensure `label_on_issues_without_tickets` and `handle_unreviewed` are called in case the other function doesn't find anything. So *maybe* this need a little bit more restructuring.

b10n1k: There is no `return` because I believe we still need to go through `handle_unreachable_or_no_log` which will perform a return anyway. And IIUC the second part is covered in this case, as it continues to follow the workflow as before. right?

openqa-label-known-issues
b10n1k: e87edaa done

kalikiana: I don't see that change applied. There's still `"` in the code 🤔

openqa-label-known-issues
Martchus: This question should be clarified via https://progress.opensuse.org/issues/164296#note-13.

openqa-label-known-issues
```diff
@@ -142,6 +142,10 @@ investigate_issue() {
     # see https://progress.opensuse.org/issues/69178
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
+        # Assume autoinst-log.txt still not found but process job anyway due to reason var
+        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
+            label_on_issues_from_issue_tracker "$id"
+        fi
         # if we can not even access the page it is something more critical
         handle_unreachable_or_no_log "$id" "$testurl" "$out"
```
okurz, line 150: https://progress.opensuse.org/issues/164296#note-13

openqa-label-known-issues
b10n1k: Following the discussion and the comments, new commit 9db6f7d

and dry_run executions 
```
    ~/Documents/Work/qatools/repos/scripts on    164296_allow_known_issues *1 !1                                                                                                                                                                           at   07:32:42
❯ dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/14897579
Requesting jobs/14897579 via openqa-cli
openqa-cli api --header User-Agent: openqa-label-known-issues (https://github.com/os-autoinst/scripts) --host https://openqa.suse.de --retries=3 -X POST jobs/14897579/comments text=poo#73375 Job incompletes with reason auto_review:"(?m)api failure$" (and no further details)
    ~/Documents/Work/qatools/repos/scripts on    164296_allow_known_issues *1 !1                                                                                                                                                                           at   07:39:42
❯ dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/15089626
Requesting jobs/15089626 via openqa-cli
[https://openqa.suse.de/tests/15089626](https://openqa.suse.de/tests/15089626): Unknown test issue, to be reviewed
-> [autoinst-log.txt](https://openqa.suse.de/tests/15089626/file/autoinst-log.txt)

    # --- 8< ---
    # [2024-08-05T12:10:26.164521+02:00] [debug] [pid:126110] QEMU status is not 'shutdown', it is 'running'
    # [2024-08-05T12:10:26.164636+02:00] [debug] [pid:125698] backend shutdown state: 
    # [2024-08-05T12:10:26.164988+02:00] [info] [pid:126110] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
    # [2024-08-05T12:10:27.220847+02:00] [debug] [pid:126110] Passing remaining frames to the video encoder
    # frame=  771 fps=1.9 q=0.0 Lsize=    1531kB time=00:00:32.08 bitrate= 390.9kbits/s speed=0.0777x    
    # video:1525kB audio:0kB subtitle:0kB other streams:0kB global headers:0kB muxing overhead: 0.355806%
    # [2024-08-05T12:10:27.961098+02:00] [debug] [pid:126110] Waiting for video encoder to finalize the video
    # [2024-08-05T12:10:27.961202+02:00] [debug] [pid:126110] The built-in video encoder (pid 126494) terminated
    # [2024-08-05T12:10:27.961265+02:00] [debug] [pid:126110] The external video encoder (pid 126492) terminated
    # [2024-08-05T12:10:27.961832+02:00] [debug] [pid:126110] QEMU: qemu-system-x86_64: terminating on signal 15 from pid 126110 (/usr/bin/isotovideo: backen)
    # --- >8 ---
    ~/Documents/Work/qatools/repos/scripts on    164296_allow_known_issues *1 !1                                                                                                                                                                           at   07:39:56
❯ dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/14890572
Requesting jobs/14890572 via openqa-cli
'https://openqa.suse.de/tests/14890572' does not have autoinst-log.txt but is rather old, ignoring
```

openqa-label-known-issues
b10n1k: I can see it https://github.com/os-autoinst/scripts/pull/335/files#diff-4b82b9655e797531626dd3da7dce9a5f1e1bcade526ba3ad0262a1a4569a3570R141

openqa-label-known-issues
```diff
@@ -143,7 +136,17 @@ investigate_issue() {
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
-        handle_unreachable_or_no_log "$id" "$testurl" "$out"
+        handle_unreachable "$testurl" "$out"
+        # Assume autoinst-log.txt still not found but process job anyway due to reason var
+        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
+            label_on_issues_from_issue_tracker "$id"
```
okurz, line 142: For consistency this could be

```suggestion
            label_on_issues_from_issue_tracker "$id"
            return
```

However in practice this does not matter as we don't do any other action within `investigate_issue` after here anyway.

okurz (approved):

openqa-label-known-issues
```diff
@@ -143,7 +136,17 @@ investigate_issue() {
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
-        handle_unreachable_or_no_log "$id" "$testurl" "$out"
+        handle_unreachable "$testurl" "$out"
```
perlpunk, line 139: `handle_unreachable` does not always exit, so we would continue even if the url is unreachable? Maybe you can influence that with a `return ...` and check the return value?

b10n1k: Looking. I considered this but the examples I run didnt fall into this case.

b10n1k: First the function didnt have any exit even before the refactoring. 
But where do you want it to exit. `handle_unreachable` looks to me more like a diagnostic operation. it should continue unless...?

perlpunk: it didn't need to exit before, because it was in a branch that did not do anything after that

perlpunk: But if @okurz is ok with it, then ignore me ;-)
He wrote the original function.

b10n1k: ok. but not sure I follow. 

when I stop vpn and try to access url I get 
```❯ dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/14890572
Requesting jobs/14890572 via openqa-cli
Can't connect: Name or service not known
./openqa-label-known-issues: ERROR: line 124
./openqa-label-known-issues: ERROR: line 124
```
So does this even used if the url is unreachable anyway?

okurz: As I tried to explain when we talked about it in Jitsi you can't reach `handle_unreachable` if the server is not reachable at all. The `handle_unreachable` tries to handle an intermittent outage or other problems. I guess we can accept the PR as is even if there is a slight unexpected change of behaviour in an execution branch which in practice might not be reachable anymore.

b10n1k: But i still need to understand where is it expected to exit? I am gonna check again on weekend

b10n1k: Not an expert but the `-e` makes the script behaves a little different and the return code from the `handle_unreachable` is not return as expected. I used also exits instead of the returns but the script output shows some errors, as it seems to execute something additional after the last statement. You can see this by adding `-xv` to the script.
So if @perlpunk is ok with you, can we proceed and accept the current changes and maybe create another ticket to refactor/review the structure of the script later?

openqa-label-known-issues
Martchus: There are still quotes around the `404`. I guess it doesn't really matter.

perlpunk (approved):

PR #341, opened 2024-08-23
Cleanup unused variable from handle_unreviewed

b10n1k:


_common
```diff
@@ -171,7 +171,6 @@ handle_unreviewed() {
     local testurl=$1 out=$2 reason=$3 group_id=$4 email_unreviewed=$5 from_email=$6 notification_address=${7:-} job_data=${8:-} dry_run=${9:-}
     local group_data group_name group_description group_mailto header email excerpt=$snip_start
     local job_name job_result info
-    to_review+=("$testurl ${reason:0:50}")
```
baierjan, line 174: https://github.com/os-autoinst/scripts/blob/37be6df7f557ec0f5b43e45f77665cb30e792ff6/openqa-label-known-issues-multi#L16

baierjan (changes requested):

_common
kalikiana: How about ensuring this is covered in a test before making this change?

Martchus: To me it looks like we don't want this change at all because the variable is in fact used.

Martchus (changes requested):

_common
perlpunk: Nice catch @baierjan 
I thought we were using a bash option to prevent undeclared variables but I guess it doesn't work that way here...

baierjan: If I am not mistaken https://github.com/os-autoinst/scripts/blob/37be6df7f557ec0f5b43e45f77665cb30e792ff6/openqa-label-known-issues-multi#L10 counts as a declaration 😄 

The catch is, that the variable is filled as a kind of "side-effect" of `label_issue` call. Fortunately, it is still kind of clear once you try `git grep` for that variable.

perlpunk: But when this is called by `openqa-label-known-issues` and not by `scripts/openqa-label-known-issues-multi` there is no initial declaration.

But now I know - in `openqa-investigate ` for example we are using `-u` (`set -euo pipefail`)- but not in `openqa-label-known-issues`.
Adding `-u` there should make it fail.

baierjan: Right, other scripts sourcing `_common` are missing the declaration.

b10n1k:
❯ dry_run=1 echo https://openqa.opensuse.org/tests/4425222 | ./openqa-label-known-issues-multi
Requesting jobs/4425222 via openqa-cli
[https://openqa.opensuse.org/tests/4425222](https://openqa.opensuse.org/tests/4425222): Unknown test issue, to be reviewed
-> [autoinst-log.txt](https://openqa.opensuse.org/tests/4425222/file/autoinst-log.txt)

    # --- 8< ---

    # --- >8 ---

1 unknown issues to be reviewed:
 - https://openqa.opensuse.org/tests/4425222 preparation failed: Error detecting remote default
    ~/Documents/Work/qatools/repos/scripts on    url_err_handling *1                                                                                                                                                                                       at   13:24:28
❯ dry_run=1 echo https://openqa.opensuse.org/tests/4425222 | ./openqa-label-known-issues-multi
Requesting jobs/4425222 via openqa-cli
[https://openqa.opensuse.org/tests/4425222](https://openqa.opensuse.org/tests/4425222): Unknown test issue, to be reviewed
-> [autoinst-log.txt](https://openqa.opensuse.org/tests/4425222/file/autoinst-log.txt)

    # --- 8< ---

    # --- >8 ---

invalid as actually it is used.

PR #342, opened 2024-08-23
Handle errors from the label_on_issues_from_issue_tracker

b10n1k:
When `label_on_issues_from_issue_tracker` fails means that it doesn't add a
label on the ticket. Not added a comment it should not signal an ERR which
triggers the trap to be executed.
The current commit tries to solve again the problem which
https://github.com/os-autoinst/scripts/pull/335 solved but introduced a
new bug with a bad design.
After this commit the `label_on_issues_from_issue_tracker` is removed from the
_if_ block which caused early exit from trap when exit code was non-zero.
Redesigned in a simplier readable format keeping the logic the same and included
the requirements intacted. `label_on_issues_from_issue_tracker` is accessed
when autoinst-log exits or reason is found.
In case of some curl issues, no labeling occurs when the job is
old based on a hard-coded variable or when none of the requisites found.
Finally if none of them matches the issues the script will continue
with the next checks in the sequence of actions, as it was before.

https://progress.opensuse.org/issues/165716

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -6,7 +6,7 @@
 set -o pipefail -o errtrace
 
 # shellcheck source=/dev/null
-. "$(dirname "$0")"/_common
+. "$(dirname "${BASH_SOURCE[0]}")"/_common
 
 host="${host:-"openqa.opensuse.org"}"
 scheme="${scheme:-"https"}"
@@ -41,14 +41,14 @@ handle_unreachable() {
         # the page might be gone, try the scheme+host we configured (might be different one though)
         if ! grep -q "$host_url" <<< "$testurl"; then
             echoerr "'$testurl' is not reachable and 'host_url' parameter does not match '$testurl', can not check further, continuing with next"
-            return
+            return 1
         fi
         if ! curl "${curl_args[@]}" -s --head "$host_url"; then
             echoerr "'$host_url' is not reachable, bailing out"
             curl "${curl_args[@]}" --head "$host_url"
         fi
         echoerr "'$testurl' is not reachable, assuming deleted, continuing with next"
-        return
+        return 1
     fi
     # resorting to downloading the job details page instead of the
     # log, overwrites $out
@@ -61,7 +61,7 @@ handle_unreachable() {
     if hxnormalize -x "$out" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
         "${client_call[@]}" -X POST jobs/"$id"/comments text='poo#62456 test incompletes after failing in GRU download task on "Inactivity timeout" with no logs'
         "${client_call[@]}" -X POST jobs/"$id"/restart
-        return
+        return 1
     fi
 }
 
@@ -136,19 +136,23 @@ investigate_issue() {
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
-        handle_unreachable "$testurl" "$out"
-        # Assume autoinst-log.txt still not found but process job anyway due to reason var
-        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
-            label_on_issues_from_issue_tracker "$id"
+        handle_unreachable "$testurl" "$out" || return
+        [[ $curl_output != 404 ]] && return
         # Checking timestamp
-        elif [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
+        if [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
             # if the page is there but not even an autoinst-log.txt exists
             # then the job might be too old and logs are already deleted.
-            echoerr "'$testurl' does not have autoinst-log.txt but is rather old, ignoring"
+            echoerr "'$testurl' job#${id} without autoinst-log.txt older than 14 days. Do not label"
             return
         fi
-    elif label_on_issues_from_issue_tracker "$id"; then
-        return
+        # not unreachable, no log, no reason, not too old
+        if [[ -z $reason ]] || [[ $reason = null ]]; then
+            echoerr "'$testurl' does not have autoinst-log.txt or reason, cannot label"
+            return
+        fi
+    fi
+
+    label_on_issues_from_issue_tracker "$id" && return
 
     ## Issues without tickets, e.g. potential singular, manual debug jobs,
     # wrong user settings, etc.
@@ -156,11 +160,9 @@ investigate_issue() {
     # $client_prefix curl -s -H "Content-Type: application/json" -X POST -H "X-Redmine-API-Key: $(sed -n 's/redmine-token = //p' ~/.query_redminerc)" --data '{"issue": {"project_id": 36, "category_id": 152, priority_id: 5, "subject": "test from command line"}}' https://progress.opensuse.org/issues.json
     # but we should check if the issue already exists, e.g. same
     # subject line
-    elif label_on_issues_without_tickets "$id"; then
-        return
-    else
-        handle_unreviewed "$testurl" "$out" "$reason" "$group_id" "$email_unreviewed" "$from_email" "$notification_address" "$job_data" "$dry_run"
-    fi
+    label_on_issues_without_tickets "$id" && return
+
+    handle_unreviewed "$testurl" "$out" "$reason" "$group_id" "$email_unreviewed" "$from_email" "$notification_address" "$job_data" "$dry_run"
 }
 
 label_issue() {
--- a/test/07-openqa-label-known-issues.t
+++ b/test/07-openqa-label-known-issues.t
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+
+source test/init
+bpan:source bashplus +err +fs +sym
+
+plan tests 15
+
+source openqa-label-known-issues
+client_args=(api --host "$host_url")
+
+export KEEP_REPORT_FILE=1
+
+client_output=''
+mock-client-output() {
+    client_output+="client_call $@"$'\n'
+    echo "$client_output\n"
+}
+
+client_call=(mock-client-output "${client_call[@]}")
+autoinst_log=$dir/data/04-autoinst.txt
+
+try-client-output() {
+    client_output=''
+    try "$*"' && echo "$client_output"'
+}
+openqa-cli() {
+    local id=$(basename "$4")
+    cat "$dir/data/$id.json"
+}
+curl() {
+    if [[ "$7" =~ /(404|414|101|102)/file/autoinst-log.txt ]]; then
+        echo -n "404"
+    else
+        echo -n "200"
+    fi
+}
+comment_on_job() {
+    local id=$1 comment=$2 force_result=${3:-''}
+    echo "$comment"
+}
+out=''
+hxnormalize() {
+    cat "$2"
+}
+hxselect() {
+    cat -
+}
+
+try investigate_issue
+is "$rc" 1 'id required'
+# error in tap output is from here
+
+issues="159876
+[security] test fails in krb auto_review:\"com:/tmp/nfsdir /tmp/mntdir' failed\":force_result:softfailed
+openqa-force-result
+137420
+[qe-sap] test fails in network_peering auto_review:\"az network vnet peering create.+failed\"
+action
+73375
+Job incompletes with reason auto_review:\"(?m)api failure$\" (and no further details)
+action"
+
+# test data with reason but 404 in curl respond
+id=404
+testurl="https://openqa.opensuse.org/tests/${id}"
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log and with reason in job_data'
+has "$got" "without autoinst-log.txt older than 14 days. Do not label" "exits succefully when is old job without autoinst-log.txt"
+
+# all assets are missing
+id=101
+testurl="https://openqa.opensuse.org/tests/${id}"
+# `cur_date` is used on the following 4 test cases
+cur_date=$(date +%F)
+sed -i "s/yyyy-mm-dd/${cur_date}/" "$dir/data/${id}.json"
++fs:mktemp
+tmplog=$temp
+echo -n "Result: <b>incomplete</b>, finished <abbr class=\"timeago\" title=\"${cur_date}T08:06:42Z\"</abbr>>" > $tmplog
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log but with reason in job_data'
+has "$got" "does not have autoinst-log.txt or reason, cannot label" "investigation exits when no reason and autoinst-log"
+# Cleanup 404.json
+sed -i "s/${cur_date}/yyyy-mm-dd/" "$dir/data/${id}.json"
+
+# Unknown reason - not included in issues
+id=102
+testurl="https://openqa.opensuse.org/tests/${id}"
+echo -n "Result: <b>incomplete</b>, finished <abbr class=\"timeago\" title=\"${cur_date}T08:06:42Z\"</abbr>>" > $tmplog
+echo -n "\nthe reason is whatever" >> $tmplog
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate no old issue with missing autoinst-log and unknown reason in job_data'
+has "$got" "Unknown test issue, to be reviewed" "investigation still label Unknown reason"
+
+id=414
+testurl="https://openqa.opensuse.org/tests/${id}"
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing old autoinst-log and without reason in job_data'
+has "$got" "does not have autoinst-log.txt or reason, cannot label" "investigation exits successfully when no reason and no autoinst-log"
+
+id=200
+testurl="https://openqa.opensuse.org/tests/${id}"
+cp $autoinst_log $tmplog
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with autoinst-log and without reason'
+has "$got" "test fails in network_peering" "investigation label job with matched autoinst-log context"
+
+# handle_unreview branch
+echo > "$tmplog"
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'job with empty autoinst-log checks unknown issue'
+has "$got" "Unknown test issue, to be reviewed" "investigation still label Unknown issue"
+
+echo -n "[error] Failed to download" > $out
+try-client-output investigate_issue $id
+is "$rc" 0 'job label without tickets'
+has "$got" "label:download_error potentially out-of-space worker?" "investistigation label correctly job without ticket"
--- a/test/data/200.json
+++ b/test/data/200.json
@@ -1,0 +1,23 @@
+{
+    "job": {
+	"id": 200,
+	"group_id": 326,
+	"parents": {
+            "Chained": [],
+            "Directly chained": [],
+            "Parallel": []
+	},
+	"result": "failed",
+	"settings": {
+            "ARCH": "x86_64",
+            "BUILD": "testing",
+            "DISTRI": "Tumbleweed",
+            "FLAVOR": "Online",
+            "WORKER_CLASS": "qemu_x86_64"
+	},
+	"state": "done",
+	"t_finished": "2024-08-19T16:53:05",
+	"t_started": "2024-08-19T16:45:53",
+	"test": "footest_failed"
+    }
+}
--- a/test/data/04-autoinst.txt
+++ b/test/data/04-autoinst.txt
@@ -1,0 +1,2 @@
+test 200 autoinst
+az network vnet peering create. something failed
--- a/test/data/101.json
+++ b/test/data/101.json
@@ -1,0 +1,23 @@
+{
+    "job": {
+	"id": 101,
+	"group_id": 326,
+	"parents": {
+            "Chained": [],
+            "Directly chained": [],
+            "Parallel": []
+	},
+	"result": "incomplete",
+	"settings": {
+            "ARCH": "x86_64",
+            "BUILD": "testing",
+            "DISTRI": "Tumbleweed",
+            "FLAVOR": "Online",
+            "WORKER_CLASS": "qemu_x86_64"
+	},
+	"state": "done",
+	"t_finished": "yyyy-mm-ddT16:53:05",
+	"t_started": "yyyy-mm-dd16:45:53",
+	"test": "footest_incomplete"
+    }
+}
--- a/test/data/102.json
+++ b/test/data/102.json
@@ -1,0 +1,24 @@
+{
+    "job": {
+	"id": 102,
+	"group_id": 326,
+	"parents": {
+            "Chained": [],
+            "Directly chained": [],
+            "Parallel": []
+	},
+	"result": "incomplete",
+	"reason": "cant find this",
+	"settings": {
+            "ARCH": "x86_64",
+            "BUILD": "testing",
+            "DISTRI": "Tumbleweed",
+            "FLAVOR": "Online",
+            "WORKER_CLASS": "qemu_x86_64"
+	},
+	"state": "done",
+	"t_finished": "yyyy-mm-ddT16:53:05",
+	"t_started": "yyyy-mm-ddT16:45:53",
+	"test": "footest_incomplete"
+    }
+}
--- a/test/data/404.json
+++ b/test/data/404.json
@@ -1,0 +1,24 @@
+{
+    "job": {
+	"id": 404,
+	"group_id": 326,
+	"parents": {
+            "Chained": [],
+            "Directly chained": [],
+            "Parallel": []
+	},
+	"result": "incomplete",
+	"reason": "api failure",
+	"settings": {
+            "ARCH": "x86_64",
+            "BUILD": "testing",
+            "DISTRI": "Tumbleweed",
+            "FLAVOR": "Online",
+            "WORKER_CLASS": "qemu_x86_64"
+	},
+	"state": "done",
+	"t_finished": "2023-08-19T16:53:05",
+	"t_started": "2023-08-19T16:45:53",
+	"test": "footest_incomplete"
+    }
+}
--- a/test/data/414.json
+++ b/test/data/414.json
@@ -1,0 +1,23 @@
+{
+    "job": {
+	"id": 414,
+	"group_id": 326,
+	"parents": {
+            "Chained": [],
+            "Directly chained": [],
+            "Parallel": []
+	},
+	"result": "failed",
+	"settings": {
+            "ARCH": "x86_64",
+            "BUILD": "testing",
+            "DISTRI": "Tumbleweed",
+            "FLAVOR": "Online",
+            "WORKER_CLASS": "qemu_x86_64"
+	},
+	"state": "done",
+	"t_finished": "2023-08-19T16:53:05",
+	"t_started": "2023-08-19T16:45:53",
+	"test": "footest_failed"
+    }
+}


openqa-label-known-issues
```diff
@@ -89,7 +89,7 @@ label_on_issues_from_issue_tracker() {
                 label-on-issue "$id" "$search" "$label" "${after//*\":retry*/1}" "$force_result" && break
             fi
         done
-    )
+    ) || echoerr "$FUNCNAME failed" && exit 1
```
perlpunk, line 92: no, `label_on_issues_from_issue_tracker` is supposed to "fail", but a failure is nothing fatal here.
failure means it didn't find a matching issue in progress.
In that case it should continue with the other `label_on_issues_*` functions.

Also, if we `exit 1` here, we gained nothing, because the hook script failed and we will get an alert.  It currently does already exit, that's what we want to prevent.

See https://progress.opensuse.org/issues/165716#note-9

perlpunk (changes requested):

openqa-label-known-issues
baierjan: I was just about to ask similar questions:
1. Why `exit`? That will not make the problem disappear, I would more likely expect something like skipping over problematic items.
2. The `||` part is after a subshell with while loop, that is not likely to fail here. What type of error are you trying to `catch` here?

baierjan: Also note that `$FUNCNAME` is an array.

b10n1k: @perlpunk I did some changes. But I am not sure that it is what you expect.

@baierjan 1. that was mistake from my side. I didnt pay attention that the function is in `if` 2. I think the problem is that the `label_on_issues_from_issue_tracker` "fails" and `trap` sees an ERR signal.

perlpunk (commented):
The current logic is:
1. `label_on_issues_from_issue_tracker`: Try to find a ticket in progress which matches something in `autoinst-log.txt` or the reason.
1. If it was not labeled, try the next method, which is `label_on_issues_without_tickets`. It does some hardcoded `grep`s
1. If that didn't find anything, call `handle_unreviewed`, which will
  * trigger investigation jobs
  * send an email about an unreviewed issue

I suppose we also want the second and third step for jobs that only have a reason but no `autoinst-log.txt`. That's a bit difficult with the current if/elif structure.

In any case, if `label_on_issues_from_issue_tracker` returns with a non-zero exit code, that's nothing bad that needs to be handled or logged. It's simply the answer whether it has found a matching ticket or not and the code needs to try the other methods or not. And I agree that's maybe not the best design, as there could also be some real error happening inside that function. That should probably be handled by using a certain custom return code. But that's out of scope here.

b10n1k:
> The current logic is:
> 
> 1. `label_on_issues_from_issue_tracker`: Try to find a ticket in progress which matches something in `autoinst-log.txt` or the reason.
> 2. If it was not labeled, try the next method, which is `label_on_issues_without_tickets`. It does some hardcoded `grep`s
> 3. If that didn't find anything, call `handle_unreviewed`, which will
>    
>    * trigger investigation jobs
>    * send an email about an unreviewed issue
> 
> I suppose we also want the second and third step for jobs that only have a reason but no `autoinst-log.txt`. That's a bit difficult with the current if/elif structure.
> 
> In any case, if `label_on_issues_from_issue_tracker` returns with a non-zero exit code, that's nothing bad that needs to be handled or logged. It's simply the answer whether it has found a matching ticket or not and the code needs to try the other methods or not. And I agree that's maybe not the best design, as there could also be some real error happening inside that function. That should probably be handled by using a certain custom return code. But that's out of scope here.

I kinda redesign the solution from https://github.com/os-autoinst/scripts/pull/335. It should look better now.
the tests I run:
- dry_run=1 ./openqa-label-known-issues https://openqa.opensuse.org/tests/4425222
- dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/15136477
- dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/14890572
- dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/14897579

openqa-label-known-issues
```diff
@@ -134,14 +134,16 @@ investigate_issue() {
     # against even in case when autoinst-log.txt is missing the details, e.g.
     # see https://progress.opensuse.org/issues/69178
     echo "$reason" >> "$out"
-    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
+    # Control flow when there is a curl issue but still want to process job due to reason var
+    # even if autoinst-log.txt not found but assume is missing.
+    [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]] \
+        && is_unreachachle_but_has_reason=1 \
+        || is_unreachachle_but_has_reason=0
+    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]] && [[ ! $is_unreachachle_but_has_reason ]]; then
```
baierjan, line 142: This is most probably not what you want. Your `[[ ! "$is_unreachachle_but_has_reason ]]`condition would be always false. You might want something like.

```suggestion
    [[ -n "$reason" ]] && [[ "$curl_output" = "404" ]] && [[ "$reason" != null ]] && is_unreachachle_but_has_reason=1
    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]] && [[ -z "$is_unreachachle_but_has_reason" ]]; then
```

Note: I suppose _null_ is just a plain string returned from the `jq`, right?

openqa-label-known-issues
```diff
@@ -134,14 +134,16 @@ investigate_issue() {
...
         # if we can not even access the page it is something more critical
         handle_unreachable "$testurl" "$out"
-        # Assume autoinst-log.txt still not found but process job anyway due to reason var
-        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
-            label_on_issues_from_issue_tracker "$id"
```
baierjan, line 142: So this is no longer called?

baierjan (changes requested):

openqa-label-known-issues
```diff
@@ -134,14 +134,16 @@ investigate_issue() {
...
         # Checking timestamp
-        elif [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
+        if [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
```
baierjan, line 146: I am not sure if I read all the conditions correctly, but within this block you need non-standard curl 
 output and `$is_unreachachle_but_has_reason == false`, which effectively means `$reason == ""` or `$reason == null`. `$out` is basically curl output plus `$reason`, which is in this special case empty or almost empty (as there is no `$reason`), that means `grep` can't do much in this case. Maybe this block needs to be one level higher?

openqa-label-known-issues
b10n1k: I think that null is indeed a plain string.
I guess I have to quote all the variables. 

But I dont get why `-z` or `condition would be always false`?! if you run the test cases, it is not always `false`

openqa-label-known-issues
b10n1k: it is not called here but it does run, when `no autoinst-log.txt but reason exist`. I found it weird that we had to handle that case in this block. I believe the approach now, it simplifies it a bit.

openqa-label-known-issues
b10n1k: I am not sure about that. This one is not modified at all from my side. and it is not directly correlated to the problem this commit tries to fix I think.

openqa-label-known-issues
baierjan: You have
```   
 [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]] \
        && is_unreachachle_but_has_reason=1 \
        || is_unreachachle_but_has_reason=0
```
which will either set `is_unreachachle_but_has_reason` to `1` or `0`, right? The following condition `[[ ! $is_unreachachle_but_has_reason ]]` is functionally equivalent of `[[ -! -n $is_unreachachle_but_has_reason ]]` which can be seen as `[[ -z $is_unreachachle_but_has_reason ]]`, in other other words will evaluate to false if the string has non-zero length. In your case it is `0` or `1` which is always a non-zero length string.

openqa-label-known-issues
baierjan: Well, you did change the condition here 😉

openqa-label-known-issues
b10n1k: https://github.com/os-autoinst/scripts/compare/1c803f6248b784d98ee87dc07abc12f398876b05..1208b09e6428bb97e65be90680b9cb8b2c3c9ed9 added the quotes

b10n1k: Done https://github.com/os-autoinst/scripts/compare/1208b09e6428bb97e65be90680b9cb8b2c3c9ed9..f6dfb2e88b82f1d9659759597095521048e4040b

openqa-label-known-issues
b10n1k: IIUC you are right. `$out` =  curl output plus `$reason` if any. However the condition is activated only for old jobs which found older than 14d. And only checked when the curl is not successful. Not successful means that it couldnt get the `autoinst-log.txt`.
Currently if the `autoinst-log.txt` not found but `$is_unreachachle_but_has_reason` is TRUE, the `label_on_issues_from_issue_tracker` branch should run

baierjan: I guess it is OK then? Would be nice to have it covered by some automated tests.

baierjan (dismissed):

openqa-label-known-issues
```diff
@@ -134,14 +134,16 @@ investigate_issue() {
     # against even in case when autoinst-log.txt is missing the details, e.g.
     # see https://progress.opensuse.org/issues/69178
     echo "$reason" >> "$out"
-    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
+    # Control flow when there is a curl issue but still want to process job due to reason var
+    # even if autoinst-log.txt not found but assume is missing.
+    [[ -n "$reason" ]] && [[ "$curl_output" = "404" ]] && [[ "$reason" != null ]] \
+        && is_unreachachle_but_has_reason=1 \
+        || is_unreachachle_but_has_reason=0
+    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]] && [[ -z $is_unreachachle_but_has_reason ]]; then
```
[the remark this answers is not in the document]
b10n1k: https://github.com/os-autoinst/scripts/compare/f6dfb2e88b82f1d9659759597095521048e4040b..3c3323df2b4a809cd68c4e8b29051ffaeb618614

openqa-label-known-issues
```diff
@@ -134,14 +134,16 @@ investigate_issue() {
     # against even in case when autoinst-log.txt is missing the details, e.g.
     # see https://progress.opensuse.org/issues/69178
     echo "$reason" >> "$out"
-    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
+    # Control flow when there is a curl issue but still want to process job due to reason var
+    # even if autoinst-log.txt not found but assume is missing.
+    [[ -n "$reason" ]] && [[ "$curl_output" = "404" ]] && [[ "$reason" != null ]] \
+        && is_unreachachle_but_has_reason=1 \
+        || is_unreachachle_but_has_reason=0
```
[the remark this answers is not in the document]
b10n1k: https://github.com/os-autoinst/scripts/compare/f6dfb2e88b82f1d9659759597095521048e4040b..3c3323df2b4a809cd68c4e8b29051ffaeb618614

```❯ dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/15216424
dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/14890572
dry_run=1 host=openqa.suse.de ./openqa-label-known-issues https://openqa.suse.de/tests/14897579
Requesting jobs/15216424 via openqa-cli
[https://openqa.suse.de/tests/15216424](https://openqa.suse.de/tests/15216424): Unknown test issue, to be reviewed
-> [autoinst-log.txt](https://openqa.suse.de/tests/15216424/file/autoinst-log.txt)

    # --- 8< ---
    # [2024-08-19T18:53:01.487177+02:00] [debug] [pid:48129] QEMU status is not 'shutdown', it is 'running'
    # [2024-08-19T18:53:01.487304+02:00] [debug] [pid:47872] backend shutdown state: 
    # [2024-08-19T18:53:01.487615+02:00] [info] [pid:48129] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
    # [2024-08-19T18:53:02.540944+02:00] [debug] [pid:48129] Passing remaining frames to the video encoder
    # frame=  774 fps=1.9 q=0.0 Lsize=    1733kB time=00:00:32.20 bitrate= 440.7kbits/s speed=0.0807x    
    # video:1727kB audio:0kB subtitle:0kB other streams:0kB global headers:0kB muxing overhead: 0.329332%
    # [2024-08-19T18:53:03.110926+02:00] [debug] [pid:48129] Waiting for video encoder to finalize the video
    # [2024-08-19T18:53:03.111018+02:00] [debug] [pid:48129] The external video encoder (pid 48320) terminated
    # [2024-08-19T18:53:03.111081+02:00] [debug] [pid:48129] The built-in video encoder (pid 48321) terminated
    # [2024-08-19T18:53:03.111636+02:00] [debug] [pid:48129] QEMU: qemu-system-x86_64: terminating on signal 15 from pid 48129 (/usr/bin/isotovideo: backen)
    # --- >8 ---
Requesting jobs/14890572 via openqa-cli
'https://openqa.suse.de/tests/14890572' does not have autoinst-log.txt but is rather old, ignoring
Requesting jobs/14897579 via openqa-cli
openqa-cli api --header User-Agent: openqa-label-known-issues (https://github.com/os-autoinst/scripts) --host https://openqa.suse.de --retries=3 -X POST jobs/14897579/comments text=poo#73375 Job incompletes with reason auto_review:"(?m)api failure$" (and no further details)
```

openqa-label-known-issues
```diff
@@ -134,14 +134,15 @@ investigate_issue() {
     # against even in case when autoinst-log.txt is missing the details, e.g.
     # see https://progress.opensuse.org/issues/69178
     echo "$reason" >> "$out"
-    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
+    # Control flow when there is a curl issue but still want to process job due to reason var
+    # even if autoinst-log.txt not found but assume is missing.
```
perlpunk, line 138: I don't understand that part of the sentence

openqa-label-known-issues
```diff
@@ -134,14 +134,15 @@ investigate_issue() {
...
+    [[ -n "$reason" ]] && [[ "$curl_output" = "404" ]] && [[ "$reason" != null ]] \
+        && is_unreachachle_but_has_reason=1
```
perlpunk, line 140: `is_unreachachle_but_has_reason` - did you mean `unreachable`?

openqa-label-known-issues
```diff
@@ -134,14 +134,15 @@ investigate_issue() {
...
+    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]] && [[ -z "$is_unreachachle_but_has_reason" ]]; then
```
perlpunk, line 141: With this change jobs that have a reason but no autoinst-log.txt are not handled in this branch anymore and checked if the job is maybe too old.
Is that the intention?

Also I find that long variable not very readable.
Also why not just use `true` and `false` to avoid having to think about `-z`?

```
var=true
something && var=false
# later
if $var && ...
```
Also, inside `[[ ]]` brackets quotes aren't needed for these things. Not sure why you added them again.

openqa-label-known-issues
Martchus: I'm also not sure how to interpret the sentence. It isn't grammatically correct.

I would also avoid using the expression "Control flow when …" because it is very clunky. Just start with a common "handle the case when …" or maybe you can come up with something less generic.

openqa-label-known-issues
Martchus: Yes, unreachable sounds like the wrong variable name.

You can just assign `1` or an empty string in my opinion. The use of `-z` for the sake of negation is also ok. But I'd also really like to avoid the quotes on the lhs within `[[ ]]` (and you can also remove them on the rhs if not necessary like here).

b10n1k: > Is that the intention?

Not sure how correct is it but yes. The logic I followed was for any not successful curl_output keep the handling as is. but skip it if there is at least an reason.

 > Also I find that long variable not very readable

I changed it. hopefully in something more meaningful

> inside [[ ]] brackets quotes aren't needed for these things. Not sure why you added them again.

That's partly my mistake. @baierjan asked me to add them but you had come before him asking to remove them.
The thing is that I know that in general is good practice to always use the quotes. So is there any link with what is good practice here to follow?

openqa-label-known-issues
b10n1k: @Martchus I changed that. Also I reply here https://github.com/os-autoinst/scripts/pull/342#discussion_r1733998927

openqa-label-known-issues
perlpunk: Regarding `[[` in bash: https://stackoverflow.com/questions/3427872/whats-the-difference-between-and-in-bash/3427931#3427931

openqa-label-known-issues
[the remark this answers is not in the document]
perlpunk: We don't need quotes in double square brackets

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,84 @@
...
+openqa-cli() {
+    local id=$(basename "$4")
+    cat "$dir/test/data/$id.json"
+}
+curl() {
+    warn "URL $7"
+    if [[ "https://openqa.opensuse.org/tests/404/file/autoinst-log.txt" = "$7" ]]; then
+	echo "404"
+    fi
+    echo "200"
```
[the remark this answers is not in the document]
perlpunk: If you're debugging variables, it can really help to put quotes around it to see where it starts and endes. Otherwise it can easily be overlooked that `$curl_output` has the wrong content here.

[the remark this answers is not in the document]
baierjan: Also there is no new line at the end when calling `curl -w "%{http_code}"` so you likely want to use `echo -n 200` to mimic that.

test/data/200.json
```diff
@@ -0,0 +1,36 @@
+{
```
perlpunk, line 1: Can this file be made a bit smaller? We don't need all of the fields for this test.

openqa-label-known-issues
```diff
@@ -134,14 +137,16 @@ investigate_issue() {
...
     echo "$reason" >> "$out"
-    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
+    # With 404 assume autoinst-log.txt is missing but we have a reason var
+    # to process the job
+    echo "DEBUG $reason .. $curl_output"
+    [[ -n "$reason" ]] && [[ "$curl_output" = "404" ]] && [[ "$reason" != null ]] \
+        && has_missing_logs_but_reason_found=1
+    if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]] && [[ "$has_missing_logs_but_reason_found" ]]; then
         # if we can not even access the page it is something more critical
         handle_unreachable "$testurl" "$out"
```
perlpunk, line 147: So now the `handle_unreachable` is only done when there is a reason found?

b10n1k: fixed and added new test cases.

test/data/200.json
b10n1k: Done

openqa-label-known-issues
perlpunk: Now `handle_unreachable` is only called if there is no `reason`?

b10n1k: yes but more accurate, it is called only when there is no `reason` and `autoinst-log.txt` isnt found. Which I think it makes sense.

perlpunk (commented):
I wrote in #343 that I think it reflects the logic that we want now.
It is different from this pull request.
If you agree with the logic in #343 I would suggest we take that one and test it. If you disagree, I'd like to know where.

b10n1k:
> I wrote in #343 that I think it reflects the logic that we want now. It is different from this pull request. If you agree with the logic in #343 I would suggest we take that one and test it. If you disagree, I'd like to know where.

https://github.com/os-autoinst/scripts/pull/343#issuecomment-2324800004

b10n1k:
As we reevaluate the poo I set the not-ready. I will need to adjust PR to the requirements

openqa-label-known-issues
```diff
@@ -136,31 +138,33 @@ investigate_issue() {
...
-        handle_unreachable "$testurl" "$out"
-        # Assume autoinst-log.txt still not found but process job anyway due to reason var
-        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
-            label_on_issues_from_issue_tracker "$id"
+        handle_unreachable "$testurl" "$out" || return
+        # If we something other than a 404, we should not continue
+        [[ $curl_output != 404 ]] && return
         # Checking timestamp
-        elif [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
+        if [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
```
Martchus, line 145: To me it looks like `out` contains the contents of autoinst-log.txt and the reason. This code is looking for a certain HTML element in `out` which looks wrong. Have you actually tested this? I don't see that the unit tests below cover anything containing `timeago`.

Considering you're not really introducing this line of code here it is probably not that important - but I cannot really make sense of the code.

openqa-label-known-issues
```diff
@@ -136,31 +138,33 @@ investigate_issue() {
...
     # $client_prefix curl -s -H "Content-Type: application/json" -X POST -H "X-Redmine-API-Key: $(sed -n 's/redmine-token = //p' ~/.query_redminerc)" --data '{"issue": {"project_id": 36, "category_id": 152, priority_id: 5, "subject": "test from command line"}}' https://progress.opensuse.org/issues.json
     # but we should check if the issue already exists, e.g. same
     # subject line
-    elif label_on_issues_without_tickets "$id"; then
-        return
-    else
-        handle_unreviewed "$testurl" "$out" "$reason" "$group_id" "$email_unreviewed" "$from_email" "$notification_address" "$job_data" "$dry_run"
-    fi
+    label_on_issues_without_tickets "$id" && return
+
```
Martchus, line 166:
```suggestion
```

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,109 @@
...
+sed -i "s/api failure//" "test/data/${id}.json"
+tmplog=/tmp/autoinst-404.txt
+echo > $tmplog
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log and unknown reason in job_data'
+has "$got" "does not have autoinst-log.txt or reason, cannot label" "investigation exits without reason and autoinst-log"
+sed -i "s/\"reason\": \"\"/\"reason\": \"api failure\"/" "test/data/${id}.json"
+
+#unknown reason
```
Martchus, line 73:
```suggestion
# unknown reason
```

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,109 @@
...
+id=404
+testurl="https://openqa.opensuse.org/tests/${id}"
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log and with reason in job_data'
+has "$got" "seems like an old job without autoinst-log.txt" "old job without autoinst-log.txt"
+has "$got" "Job incompletes with reason" "investigation keep running without autoinst-log"
+
+# all requisite evidence are missing
+sed -i "s/api failure//" "test/data/${id}.json"
+tmplog=/tmp/autoinst-404.txt
```
Martchus, line 65: It is not nice to hardcode paths like this but I guess I can live with that.

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,109 @@
...
+testurl="https://openqa.opensuse.org/tests/${id}"
+tmplog=/tmp/autoinst-200.txt
+cp $autoinst_log $tmplog
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with autoinst-log and without reason'
+has "$got" "test fails in network_peering" "investigation label job with matched autoinst-log context"
+
+# handle_unreview branch
+echo >${tmplog}
```
Martchus, line 99: This also needs quoting and should use the style you used before for consistency:
```suggestion
echo > "$tmplog"
```

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,109 @@
...
+testurl="https://openqa.opensuse.org/tests/${id}"
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log and with reason in job_data'
+has "$got" "seems like an old job without autoinst-log.txt" "old job without autoinst-log.txt"
+has "$got" "Job incompletes with reason" "investigation keep running without autoinst-log"
+
+# all requisite evidence are missing
+sed -i "s/api failure//" "test/data/${id}.json"
+tmplog=/tmp/autoinst-404.txt
+echo > $tmplog
```
Martchus, line 66: This and other occurrences should use quoting (even though the path doesn't contain any white-spaces right now):
```suggestion
echo > "$tmplog"
```
(You can of course skip the quoting on variable assignments like the in the next line.)

openqa-label-known-issues
```diff
@@ -136,31 +138,33 @@ investigate_issue() {
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
-        handle_unreachable "$testurl" "$out"
-        # Assume autoinst-log.txt still not found but process job anyway due to reason var
-        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
-            label_on_issues_from_issue_tracker "$id"
+        handle_unreachable "$testurl" "$out" || return
+        # If we something other than a 404, we should not continue
```
Martchus, line 142: This comment is not grammatically correct. I would just remove it completely because it does not add any information anyway. (The comment just repeats a simple line of code anybody can just read directly.)

```suggestion
```

Martchus (commented):
I guess the new tests are generally good. I only have a few remarks.

I'm really wondering how the `$(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)')` part is supposed to work but you're not really introducing that code.

openqa-label-known-issues
Martchus: I would also avoid quotes on the lhs of `[[` and on the rhs if the rhs value is a literal string without spaces.

openqa-label-known-issues
perlpunk: The content in `$out` is overwritten in `handle_unreachable`:
```
if ! curl "${curl_args[@]}" -s "$testurl" -o "$out"; then
```

Martchus: Ah, *of course* :-)

openqa-label-known-issues
```diff
@@ -136,31 +138,33 @@ investigate_issue() {
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
-        handle_unreachable "$testurl" "$out"
-        # Assume autoinst-log.txt still not found but process job anyway due to reason var
-        if [[ -n $reason ]] && [[ $curl_output = "404" ]] && [[ $reason != null ]]; then
-            label_on_issues_from_issue_tracker "$id"
+        handle_unreachable "$testurl" "$out" || return
```
perlpunk, line 141: You didn't actually change `handle_unreachable` like in #343 
If it doesn't return `1` then this `return` here is not changing anything.

I suggested that you could just cherry-pick my commit.

openqa-label-known-issues
```diff
@@ -5,8 +5,10 @@
 
 set -o pipefail -o errtrace
 
+dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+PATH=$dir:$PATH
```
perlpunk, line 9: What is that needed for?

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,109 @@
...
+export KEEP_REPORT_FILE=1
+
+client_output=''
+mock-client-output() {
+    client_output+="client_call $@"$'\n'
+    echo "$client_output\n"
+}
+
+client_call=(mock-client-output "${client_call[@]}")
+autoinst_log=$dir/test/data/04-autoinst.txt
```
perlpunk, line 19: `$dir` is set `in `test/init` to the test directory itself.

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,109 @@
...
+# test data with reason but 404 in curl respond
+id=404
+testurl="https://openqa.opensuse.org/tests/${id}"
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log and with reason in job_data'
+has "$got" "seems like an old job without autoinst-log.txt" "old job without autoinst-log.txt"
+has "$got" "Job incompletes with reason" "investigation keep running without autoinst-log"
+
+# all requisite evidence are missing
+sed -i "s/api failure//" "test/data/${id}.json"
```
perlpunk, line 64: Messing with files stored in git during a test is not a good idea.

When the test fails in between, you are left with modified files.

test/07-openqa-label-known-issues.t
perlpunk: At least it should be cleaned up at the end.
Have a look at e0ccf2e4a9e3b19442abec12b5cea6cf9c67f0de how we create temp files with the bashplus framework.
And I hope it wouldn't overwrite any file I manually created, so a more unique name would be good.

openqa-label-known-issues
perlpunk: And this was something I already suggested in the original PR: https://github.com/os-autoinst/scripts/pull/335#discussion_r1711300333

openqa-label-known-issues
perlpunk: This might actually be another thing that is now kind of broken after the #335.
If the logfile is overwritten, then trying to label it doesn't make sense anymore.
Maybe `handle_unreachable` and this check here shuold use a different filename.
Originally this check was inside of `handle_unreachable`, and the code didn't continue after that.

openqa-label-known-issues
perlpunk: I disagree.
We have a couple of top level `if`s here and I think it makes sense to add a blank line between all of them. And the final `handle_unreviewed`.
We were all struggling to understand the logic of the code.

test/07-openqa-label-known-issues.t
b10n1k: is not.  another solution is separate files. but they are gonna be duplicated. or copy them and remove the copies at the end. what do you think

test/07-openqa-label-known-issues.t
b10n1k: if you print it is `~/some_path/scripts`. what do you mean?

openqa-label-known-issues
b10n1k: I consider this closed as we have added tests. Lets me know in the latest commit.

openqa-label-known-issues
b10n1k: `return`  as is here, should return the last exit code, no?

openqa-label-known-issues
b10n1k: When the `test/07-openqa-label-known-issues.t` runs tries and tries `source openqa-label-known-issues`  and it cant find the `_common`

perlpunk: Try what we do in `openqa-investigate`:

```suggestion
. "$(dirname "${BASH_SOURCE[0]}")"/_common 
```

openqa-label-known-issues
perlpunk: What I mean is this and the other `return 1` lines inside of `handle_unreachable.
https://github.com/os-autoinst/scripts/pull/343/files#diff-4b82b9655e797531626dd3da7dce9a5f1e1bcade526ba3ad0262a1a4569a3570R44

The function needs to return 1 to indicate that we should not continue with labeling. Then the `|| return` will be executed.

test/07-openqa-label-known-issues.t
perlpunk: What I mean is, that `test/init` sets `dir=...`.
And it is set to the test directory itself. So `/some/path/scripts/test`.

You don't need to define `dir=...` in `openqa-label-known-issues`.
Remove that definition and then you can use
```suggestion
autoinst_log=$dir/data/04-autoinst.txt
```

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash -xv
```
baierjan, line 1: See https://www.shellcheck.net/wiki/SC2096
```suggestion
#!/usr/bin/env bash
set -xv
```
or
```suggestion
#!/usr/bin/bash -xv
```

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,114 @@
...
+echo -n "Result: <b>incomplete</b>, finished <abbr class=\"timeago\" title=\"${cur_date}T08:06:42Z\"</abbr>>"> $tmplog
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log but with reason in job_data'
+has "$got" "does not have autoinst-log.txt or reason, cannot label" "investigation exits when no reason and autoinst-log"
+# Cleanup 404.json
+sed -i "s/\"reason\": \"\"/\"reason\": \"api failure\"/" "test/data/${id}.json"
+sed -i "s/${cur_date}/2023-08-19/" "test/data/${id}.json"
+
+# Unknown reason - not icluded in issues
```
baierjan, line 79:
```suggestion
# Unknown reason - not included in issues 
```

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,114 @@
...
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with autoinst-log and without reason'
+has "$got" "test fails in network_peering" "investigation label job with matched autoinst-log context"
+
+# handle_unreview branch
+echo > "$tmplog"
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'job with emtpy autoinst-log checks unknown issue'
```
baierjan, line 108:
```suggestion
is "$rc" 0 'job with empty autoinst-log checks unknown issue'
```

test/07-openqa-label-known-issues.t
b10n1k: Actually those should both removed. I tried to used them for some debugging. they are leftovers. thanks man

test/07-openqa-label-known-issues.t
b10n1k: fec0624

test/07-openqa-label-known-issues.t
b10n1k: fec0624

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,114 @@
...
+cur_date=$(date +%F)
+sed -i "s/2023-08-19/${cur_date}/" "test/data/${id}.json"
++fs:mktemp; tmplog=$temp
+echo -n "Result: <b>incomplete</b>, finished <abbr class=\"timeago\" title=\"${cur_date}T08:06:42Z\"</abbr>>"> $tmplog
+out=$tmplog
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log but with reason in job_data'
+has "$got" "does not have autoinst-log.txt or reason, cannot label" "investigation exits when no reason and autoinst-log"
+# Cleanup 404.json
+sed -i "s/\"reason\": \"\"/\"reason\": \"api failure\"/" "test/data/${id}.json"
```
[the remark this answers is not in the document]
b10n1k: fec0624

test/07-openqa-label-known-issues.t
b10n1k: fec0624

dependencies.yaml
```diff
@@ -43,6 +43,7 @@ test_requires:
   perl(Test::Warnings):
   perl(Test::MockModule):
   perl(Test::Most):
+  html-xml-utils:
```
b10n1k, line 46: Why this doesnt work?

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
           python --version
           pytest --version
       - run: |
-          sudo apt-get install cpanminus
+          sudo apt-get install cpanminus html-xml-utils
```
b10n1k, line 15: I am not sure about that. shouldnt be https://open.spotify.com/track/3Le4N9DFC3fZQM7jS4gO6F?si=180845d991ae4b60 enough?

openqa-label-known-issues
b10n1k: I tried. And Unfortunately breaks some tests. I am not sure it is necessary. or.. is it?

openqa-label-known-issues
b10n1k: 27b7e99

openqa-label-known-issues
perlpunk: I tried it as well, and it worked for me

perlpunk: like I wrote below, you don't want to set `dir` here, because it is set by `test/init`. I see you are doing it to use it in the test, but it is not necessary, and confusing.
I don't know how to explain it better.

.github/workflows/ci.yml
perlpunk: I guess that's not the URL you meant to paste.

b10n1k: no I hope you like it tho :)

openqa-label-known-issues
b10n1k: 00a7075

test/07-openqa-label-known-issues.t
b10n1k: 00a7075

test/07-openqa-label-known-issues.t
b10n1k: Some changes regarding that on cc2c202. can improve further tho

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,113 @@
...
+try-client-output() {
+  client_output=''
+  try "$*"' && echo "$client_output"'
+}
+openqa-cli() {
+    local id=$(basename "$4")
+    cat "$dir/data/$id.json"
+}
+curl() {
+    if [[ "https://openqa.opensuse.org/tests/404/file/autoinst-log.txt" = "$7" ]] || [[ "https://openqa.opensuse.org/tests/414/file/autoinst-log.txt" = "$7" ]] || [[ "https://openqa.opensuse.org/tests/101/file/autoinst-log.txt" = "$7" ]] || [[ "https://openqa.opensuse.org/tests/102/file/autoinst-log.txt" = "$7" ]] ; then
```
perlpunk, line 32:
```suggestion
if [[ $7 =~ /(404|414|101|102)/file/autoinst-log.txt ]]; then
```

perlpunk (commented):
I have just this one suggestion left, and maybe run a `shfmt` on the new test file.
We currently don't shellcheck the files under `test`

test/07-openqa-label-known-issues.t
b10n1k: 7400c12

test/07-openqa-label-known-issues.t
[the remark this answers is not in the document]
kalikiana: @b10n1k I don't see this anywhere in the iff

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,114 @@
...
+action"
+
+# test data with reason but 404 in curl respond
+id=404
+testurl="https://openqa.opensuse.org/tests/${id}"
+try-client-output investigate_issue $id
+is "$rc" 0 'investigate_issue with missing autoinst-log and with reason in job_data'
+has "$got" "without autoinst-log.txt older than 14 days. Do not label" "exits succefully when is old job without autoinst-log.txt"
+
+# all requisite evidence are missing
```
kalikiana, line 64:
```suggestion
# all assets are missing
```

I assume that's what you meant?

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,114 @@
...
+id=101
+testurl="https://openqa.opensuse.org/tests/${id}"
+# `cur_date` is used on the following 4 test cases
+cur_date=$(date +%F)
+sed -i "s/yyyy-mm-dd/${cur_date}/" "$dir/data/${id}.json"
```
kalikiana, line 69: Please don't modify any files in git. You can create or copy the files if they need to be changed on the fly.

test/07-openqa-label-known-issues.t
kalikiana: https://github.com/os-autoinst/scripts/pull/342/files#r1753452664

openqa-label-known-issues
kalikiana: @b10n1k Did you see this?

openqa-label-known-issues
kalikiana: @b10n1k Is this case covered in unit tests?

openqa-label-known-issues
kalikiana: Agreed. @b10n1k Please just drop the comment.

test/07-openqa-label-known-issues.t
perlpunk: I think at this point I would go with it and fix it later, just to get this merged.

test/07-openqa-label-known-issues.t
b10n1k: i dont think it is accurate but it s fine. done

b10n1k: 9b5744d

dependencies.yaml
kalikiana: Let's leave it for a follow up ticket as we don't use this properly yet

.github/workflows/ci.yml
okurz: We discussed this in the unblock meeting. The line `sudo apt-get install cpanminus` is meant to *only* install cpanm because we call this in the next line. The line was never meant to provide proper test dependencies not handled by cpanm. Anyway, you guys overruled me so let's accept it as is :)

test/07-openqa-label-known-issues.t
```diff
@@ -0,0 +1,120 @@
...
+    local id=$1 comment=$2 force_result=${3:-''}
+    echo "$comment"
+}
+out=''
+hxnormalize() {
+    echo -n "$out";
+}
+hxselect() {
+    echo -n "$out";
+}
```
perlpunk, line 47: This fix works for me in github actions:
```suggestion
hxnormalize() {
    cat "$out"
}
hxselect() {
    cat "$out"
}
```

baierjan: Is that supposed to mock the `if hxnormalize -x "$out" | hxselect -s '\n' -c '` line?

If yes, the mocked `hxnormalize` could just output `$2` (unless you need some fancy features) and `hxselect` its input as it clearly works as a filter... although I am not exactly sure what is the intended goal.

```suggestion
hxnormalize() {
    echo -n "$2"
}
hxselect() {
    cat -
}
```

b10n1k: @baierjan u nail it. I just apply it but I would go back to read the comment to understand it a bit better. but it is very nice to see it working.

b10n1k: and I see how silly I was if I had `;` in all the attempts. thanks again

b10n1k: But it has been resolved.

openqa-label-known-issues
kalikiana: @b10n1k Do you see unresolved comments? 🙃

b10n1k: As we discusssed, this might be a thing but it is more that just change the name of the file. So I left it out of this PR and it is something to look at later

perlpunk:
This was never deployed on o3, due to https://progress.opensuse.org/issues/166721#note-15
This explains why we still saw the error in line 68: https://progress.opensuse.org/issues/166772

PR #343, opened 2024-08-28
[Suggestion] Handle unsuccessful label call

perlpunk:
If label_on_issues_from_issue_tracker returns non-zero, it doesn't usually mean a fatal error, but simply that it didn't find a matching ticket. In that case we want to go on with
* label_on_issues_without_tickets
* handle_unreviewed

For that I rearranged the if/elif a bit

Issue: https://progress.opensuse.org/issues/165716

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -41,14 +41,14 @@ handle_unreachable() {
         # the page might be gone, try the scheme+host we configured (might be different one though)
         if ! grep -q "$host_url" <<< "$testurl"; then
             echoerr "'$testurl' is not reachable and 'host_url' parameter does not match '$testurl', can not check further, continuing with next"
-            return
+            return 1
         fi
         if ! curl "${curl_args[@]}" -s --head "$host_url"; then
             echoerr "'$host_url' is not reachable, bailing out"
             curl "${curl_args[@]}" --head "$host_url"
         fi
         echoerr "'$testurl' is not reachable, assuming deleted, continuing with next"
-        return
+        return 1
     fi
     # resorting to downloading the job details page instead of the
     # log, overwrites $out
@@ -61,7 +61,7 @@ handle_unreachable() {
     if hxnormalize -x "$out" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
         "${client_call[@]}" -X POST jobs/"$id"/comments text='poo#62456 test incompletes after failing in GRU download task on "Inactivity timeout" with no logs'
         "${client_call[@]}" -X POST jobs/"$id"/restart
-        return
+        return 1
     fi
 }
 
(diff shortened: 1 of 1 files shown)


b10n1k:
I had to adjust and fix some closing ifs to run against the tests from https://github.com/os-autoinst/scripts/pull/342/files
Here is the output

```
❯ prove -v test/07-openqa-label-known-issues.t
test/07-openqa-label-known-issues.t .. 
1..9
/home/iob/Documents/Work/qatools/repos/scripts/.bpan/.bpan/lib/test-tap.bash: ERROR: line 168
ok 1 - id required
ok 2 - investigate_issue with missing autoinst-log and with reason in job_data
not ok 3 - investigation detects and reports old missing autoinst-log
#   Failed test 'investigation detects and reports old missing autoinst-log'
#   at test/07-openqa-label-known-issues.t line 54.
# Text 'does not have autoinst-log.txt but is rather old, ignoring' NOT found in: '
# Requesting jobs/404 via openqa-cli
# 200200openqa-label-known-issues: line 75: issues: unbound variable
# [https://openqa.opensuse.org/tests/404](https://openqa.opensuse.org/tests/404): Unknown test issue, to be reviewed
# -> [autoinst-log.txt](https://openqa.opensuse.org/tests/404/file/autoinst-log.txt)
# 
# Last lines before SUT shutdown:
# 
#     # --- 8< ---
# 
#     # --- >8 ---
# '
OUT  test 200 autoinst az network vnet peering create. something failed
ok 4 - investigate_issue with autoinst-log and without reason
ok 5 - investigation label job with matched autoinst-log context
ok 6 - job with emtpy autoinst-log checks unknown issue
ok 7 - check output
ok 8 - job label without tickets
ok 9 - investistigation label correctly job without ticket
Failed 1/9 subtests 

Test Summary Report
-------------------
test/07-openqa-label-known-issues.t (Wstat: 0 Tests: 9 Failed: 1)
  Failed test:  3
Files=1, Tests=9,  0 wallclock secs ( 0.02 usr  0.01 sys +  0.27 cusr  0.11 csys =  0.41 CPU)
Result: FAIL
```

Keep in mind that the test is still under review and there might be mistakes in it.

perlpunk:
My previous version tried to reflect the logic from your PR.

I now changed the logic to what I think might be what we want.

b10n1k:
> My previous version tried to reflect the logic from your PR.
> 
> I now changed the logic to what I think might be what we want.

I run the tests against this

```
❯ prove -v test/07-openqa-label-known-issues.t
test/07-openqa-label-known-issues.t .. 
1..11
/home/iob/Documents/Work/qatools/repos/scripts/.bpan/.bpan/lib/test-tap.bash: ERROR: line 168
ok 1 - id required
ok 2 - investigate_issue with missing autoinst-log and with reason in job_data
not ok 3 - investigation keep running without autoinst-log
#   Failed test 'investigation keep running without autoinst-log'
#   at test/07-openqa-label-known-issues.t line 60.
# Text 'Job incompletes with reason' NOT found in: '
# Requesting jobs/404 via openqa-cli
# 200200'https://openqa.opensuse.org/tests/404' does not have autoinst-log.txt but is rather old, ignoring
# '
ok 4 - investigate_issue with missing old autoinst-log and without reason in job_data
ok 5 - investigation detects and reports old missing autoinst-log
ok 6 - investigate_issue with autoinst-log and without reason
ok 7 - investigation label job with matched autoinst-log context
ok 8 - job with emtpy autoinst-log checks unknown issue
ok 9 - investigation still label Unknown issue
ok 10 - job label without tickets
ok 11 - investistigation label correctly job without ticket
Failed 1/11 subtests 

Test Summary Report
-------------------
test/07-openqa-label-known-issues.t (Wstat: 0 Tests: 11 Failed: 1)
  Failed test:  3
Files=1, Tests=11,  1 wallclock secs ( 0.02 usr  0.01 sys +  0.28 cusr  0.11 csys =  0.42 CPU)
Result: FAIL
```

Now the *investigation keep running without autoinst-log* gets into the block which checks if it is old. but there is reason. 
Note that tests do not cover 

```
if [[ -z $reason ]] || [[ $reason = null ]]; then
  echoerr "'$testurl' does not have autoinst-log.txt or reason, cannot label"
  return
fi
```

I like how your code looks like. I am a little troubled about AC2 as you mentioned https://github.com/os-autoinst/scripts/pull/342#discussion_r1740642391 because if I understand correctly the AC2 was not satisfied in the PR on that ticket, and instead was approved. Does it need further refactoring? I am still confuse with `[[ $curl_output != 404 ]] && return`. I assumed that the next if should run exactly when we have 404, not the other way around.

But if you feel confident, please assign the ticket to you. we both work in the same problem and you are faster

kalikiana:
> But if you feel confident, please assign the ticket to you. we both work in the same problem and you are faster

I think it's worth taking the time to learn it.

Maybe use this as a base and add tests to it.

perlpunk:
> > > But if you feel confident, please assign the ticket to you. we both work in the same problem and you are faster

If you really don't wanna work on this ticket anymore, we shouldn't force you. Sorry if this is frustrating. bash is not the easiest language when the logic get more complex.
But it's not the point who can solve a problem fastest. Then everyone would work on those tickets which they are most experienced with, and our bus factor would get worse.

I currently have two other tickets with higher prio, so I was looking forward to you writing the tests.
You made a good start; I think you hadn't worked with that bash test framework before, right?

> > I think it's worth taking the time to learn it.
> > Maybe use this as a base and add tests to it.
> 
> The two PR have slightly different approach and results. it is not anymore about learning.
> And taking this is not anymore my code which I believe it is unfair to commit as mine.

That shouldn't be a problem. You could just cherry-pick the commit. I think this is the logic we want, and I think with your approach of adding a variable it will look more complicated to get that same logic.

> In additional they are two weeks trying to unblock this issue.

Fortunately it is not that urgent.

perlpunk:
Here I wrote down the main logic that I have in mind:
https://progress.opensuse.org/issues/165716#note-19

PR #345, opened 2024-09-09
Skip posting investigation comment if original job has already been deleted

Martchus:
Besides the added unit tests, I tested this by commenting out `main()` and invoking the relevant function manually against my local test openQA instance.

Test with existing job:
```
$ export scheme=http host=127.0.0.1:9526
$ export job_data='{"job": { "result": "failed", "settings": { "OPENQA_INVESTIGATE_ORIGIN": "http://127.0.0.1:9526/t4266" } } }'
$ bash -c 'source ./openqa-investigate; post-investigate 12345 investigate:retry'
Job investigate:retry (12345) is not a retry of wireguard_client (4266)
```

Test with non-existing job:
```
$ export job_data='{"job": { "result": "failed", "settings": { "OPENQA_INVESTIGATE_ORIGIN": "http://127.0.0.1:9526/t5678" } } }'
$ bash -c 'source ./openqa-investigate; post-investigate 12345 investigate:retry'
openqa-cli (317 …/_common): Error making API request (jobs/5678): 404 Not Found
{"error_status":404}
Skip posting investigation comment on original job 5678 as it does not exist anymore
$ echo $?
0
```

Related ticket: https://progress.opensuse.org/issues/166403

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -283,7 +283,10 @@ post-investigate() {
     retry_result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
     investigate_origin="$(echo "$job_data" | runjq -r '.job.settings.OPENQA_INVESTIGATE_ORIGIN')" || return 1
     origin_job_id=${investigate_origin#"$host_url/t"}
-    origin_job_data=$(client-get-job "$origin_job_id") || return $?
+    origin_job_data=$(client-get-job "$origin_job_id") || rc=$?
+    # shellcheck disable=SC2154
+    [[ $origin_job_data =~ $not_found_regex ]] && echo "Skipping posting investigation comment on original job $origin_job_id as it does not exist anymore" && return 0
+    [[ $rc != 0 ]] && return $rc
     origin_name="$(echo "$origin_job_data" | runjq -r '.job.test')" || return $?
     # cluster jobs might have the same OPENQA_INVESTIGATE_ORIGIN as the root retry job
     [[ $retry_name != "$origin_name:investigate:retry" ]] && echo "Job $retry_name ($id) is not a retry of $origin_name ($origin_job_id)" && return 0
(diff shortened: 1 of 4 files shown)


openqa-investigate
```diff
@@ -283,7 +283,10 @@ post-investigate() {
     retry_result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
     investigate_origin="$(echo "$job_data" | runjq -r '.job.settings.OPENQA_INVESTIGATE_ORIGIN')" || return 1
     origin_job_id=${investigate_origin#"$host_url/t"}
-    origin_job_data=$(client-get-job "$origin_job_id") || return $?
+    origin_job_data=$(client-get-job "$origin_job_id") || rc=$?
+    # shellcheck disable=SC2154
+    [[ $origin_job_data =~ $not_found_regex ]] && warn "Skip posting investigation comment on original job $origin_job_id as it does not exist anymore" && return 0
```
okurz, line 288:
```suggestion
    [[ $origin_job_data =~ $not_found_regex ]] && warn "Skipping posting investigation comment on original job $origin_job_id as it does not exist anymore" && return 0
```

but why warn and not just "echo"?

okurz (changes requested):

kalikiana (approved):

openqa-investigate
Martchus: To print it to stderr I suppose. I could also use echoerr if we have such a helper. However, I don't think considering this a warning is wrong because normally the original job is still supposed to be there. (And a warning is still not an error so it is in-line with returning 0.)

kalikiana: There is nothing actionable, though? To my mind a warning means there's something to be sorted out. I don't see what one would be able to do here?

kalikiana: Put another way if we wanted to ensure the original job was kept around we would have to add a comment to it before starting the investigation or some other means of keeping it around. Instead of a warning.

okurz (approved):

PR #346, opened 2024-09-12
Add systemd timer and service for updates

kalikiana:
See: https://progress.opensuse.org/issues/166721

### DIFF
--- a/systemd/os-autoinst-scripts-update-git.service
+++ b/systemd/os-autoinst-scripts-update-git.service
@@ -1,0 +1,8 @@
+[Unit]
+Description=Update git repository of os-autoinst-scripts
+After=network-online.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/git -C /opt/os-autoinst-scripts pull --quiet --rebase origin
+User=geekotest
--- a/systemd/os-autoinst-scripts-update-git.timer
+++ b/systemd/os-autoinst-scripts-update-git.timer
@@ -1,0 +1,9 @@
+[Unit]
+Description=Update git repository of os-autoinst-scripts
+
+[Timer]
+OnCalendar=*:0/3
+Unit=os-autoinst-scripts-update-git.service
+
+[Install]
+WantedBy=timers.target


systemd/os-autoinst-scripts-update-git.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update git repository of os-autoinst-scripts
+
+[Timer]
+OnCalendar=*:0/3
+Unit=os-autoinst-scripts-update-git.service
```
baierjan, line 6: That's in the `.timer`, you are missing `[Service]` section.

baierjan (changes requested):

systemd/os-autoinst-scripts-update-git.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update git repository of os-autoinst-scripts
+
+[Timer]
+OnCalendar=*:0/3
+Unit=os-autoinst-scripts-update-git.service
+
+[Install]
+WantedBy=default.target
```
baierjan, line 9: I think the default target should be
```suggestion
[Install]
WantedBy=timers.target
```

kalikiana: I've seen both 🤔 Not sure what the difference is in practice. I don't see an explicit mention of that in [systemd timer upstream docs](https://man.archlinux.org/man/systemd.timer.5#OPTIONS) or systemd.unit.

baierjan: https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#timers.target

systemd/os-autoinst-scripts-update-git.service
```diff
@@ -0,0 +1,11 @@
...
+Description=Update git repository of os-autoinst-scripts
+After=network-online.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/git -C /opt/os-autoinst-scripts pull --quiet --rebase origin
+User=geekotest
+
+[Install]
+WantedBy=default.target
```
baierjan, line 11: You might want to drop the install section as this service is supposed to be started only with the timer and not actually enabled?

Martchus (approved):

baierjan (approved):

PR #347, opened 2024-09-15
Prevent override the autoinst-log context within handle_unreachable

b10n1k:
`handle_unreachable` can override the log file passed to the function. This should not allowed because when the `handle_unreviewed` function is reach will run with incorrect logs. Solving the issue creating a separate temp local file
 for handle_unreachable and extract timeago in a variable in order to use it to
 determine the age of the job. This keeps clear now the out variable which
 should only contain the autoinst-log+reason.

https://progress.opensuse.org/issues/166772

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -130,7 +130,8 @@ label_on_issues_without_tickets() {
 }
 
 investigate_issue() {
-    local id="${1##*/}"
+    local testurl=$1
+    local id="${testurl##*/}"
     local reason
     local curl_output
     echo "Requesting jobs/${id} via openqa-cli"
@@ -174,7 +175,7 @@ investigate_issue() {
 }
 
 label_issue() {
-    testurl="${1:?"Need 'testurl'"}"
+    local testurl="${1:?"Need 'testurl'"}"
     [[ "$dry_run" = "1" ]] && client_prefix="echo"
     if [[ -z "$client_call" ]]; then
         client_call=(openqa-cli "${client_args[@]}")
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -28,15 +28,16 @@ MOJO_CONNECT_TIMEOUT=${MOJO_CONNECT_TIMEOUT:-30}
 client_args=(api --header 'User-Agent: openqa-label-known-issues (https://github.com/os-autoinst/scripts)' --host "$host_url" --retries="$retries")
 
 out="${REPORT_FILE:-$(mktemp -t openqa-label-known-issues--output-XXXX)}"
+declare timeago
 trap 'error-handler "$LINENO"' ERR
 trap 'test "$KEEP_REPORT_FILE" == "1" || rm "$out"' EXIT
+trap 'test "$KEEP_JOB_HTML_FILE" == "1" || rm "$html_out"' EXIT
 
 echoerr() { echo "$@" >&2; }
 
 handle_unreachable() {
     local testurl=$1
-    local out=$2
-
+    html_out=${JOB_HTML_FILE:-$(mktemp -t openqa-label-known-issues--job-details-XXXX)}
     if ! curl "${curl_args[@]}" -s --head "$testurl" -o /dev/null; then
         # the page might be gone, try the scheme+host we configured (might be different one though)
         if ! grep -q "$host_url" <<< "$testurl"; then
@@ -51,17 +52,18 @@ handle_unreachable() {
         return 1
     fi
     # resorting to downloading the job details page instead of the
-    # log, overwrites $out
-    if ! curl "${curl_args[@]}" -s "$testurl" -o "$out"; then
+    # log
+    if ! curl "${curl_args[@]}" -s "$testurl" -o "${html_out}"; then
         echoerr "'$testurl' can be reached but not downloaded, bailing out"
         curl "${curl_args[@]}" "$testurl"
         exit 2
     fi
+    timeago=$(grep timeago "$html_out" | hxselect -s '\n' -c '.timeago::attr(title)')
 
-    if hxnormalize -x "$out" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
+    if hxnormalize -x "${html_out}" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
         "${client_call[@]}" -X POST jobs/"$id"/comments text='poo#62456 test incompletes after failing in GRU download task on "Inactivity timeout" with no logs'
         "${client_call[@]}" -X POST jobs/"$id"/restart
-        return 1
+        return
     fi
 }
 
@@ -137,9 +139,10 @@ investigate_issue() {
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
         handle_unreachable "$testurl" "$out" || return
+
         [[ $curl_output != 404 ]] && return
-        # Checking timestamp
-        if [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
+        # Checking timestamp given by job details page
+        if [[ $(date -uIs -d '-14days') > $timeago ]]; then
             # if the page is there but not even an autoinst-log.txt exists
             # then the job might be too old and logs are already deleted.
             echoerr "'$testurl' job#${id} without autoinst-log.txt older than 14 days. Do not label"
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -28,15 +28,19 @@ MOJO_CONNECT_TIMEOUT=${MOJO_CONNECT_TIMEOUT:-30}
 client_args=(api --header 'User-Agent: openqa-label-known-issues (https://github.com/os-autoinst/scripts)' --host "$host_url" --retries="$retries")
 
 out="${REPORT_FILE:-$(mktemp -t openqa-label-known-issues--output-XXXX)}"
-declare timeago
 trap 'error-handler "$LINENO"' ERR
-trap 'test "$KEEP_REPORT_FILE" == "1" || rm "$out"' EXIT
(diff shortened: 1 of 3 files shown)


openqa-label-known-issues
```diff
@@ -35,7 +35,8 @@ echoerr() { echo "$@" >&2; }
 
 handle_unreachable() {
     local testurl=$1
-    local out=$2
+    local out
+    out=$(cat "$2")
```
perlpunk, line 39: The content of the variable `out` is supposed to be a filename.
Now it contains the content of the autoinst-log.txt.
That will probably fail if it contains a slash or anything that cannot be part of a filename, and it's a mess anyway.
Also the file should be created via `mktemp`, as we do for the other file.

still, that's not enough, because we are calling
```
f [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
```
later in line 144 (that's the part you extracted from `handle_unreachable` in one of the previous PRs), and that expects the HTML content in the `$out` file, not the autoinst log.
So the filename / variable has to be available at that point.
There should be two files, and two distinct variable names.

openqa-label-known-issues-and-investigate-hook
```diff
@@ -44,6 +44,7 @@ hook() {
     result="$(echo "$job_data" | runjq -r '.job.result')" || return $?
     [[ "$state" != "done" ]] && return
     if [[ "$result" != passed ]]; then
+        echo "running $0 against $url"
```
perlpunk, line 47: Please remove that again

openqa-label-known-issues
```diff
@@ -51,14 +52,15 @@ handle_unreachable() {
...
+    # log
+    if ! curl "${curl_args[@]}" -s "$testurl" -o "${html_out}"; then
         echoerr "'$testurl' can be reached but not downloaded, bailing out"
         curl "${curl_args[@]}" "$testurl"
         exit 2
     fi
 
-    if hxnormalize -x "$out" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
+    if hxnormalize -x "${html_out}" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
+        echo "POST jobs/${id}/comments"
```
perlpunk, line 63: Please remove that again

openqa-label-known-issues
```diff
@@ -35,7 +35,8 @@ echoerr() { echo "$@" >&2; }
 
 handle_unreachable() {
     local testurl=$1
-    local out=$2
+    local html_out
+    html_out=$(mktemp -t openqa-label-known-issues--job-details-XXXX)
```
perlpunk, line 39: Ok. now this file needs to be made available for the code in line 144 as well:
```
if [[ $(date -uIs -d '-14days') > $(grep timeago "$out" | hxselect -s '\n' -c '.timeago::attr(title)') ]]; then
```

openqa-label-known-issues
```diff
@@ -51,18 +52,19 @@ handle_unreachable() {
...
         exit 2
     fi
 
-    if hxnormalize -x "$out" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
+    if hxnormalize -x "${html_out}" | hxselect -s '\n' -c '.links_a .resborder' | grep -qPzo '(?s)Gru job failed.*connection error.*Inactivity timeout'; then
         "${client_call[@]}" -X POST jobs/"$id"/comments text='poo#62456 test incompletes after failing in GRU download task on "Inactivity timeout" with no logs'
         "${client_call[@]}" -X POST jobs/"$id"/restart
         return 1
     fi
+    timeago=$(grep timeago "$html_out" | hxselect -s '\n' -c '.timeago::attr(title)')
```
perlpunk, line 67: global, undeclared variables should be avoided.
Would it make sense to just move this call back into `handle_unreachable`?
What was the reason to extract it in the first place?

openqa-label-known-issues
```diff
@@ -51,18 +51,26 @@ handle_unreachable() {
...
         "${client_call[@]}" -X POST jobs/"$id"/comments text='poo#62456 test incompletes after failing in GRU download task on "Inactivity timeout" with no logs'
         "${client_call[@]}" -X POST jobs/"$id"/restart
         return 1
     fi
+    # Checking timestamp given by job details page
+    if [[ $(date -uIs -d '-14days') > $timeago ]]; then
+        # if the page is there but not even an autoinst-log.txt exists
+        # then the job might be too old and logs are already deleted.
+        echoerr "'$testurl' job#${id} without autoinst-log.txt older than 14 days. Do not label"
+        return
```
perlpunk, line 72:
```suggestion
        return 1
```
Like I said in the meeting, you have to return with non zero here now of course, so that the later check when calling handle_unreachable will return

perlpunk (commented):
hxselect is overwritten in the test with just `cat -`, so it won't return the date. That's why the test fails

test/07-openqa-label-known-issues.t
```diff
@@ -42,8 +43,18 @@ out=''
...
+    if [ -p /dev/stdin ]; then
+	myinput="$(cat)"
+	if echo "$myinput" | grep -q "incomplete"; then
+	    echo "$myinput" | awk -F'title="' '{split($2, a, "\""); print a[1]}'
+	else
+	    cat -
+	fi
+    else
+	cat -
+    fi
```
Martchus, line 57: We are using spaces in this file for indentation. You could also avoid the duplicate branches:

```suggestion
    if [[ -p /dev/stdin ]]; then
        myinput=$(cat)
        if echo "$myinput" | grep -q "incomplete"; then
            echo "$myinput" | awk -F'title="' '{split($2, a, "\""); print a[1]}'
            return 0
        fi
    cat -
```

Martchus (changes requested):

openqa-label-known-issues
b10n1k: @perlpunk as I didnt follow the exact logic, this request havent applied. Can we close and open once again if you find it neccessary?

openqa-label-known-issues
b10n1k: I am not if this better, but at least I declared the variable. wdyt?

openqa-label-known-issues
```diff
@@ -28,15 +28,16 @@ MOJO_CONNECT_TIMEOUT=${MOJO_CONNECT_TIMEOUT:-30}
 client_args=(api --header 'User-Agent: openqa-label-known-issues (https://github.com/os-autoinst/scripts)' --host "$host_url" --retries="$retries")
 
 out="${REPORT_FILE:-$(mktemp -t openqa-label-known-issues--output-XXXX)}"
+declare timeago
 trap 'error-handler "$LINENO"' ERR
 trap 'test "$KEEP_REPORT_FILE" == "1" || rm "$out"' EXIT
+trap 'test "$KEEP_JOB_HTML_FILE" == "1" || rm "$html_out"' EXIT
```
perlpunk, line 34: If you have multiple handlers for the same signal, the last one will be winning and just overwrite the previous ones.

perlpunk:
I pushed a commit.
Can you now please check the tests?

test/07-openqa-label-known-issues.t
```diff
@@ -42,8 +43,16 @@ out=''
 hxnormalize() {
     cat "$2"
 }
+
 hxselect() {
-    cat -
+    if [ -p /dev/stdin ]; then
```
perlpunk, line 48: Please check if it is still necessary at all to overwrite hxnormalize and hxselect here.

perlpunk (dismissed):

openqa-label-known-issues
perlpunk: I can tell you from > 25 years of experience that it might seem convenient to just make a variable global, and yes in bash that's even more often the case, but it creates technical debt.
It's hard to avoid for the output files that we have to remove in the exit handler, but don't just do it because it seems convenient.
In my commit I moved that statement back into `_handle_unreachable`.

openqa-label-known-issues
```diff
@@ -136,15 +150,9 @@ investigate_issue() {
     echo "$reason" >> "$out"
     if [[ "$curl_output" != "200" ]] && [[ "$curl_output" != "301" ]]; then
         # if we can not even access the page it is something more critical
-        handle_unreachable "$testurl" "$out" || return
+        handle_unreachable "$testurl" "$out" || return 0
```
perlpunk, line 153: I noticed during my debugging that we have to explicitly return 0 here. Otherwise it would return the 1 from inside `handle_unreachable`.

test/07-openqa-label-known-issues.t
b10n1k: they seems to work. What have changed? I think the reason I did this was that the tools were not available before

openqa-label-known-issues
b10n1k: wonder if there should be a test for this

perlpunk: If you remove the 0, the tests will fail

test/07-openqa-label-known-issues.t
perlpunk: what for? tests are passing for me locally without those functions

test/07-openqa-label-known-issues.t
```diff
@@ -29,9 +30,9 @@ openqa-cli() {
 }
 curl() {
     if [[ "$7" =~ /(404|414|101|102)/file/autoinst-log.txt ]]; then
-        echo -n "404"
+	echo -n "404"
     else
-        echo -n "200"
+	echo -n "200"
```
perlpunk, line 35: no tabs please

test/07-openqa-label-known-issues.t
```diff
@@ -42,8 +43,16 @@ out=''
 hxnormalize() {
     cat "$2"
 }
+
 hxselect() {
-    cat -
+    if [ -p /dev/stdin ]; then
+	in="$(cat)"
```
perlpunk, line 49: no tabs please

Martchus (approved):

test/07-openqa-label-known-issues.t
perlpunk: html-xml-utils is now installed in CI.
Otherwise a comment in the test would have been useful to tell people who don't know it and forgetful people like me like "These commands are not available in our CI, mocking"

PR #348, opened 2024-10-11
Use a single API query in `openqa-advanced-retrigger-jobs`

Martchus:
Tested with a dry run against o3 and a regular run against my local
instance overriding `JOB_IDS`.

Related ticket: https://progress.opensuse.org/issues/168013


Martchus:
I've just pushed an additional commit to avoid timeouts because I think if we get many job IDs it is actually not a good idea to handle all of them in a single query. I guess this is now a good compromise and it'll still be much faster than doing one query per job.

perlpunk:
> I've just pushed an additional commit to avoid timeouts because I think if we get many job IDs it is actually not a good idea to handle all of them in a single query.

I guess that same strategy should be used when using the new comment+restart feature from the webui.

okurz (approved):

perlpunk (approved):

PR #351, opened 2024-11-06
Add script for building ipxe

asdil12:
Ticket: https://progress.opensuse.org/issues/155524

### DIFF
--- a/ipxe/build.sh
+++ b/ipxe/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -e
+
+sudo zypper -n install cross-aarch64-gcc11 xz-devel
+
+test -d ipxe || git clone https://github.com/ipxe/ipxe.git
+pushd ipxe/src
+
+# this would compile the ipxe version using its internal driver implementations
+#make -j3 EMBED=../../boot.ipxe bin/ipxe.pxe bin-x86_64-efi/ipxe.efi
+#make -j3 EMBED=../../boot.ipxe CROSS=aarch64-suse-linux- bin-arm64-efi/ipxe.efi
+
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

Martchus (commented):
It is considered a bad practice to conduct a build as root but as part of this script it would have to be conducted as root (otherwise the rest of the script doesn't work). Maybe it makes more sense to use sodo within the script as needed.

ipxe/build.sh
```diff
@@ -0,0 +1,19 @@
...
+
+test -d ipxe || git clone https://github.com/ipxe/ipxe.git
+pushd ipxe/src
+
+# this would compile the ipxe version using its internal driver implementations
+#make -j3 EMBED=../../boot.ipxe bin/ipxe.pxe bin-x86_64-efi/ipxe.efi
+#make -j3 EMBED=../../boot.ipxe CROSS=aarch64-suse-linux- bin-arm64-efi/ipxe.efi
+
+# undionly / snponly uses the network driver of the PXE / UEFI stack
+make -j3 EMBED=../../boot.ipxe bin/undionly.kpxe bin-x86_64-efi/snponly.efi
```
Martchus, line 13: Is the build system broken or why are you limiting this to 3 cores. If it is broken it would make more sense to limit it to 1 core. Otherwise I'd leave it out and not override MAKEFLAGS.

Martchus (approved):
Despite the issue mentioned in the in-line comment good enough to merge.

ipxe/build.sh
asdil12: netboot only has 2 cores

PR #352, opened 2024-11-15
Avoid submitting packages to Factory that we haven't actually tested yet

Martchus:
The packages on `devel:openQA` might change while openQA-in-openQA tests are executing. So far we submit packages directly from `devel:openQA` after openQA-in-openQA have passed. This means we don't neccassarily submit the version of the packages we have actually tested.

With this change we "release" a snapshot of `devel:openQA` to the staging repository `devel:openQA:testing` when triggering tests and use that within openQA-in-openQA tests (see
https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/215). We then only submit from this staging repository when tests have passed. This way we only submit what we've tested despite changes on `devel:openQA`.

Packages in the staging repository are cleaned up when openQA-in-openQA tests fail or after packages have been submitted. This trigger script will only release a new version into the staging project if it has been cleaned up ensuring its contents stay the same while tests/submissions are pending.

This commit also adds the new script `cleanup-obs-project`. It conveniently deletes all packages in an OBS project without deleting the project itself. We might want to add it on Jenkins as post-failure action for monitoring and submissions to avoid leftovers. (It is also useful to test the deletion code on its own.)

---

I tested this locally by running the individual added/changed parts of the scripts. I tested whether the osc commands to what they are supposed to against the real OBS instance where the project https://build.opensuse.org/project/show/devel:openQA:testing already exists.

I haven't adjusted any pipelines on Jenkins yet. I suppose this is also not necessary. It might be nice invoke `cleanup-obs-project` as post-failure action on http://jenkins.qa.suse.de/job/monitor-openQA_in_openQA-TW/configure and http://jenkins.qa.suse.de/job/submit-openQA-TW-to-oS_Fctry/configure. One cannot configure post-failure actions explicitly but I suppose we can just append an ` || …` in the shell script.

### DIFF
--- a/monitor-openqa_job
+++ b/monitor-openqa_job
@@ -16,6 +16,7 @@ host="${host:-"https://openqa.opensuse.org"}"
 openqa_groupid="${openqa_groupid:-"24"}"
 obs_component="${obs_component:-"package"}"
 obs_package_name="${1:-""}"
+staging_project=${staging_project:-devel:openQA:testing}
 comment_on_obs=${comment_on_obs:-}
 
 # shellcheck source=/dev/null
@@ -41,6 +42,11 @@ for job_id in $(job_ids job_post_response); do
 done
 
 [[ ${failed_jobs[*]} ]] || exit 0
+
+# delete packages from staging project in error case as we will not continue submitting those packages
+delete_packages_from_obs_project "$staging_project"
+
+# comment on failed jobs
 [[ $comment_on_obs ]] || exit 1
 
 osc api "/comments/$obs_component/$obs_package_name" | grep id= | sed -n 's/.*id="\([^"]*\)">.*test.* failed.*/\1/p' | while read -r id; do
--- a/trigger-openqa_in_openqa
+++ b/trigger-openqa_in_openqa
@@ -22,12 +22,20 @@ build_tag=${BUILD_TAG:-}
 client_prefix=${client_prefix:-}
 full_run=${FULL:-}
 group_id="${group_id:-"openQA"}"
+osc=${osc:-osc}
+src_project=${src_project:-devel:openQA}
+staging_project=${staging_project:-${src_project}:testing}
+dst_project=${dst_project:-${src_project}:tested}
+
+# shellcheck source=/dev/null
+. "$(dirname "$0")"/_common
 
 main() {
-    [ "$dry_run" = "1" ] && client_prefix="echo"
+    [[ $dry_run == 1 ]] && client_prefix="echo" osc="echo $osc"
     local qcow build
     download_scenario
     download_latest_published_tumbleweed_image
+    create_devel_openqa_snapshot
     trigger
 }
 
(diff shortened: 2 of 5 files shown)


trigger-openqa_in_openqa
```diff
@@ -22,12 +22,20 @@ build_tag=${BUILD_TAG:-}
 client_prefix=${client_prefix:-}
 full_run=${FULL:-}
 group_id="${group_id:-"openQA"}"
+osc=${osc:-osc}
+src_project=${dst_project:-devel:openQA}
+staging_project=${dst_project:-devel:openQA:testing}
+dst_project=${dst_project:-devel:openQA:tested}
```
baierjan, line 28: Is there a use-case to allow overriding all the project to the same value and have `$src_project` and `$dst_project` the same or is it a typo?

baierjan (commented):
So if I get it correctly, we have one staging place (testing) which will be used by the oQA-in-oQA and if the tests pass, the packages will be "released" to the tested for the automated submission into Factory, otherwise it will be cleanup upon failure to make space for next (I guess Jenkins) run to copy the code from the devel repo to testing for the actual testing. If we manage to make multiple changes and there is still something in the staging, it will be hold for a moment and batched together in the next opportunity, right? And the current release process does not need any change, because we only changed the way how to get the packages into the "tested" space...

I do have one question inline, otherwise it looks good to me.

trigger-openqa_in_openqa
Martchus: It is a typo :-)

Martchus:
> So if I get it correctly, we have one staging place (testing) which will be used by the oQA-in-oQA and if the tests pass, the packages will be "released" to the tested for the automated submission into Factory,

Not exactly. `devel:openQA` will be "release" to `devel:openQA:testing`. Only then we run tests. If tests pass, `devel:openQA:testing` will be copied to `devel:openQA:tested`. There we disable reliance on source services. Then we submit that to `openSUSE:Factory`.

I am quoting "release" here because this is how the osc command is called. What we are actually doing here is creating a snapshot for the sake of testing and creating a submit request if tests pass.

> otherwise it will be cleanup upon failure to make space for next (I guess Jenkins) run

Yes, we're talking about hourly scheduled pipeline runs on Jenkins here.

> If we manage to make multiple changes and there is still something in the staging, it will be hold for a moment and batched together in the next opportunity, right?

That will effectively happen. However, nothing is ever explicitly batched up. If there are still pending tests (or the submission itself is still pending) when Jenkins runs the next trigger job this trigger job will simply return early and do nothing. One hour later there will hopefully no pending tests/submissions anymore so we can trigger tests again (with all the changes that have been made meanwhile).

> And the current release process does not need any change, because we only changed the way how to get the packages into the "tested" space...

Yes, there's no additional step involved (that needed to be configured in Jenkins).

monitor-openqa_job
```diff
@@ -16,6 +16,7 @@ host="${host:-"https://openqa.opensuse.org"}"
 openqa_groupid="${openqa_groupid:-"24"}"
 obs_component="${obs_component:-"package"}"
 obs_package_name="${1:-""}"
+staging_project=${dst_project:-devel:openQA:testing}
```
baierjan, line 19: Ok, now I am not sure if this is also a typo 😄

baierjan:
Thank you for the explanation. Yes, we are on the same page here.

trigger-openqa_in_openqa
```diff
@@ -22,12 +22,20 @@ build_tag=${BUILD_TAG:-}
 client_prefix=${client_prefix:-}
 full_run=${FULL:-}
 group_id="${group_id:-"openQA"}"
+osc=${osc:-osc}
+src_project=${src_project:-devel:openQA}
+staging_project=${staging_project:-devel:openQA:testing}
+dst_project=${dst_project:-devel:openQA:tested}
```
baierjan, line 28: Last nitpick because I like how you solved it in [os-autoinst-obs-auto-submit](https://github.com/os-autoinst/scripts/pull/352/files#diff-e9e4a2f7fd0feb7816f56d2fcd9ad02870fc9344ea52dfc275699c1a7d2cbff8) and I think it should behave in the same way, when I only override `$src_project`.

```suggestion
staging_project=${staging_project:-${src_project}:testing}
dst_project=${dst_project:-${src_project}:tested}
```

Martchus:
I suppose it would be best if we merge this only after hackweek. There might be further changes required we'll only discover when it actually runs on Jenkins.

perlpunk (approved):

PR #357, opened 2024-11-27
Ensure staging project is also cleaned up in case of an early return

Martchus:


Martchus:
Looks like this worked including the cleanup. Now a new set of jobs is pending and being monitored. Let's see whether this will create a new submission (or at least cleanup the staging project).

perlpunk (approved):

kalikiana (approved):

PR #360, opened 2024-12-05
Fix sporadic error about broken pipe when checking recent SRs

Martchus:
This code sometimes runs into:

```
Exception ignored in: <osc.util.safewriter.SafeWriter object at 0x7fbe42c17860>
BrokenPipeError: [Errno 32] Broken pipe
Retrying up to 3 more times after sleeping 3s …
```

This is easily reproducible by adding the condition into a separate script
and running it a few times. The additional `cat > /dev/null` which this
change introduces seems to fix the issue.

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -183,7 +183,7 @@ fi
 
 # throttle number of submissions to allow only one SR within a certain number of days
 if [[ $throttle_days != 0 ]] && $osc request list --project "$dst_project" --type submit --state new,review --mine --days "$throttle_days" | grep --quiet 'Created by'; then
-    echo "Skipping submission, there is still a pending SR younger than $throttle_days."
+    echo "Skipping submission, there is still a pending SR younger than $throttle_days days."
     exit 0
 fi
 
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -182,13 +182,8 @@ if [[ -e job_post_skip_submission ]]; then
 fi
 
 # throttle number of submissions to allow only one SR within a certain number of days
-if [[ $throttle_days != 0 ]] && $osc request list --project "$dst_project" --type submit --state new,review --mine --days "$throttle_days" \
-    | {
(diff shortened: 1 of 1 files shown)


Martchus:
Note that the underlying problem is probably similar to the problem fixed by https://github.com/os-autoinst/scripts/commit/d7c19b1b4fdda4cde09dd6b1a24922cbfb2e6e01.

Martchus:
I tested this on the Jenkins VM and it seems to work:

```
martchus@jenkins:/opt/openqa-scripts> sudo -u jenkins ./os-autoinst-obs-auto-submit 
Skipping submission, there is still a pending SR younger than 2 days.
```

(But I also tested the negative case by putting something else in the grep filter.)

os-autoinst-obs-auto-submit
```diff
@@ -182,8 +182,14 @@ if [[ -e job_post_skip_submission ]]; then
...
 # throttle number of submissions to allow only one SR within a certain number of days
-if [[ $throttle_days != 0 ]] && $osc request list --project "$dst_project" --type submit --state new,review --mine --days "$throttle_days" | grep --quiet 'Created by'; then
-    echo "Skipping submission, there is still a pending SR younger than $throttle_days."
+if [[ $throttle_days != 0 ]] && $osc request list --project "$dst_project" --type submit --state new,review --mine --days "$throttle_days" \
+    | {
+        grep --quiet 'Created by'
+        rc=$?
+        cat > /dev/null
+        exit $rc
+    }; then
```
nicksinger, line 191: I only understand why and how this works because you explained me in Slack. Might be worth a comment or something. `#black magic` might be fitting :laughing:

nicksinger:
do you want to squash away https://github.com/os-autoinst/scripts/pull/360/commits/97d2600c23169cbeba798854e8b55449ba7e2da8 or keep it as history/alternative solution?

nicksinger (approved):

Martchus:
I want to keep it as history/alternative solution :-)

baierjan (approved):

PR #363, opened 2025-01-09
Treat obsoleted and skipped results as passed

kalikiana:
This can be configured via allowed_results.

See: https://progress.opensuse.org/issues/174583


Martchus (changes requested):
This only adjusts from what jobs we get the HDD image from. I think we actually need to adjust the scheduling/monitoring below.

openqa-schedule-mm-ping-test
```diff
@@ -53,7 +54,8 @@ job_templates:
       PARALLEL_WITH: ping_server
 EOF
 
-hdd=$(runcli openqa-cli api --host "$openqa_url" jobs version="$version" scope=relevant arch="$arch" flavor="$flavor" test="$test_name" latest=1 | runjq -r '.jobs | map(select(.result == "passed")) | max_by(.settings.BUILD) .settings.HDD_1')
```
baierjan, line 56: This is just HDD selection, we cannot use HDD_1 from obsoleted/skipped hosts, right?

openqa-schedule-mm-ping-test
```diff
@@ -53,7 +54,8 @@ job_templates:
...
+jobs=$(runcli openqa-cli api --host "$openqa_url" jobs version="$version" scope=relevant arch="$arch" flavor="$flavor" test="$test_name" latest=1 | runjq -r .jobs)
+hdd=$(echo $jobs | runjq -r "map(select(.result | test(\"(${allowed_results})\"))) | max_by(.settings.BUILD) .settings.HDD_1")
 time openqa-cli schedule \
```
baierjan, line 59: I guess this call is the problematic one.

okurz (changes requested):
That's not right. We can't just assume from obsolete jobs that the cloned jobs are passed. That's why I suggested to follow the cloned jobs but not within our downstream scripting but upstream API routes

openqa-schedule-mm-ping-test
kalikiana: Right. I'm looking into extending the monitor command now: https://github.com/os-autoinst/openQA/pull/6101

perlpunk:
> That's not right. We can't just assume from obsolete jobs that the cloned jobs are passed. That's why I suggested to follow the cloned jobs but not within our downstream scripting but upstream API routes

We discussed in the estimation that there were no automatic restarts, so there is nothing we can "follow".
What we could do - if we get obsoleted jobs, we could retry the whole schedule call. For that we would probably need to `tee` the JSON output.

okurz:
@perlpunk isn't that a symptom of a more significant problem? Our scripts-ci shouldn't be that sophisticated but rather openQA should behave in a more usable manner

Martchus:
Not sure whether obsolescence is tracked in the database so we could easily "follow" the new set of jobs in the monitoring command the openqa-cli. Even if it was tracked it would probably be not super easy to implement.

PR #365, opened 2025-01-16
Add check for build state of packages to be tested

baierjan:
Reference: https://progress.opensuse.org/issues/174451

### DIFF
--- a/trigger-openqa_in_openqa
+++ b/trigger-openqa_in_openqa
@@ -68,8 +68,15 @@ create_devel_openqa_snapshot() {
     elif [[ -e job_post_skip_submission ]]; then
         rm job_post_skip_submission
     fi
-    echo "Creating snapshots under $staging_project"
+    echo "Checking if all relevant packages are built and published before creating snapshot under $staging_project"
     for package in $auto_submit_packages; do
+        local state
+        state=$($osc results -r openSUSE_Tumbleweed -a x86_64 --no-multibuild "$src_project" "$package")
+        if grep -q -v 'succeeded\|disabled' <<< "$state"; then
+            echo -e "Package $package is not yet ready for release\n$state" | tee job_post_skip_submission
+            delete_packages_from_obs_project "$staging_project"
+            return
+        fi
         echo "Creating snapshots of $package"
         $osc release --no-delay --target-project "$staging_project" --target-repository=openSUSE_Tumbleweed -r openSUSE_Tumbleweed -a x86_64 "$src_project" "$package"
     done


trigger-openqa_in_openqa
```diff
@@ -68,6 +68,15 @@ create_devel_openqa_snapshot() {
     elif [[ -e job_post_skip_submission ]]; then
         rm job_post_skip_submission
     fi
+    echo "Checking if all relevant packages are build and published"
```
Martchus, line 71:
```suggestion
    echo "Checking if all relevant packages are built and published"
```

trigger-openqa_in_openqa
```diff
@@ -68,6 +68,15 @@ create_devel_openqa_snapshot() {
...
+    for package in $auto_submit_packages; do
+        local state
+        state=$($osc results -r openSUSE_Tumbleweed -a x86_64 --no-multibuild "$src_project" "$package")
+        if grep -q -v 'succeeded\|disabled' <<< "$state"; then
+            echo -e "Package $package is not yet ready for release\n$state" | tee job_post_skip_submission
+            return
+        fi
+    done
```
Martchus, line 79: This code needs to run *after* the snapshot creation and on the snapshot. Otherwise it still isn't guaranteed that packages are built and published when the snapshot is created. Of course this means the code needs to delete the problematic snapshot in the error case.

Martchus (changes requested):

trigger-openqa_in_openqa
baierjan: Yes, that's exactly the problem I described in the linked progress ticket. This was intended as a quick hack without changing much. I plan to address this issue in a subsequent work.

baierjan: As I though about it a little, I still think I really want to make that check *before* telling osc to make the copy as I want to know the state of the package I am about to copy over (remember, we are not doing any rebuild in the target project). I reordered the code a little to minimize the window between checking the state and calling `osc release` which should help. As we have only one source package for openQA (and one for os-autoinst) it should be just a quick look and copy over for each of them.

Martchus (approved):
It would probably still be better to check this after the snapshot was created. Of course we then needed to discard the snapshot in the (rare) error case but it would be race-free.

Wabri (approved):

baierjan:
> It would probably still be better to check this after the snapshot was created. Of course we then needed to discard the snapshot in the (rare) error case but it would be race-free.

Well, yes I also thought so, but what exactly do you want to check? We do not rebuild in the new project, so checking build status does not make much sense.

PR #366, opened 2025-01-23
Prevent test jobs for MM setup from being obsoleted

Martchus:
Those test jobs are not scheduled with `_OBSOLETE=1` themselves but so far they use a very common combination of DISTRI, VERSION, FLAVOR and ARCH. This means that they are likely obsoleted by other scheduled products using the same combination.

The FLAVOR variable is not used by the test code so this changes simply overrides it making the combination of settings unique. I spawned to set of test jobs on o3 (by simply invoking `./openqa-schedule-mm-ping-test` locally) to verify that this works.

Related ticket: https://progress.opensuse.org/issues/174583#note-48

### DIFF
--- a/openqa-schedule-mm-ping-test
+++ b/openqa-schedule-mm-ping-test
@@ -6,6 +6,7 @@ set -eu -o pipefail
 openqa_url=${openqa_url:-https://openqa.opensuse.org}
 distri=${distri:-opensuse}
 flavor=${flavor:-DVD}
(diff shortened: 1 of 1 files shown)


Martchus:
This is one of the test jobs: https://openqa.opensuse.org/tests/4803164
So the tests still pass despite the different FLAVOR.

perlpunk (approved):

Wabri (approved):

PR #368, opened 2025-01-30
os-autoinst-obs-auto-submit: Mark expected failure explicitly

okurz:
This is a follow-up to 6fd3443 keeping the pipefail flag intact but
explicitly discarding the exit code of the "diff" call which should not
impact the further evaluation. I tested this manually with two differing
fails and checking the exit code like this:

```
set -o pipefail
diff file1 file2 | grep $content
(diff file1 file2 ||:) | grep $content
```

Related progress issue: https://progress.opensuse.org/issues/176316

Related to
* #367

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -143,9 +143,7 @@ handle_auto_submit() {
     $osc co --server-side-source-service-files "$src_project"/"$package"
     if [[ "$package" == "os-autoinst-distri-opensuse-deps" ]]; then
         $osc cat "$submit_target" "$package" "$package.spec" > "$package-factory.spec"
-        set +o pipefail
-        if diff -u "$package"-factory.spec "$src_project/$package/_service:obs_scm:$package".spec | grep "^[+-]Requires"; then
-            set -o pipefail
+        if { diff -u "$package"-factory.spec "$src_project/$package/_service:obs_scm:$package".spec || :; } | grep "^[+-]Requires"; then
             # dependency added or removed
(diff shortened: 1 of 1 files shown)


os-autoinst-obs-auto-submit
```diff
@@ -143,9 +143,7 @@ handle_auto_submit() {
     $osc co --server-side-source-service-files "$src_project"/"$package"
     if [[ "$package" == "os-autoinst-distri-opensuse-deps" ]]; then
         $osc cat "$submit_target" "$package" "$package.spec" > "$package-factory.spec"
-        set +o pipefail
-        if diff -u "$package"-factory.spec "$src_project/$package/_service:obs_scm:$package".spec | grep "^[+-]Requires"; then
-            set -o pipefail
+        if (diff -u "$package"-factory.spec "$src_project/$package/_service:obs_scm:$package".spec ||:) | grep "^[+-]Requires"; then
```
baierjan, line 146: It should be even possible to do it without a sub-shell with a different set of smileys:

```suggestion
        if { diff -u "$package"-factory.spec "$src_project/$package/_service:obs_scm:$package".spec || :; } | grep "^[+-]Requires"; then
```

baierjan (approved):

os-autoinst-obs-auto-submit
okurz: Sure, I don't mind. I guess in theory `{}` is less overhead?

okurz: sure, updated

Martchus (approved):
Looks good, you only need to fix the style check.

PR #369, opened 2025-01-30
Wait until all packages in the snapshot are actually published

baierjan:
First we ensure all wanted packages inside devel:openQA are published properly, after that we "release" them via `osc release`. As this is not an atomic operation and OBS still needs some time to properly publish the copied packages, we can use `osc prjresult -w` to wait until that happens. That will ensure that the staging project (i.e. devel:openQA:testing) does contain some binaries packages to test.

Reference: https://progress.opensuse.org/issues/174451

### DIFF
--- a/trigger-openqa_in_openqa
+++ b/trigger-openqa_in_openqa
@@ -80,6 +80,8 @@ create_devel_openqa_snapshot() {
         echo "Creating snapshots of $package"
         $osc release --no-delay --target-project "$staging_project" --target-repository=openSUSE_Tumbleweed -r openSUSE_Tumbleweed -a x86_64 "$src_project" "$package"
     done
+    echo "Wait until all packages are published under $staging_project"
+    $osc prjresults --watch --xml -r openSUSE_Tumbleweed -a x86_64 "$staging_project"
 }
 
 trigger() {


trigger-openqa_in_openqa
```diff
@@ -80,6 +80,8 @@ create_devel_openqa_snapshot() {
         echo "Creating snapshots of $package"
         $osc release --no-delay --target-project "$staging_project" --target-repository=openSUSE_Tumbleweed -r openSUSE_Tumbleweed -a x86_64 "$src_project" "$package"
     done
+    echo "Wait until all packages are published under $staging_project"
+    $osc prjresults --watch --xml -r openSUSE_Tumbleweed -a x86_64 "$staging_project"
```
Martchus, line 84: Since the packages in the staging project are created via `osc release` I'm not sure whether waiting will help here. At least my theory was that when invoking `osc release` not all packages were published so the packages released into the staging project were stuck with no package being published. If waiting the releasing would help then the retry within the test code should already be sufficient (but it is not as far as I remember; otherwise we would not have this issue in the first place, right?).

baierjan: See the linked ticket and slack conversation there; in our case `osc release` will "release" the package just fine. It is just not published right away, but the job takes a while to complete (hence we can monitor the state with the `osc results` / `osc prjresults`). This change should make sure, all packages which were copied over by the `osc release` are published and we don't end up with starting the test with an empty repository like we currently do in https://openqa.opensuse.org/tests/4819550.

> If waiting the releasing would help then the retry within the test code should already be sufficient.
Yes, unless OBS is stuck with the build process (like it is apparently now) and the release process takes longer. Also you can end up testing Factory instead of devel:openQA:testing like exactly now, because there is nothing preventing you from installing openQA from Factory if devel:openQA:testing is empty.

Martchus (approved):

okurz (commented):
As this isn't entirely obvious please put more details into the git commit message. Rest is fine

baierjan:
> As this isn't entirely obvious please put more details into the git commit message. Rest is fine

I tried to formulate the "story" into the commit.

PR #370, opened 2025-01-30
refactor trigger-openqa-in_openqa extracting package state and snapshot

Wabri:


trigger-openqa_in_openqa
```diff
@@ -56,6 +56,27 @@ download_scenario() {
...
+    state=$($osc results -r openSUSE_Tumbleweed -a x86_64 --no-multibuild "$src_project" "$package")
+
+    case "$state" in
+        *succeeded* | *disabled*)
+            return 0
+            ;;
+        *)
+            return 1
+            ;;
+    esac
```
okurz, line 71: So you are replacing a grep call with bash internal equality match but which unfortunately leads to the rather clunky multi-line case-statement. If you want to use bash then you could look into the bash pattern match operator, something like

```
[[ "$state" =~ 'succeeded\|disabled' ]]
```

but in the original place. Note that this didn't immediately work for me. I would just keep the current state as is

okurz, line 71: So you are replacing a grep call with bash internal equality match but which unfortunately leads to the rather clunky multi-line case-statement. If you want to use bash then you could look into the bash pattern match operator, something like

```
[[ "$state" =~ 'succeeded\|disabled' ]]
```

but in the original place. Note that this didn't immediately work for me. I would just keep the current state as is

okurz (changes requested):

trigger-openqa_in_openqa
```diff
@@ -70,15 +91,12 @@ create_devel_openqa_snapshot() {
     fi
     echo "Checking if all relevant packages are built and published before creating snapshot under $staging_project"
     for package in $auto_submit_packages; do
-        local state
-        state=$($osc results -r openSUSE_Tumbleweed -a x86_64 --no-multibuild "$src_project" "$package")
-        if grep -q -v 'succeeded\|disabled' <<< "$state"; then
+        if ! check_package_state "$package"; then
             echo -e "Package $package is not yet ready for release\n$state" | tee job_post_skip_submission
```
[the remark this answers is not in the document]
Wabri: You are right!

trigger-openqa_in_openqa
Wabri: You're absolutely right... I've overengineer that part with the case statement. Thanks for the feedback and I've reverted that change.

Wabri:
Just a proposal, up to you to close the pr if not needed!

Martchus (approved):
Those are just two lines (now actually three) but I guess it can't hurt.

okurz (approved):

Wabri:
Yes, actually I've open this pr just because while doing review of #365 I was checking locally how to reduce the 25 lines of code of create_devel_openqa_snapshot and that was the result

PR #372, opened 2025-02-03
os-autoinst-obs-auto-submit: Do not fail on expected diff output

baierjan:
Reference: https://progress.opensuse.org/issues/176316

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -121,6 +121,12 @@ sync_changesrevision() {
     fi
 }
 
+get_spec_changes_in_requirements() {
+    local package=$1
+    local project=$2
+    { diff -u "$package-factory.spec" "$project/$package/_service:obs_scm:$package.spec" || :; } | grep '^[+-]Requires'
+}
+
 generate_os-autoinst-distri-opensuse-deps_changelog() {
     dir=$1
     package=$2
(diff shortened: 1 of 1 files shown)


Martchus (approved):

os-autoinst-obs-auto-submit
```diff
@@ -129,7 +129,7 @@ generate_os-autoinst-distri-opensuse-deps_changelog() {
         echo "-------------------------------------------------------------------"
         echo "$(LANG=c date -u) - Dominik Heidler <dheidler@suse.de>"
         echo
-        diff -u "$package"-factory.spec "$dir"/"$package"/_service:obs_scm:"$package".spec \
+        { diff -u "$package"-factory.spec "$dir"/"$package"/_service:obs_scm:"$package".spec || :; } \
```
okurz, line 132: *sigh* can we extract that into a function to prevent the confusion about the duplicate code? Like

```bash

spec_has_changes_in_requires() {
    local dir="$1"
    local package="$2"
    { diff -u "$package"-factory.spec "$dir"/"$package"/_service:obs_scm:"$package".spec || :; } | grep "^[+-]Requires"
}
…
-    { diff -u "$package"-factory.spec "$dir"/"$package"/_service:obs_scm:"$package".spec || :; } | grep "^[+-]Requires"
+    spec_has_changes_in_requires $dir $package
```

baierjan: I didn't even notice it is indeed diffing the very same files. Yes that's a good candidate for refactor then. Will try to do that.

baierjan:
I added a new commit which extracts the common code into a new function.

Martchus (approved):

okurz (approved):

PR #373, opened 2025-02-06
Make it harder to accidentally delete package from devel:openQA

baierjan:


okurz (approved):

os-autoinst-obs-auto-submit
```diff
@@ -178,7 +178,7 @@ prefix="${prefix:-""}"
 osc="${osc:-"$prefix retry -e -- osc"}"
 
 # submit from staging project if specified
-[[ $staging_project && $staging_project != none ]] && src_project=$staging_project from_staging=1
+[[ $staging_project && $staging_project != none && $staging_project != "$src_project" ]] && src_project=$staging_project from_staging=1
```
b10n1k, line 181: Could we update also the comment to reflect exactly the condition?

PR #375, opened 2025-02-07
Avoid python trace from subprocess.check_output

b10n1k:
Wrap the subprocess.check_output inside a try/catch to log a custom error. The check_output run the command which sets check parameter to True and causes the trace output, when the exit code is non-zero.

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -7,8 +7,8 @@ import json
 import logging
 import os
 import re
-import subprocess
 import sys
+import subprocess
 from urllib.parse import urlparse, urlunparse
 
 import requests
@@ -111,18 +111,31 @@ client_args = [
 def call(cmds, dry_run=False):
     try:
         log.debug("call: %s" % cmds)
-        return subprocess.check_output(
-            (["echo", "Simulating: "] if dry_run else []) + cmds
-        ).decode("utf-8")
-    except subprocess.CalledProcessError as e:
-        subprocess_error = json.dumps({
-            "cmd": cmds[0],
-            "returncode": e.returncode,
-            "stderr": e.stderr if e.stderr else None
-        })
-        log.error(json.loads(subprocess_error))
-        print("DEBUG: Calling sys.exit()") 
-        sys.exit(e.returncode)
+        return {
+            "output": subprocess.check_output(
+            (["echo", "Simulating: "] if dry_run else []) + cmds,
+            stderr=subprocess.STDOUT,
+            shell=True).decode("utf-8"),
+            "error": False
+        }
+    except subprocess.SubprocessError as err:
+        url_pattern = re.compile(r"(https://.*)/tests/(\d+)")
+        url_match = next(
+            ((i, item) for i, item in enumerate(cmds) 
+            if isinstance(item, str) and url_pattern.match(item)),
+            None
+        )
+        found = url_pattern.match(url_match[1])
+        error_info = {
+            "cmd": cmds,
+            "failed_id": found.group(1),
+            "host": found.group(0),
+            "err_type": type(err).__name__,
+            "stderr": err.stderr if err.stderr else "No error putput",
+            "error": True
+        }
+        log.error(error_info["stderr"])
+        return json.dumps(error_info)
 
 def openqa_comment(job, host, comment, dry_run):
     args = client_args + [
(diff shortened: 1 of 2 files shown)


b10n1k:
NOTE: I submit it as draft because I am not sure about the e.output(L119). Need to check wheather is actually needed or wheather contains any useful output.

openqa-trigger-bisect-jobs
```diff
@@ -223,7 +226,7 @@ def main(args):
         )
         return
     if job.get("clone_id") is not None:
-        log.info("Job %d already has a clone, skipping bisection" % job["id"])
+        log.info("Job %d already a clone, skipping bisection" % job["id"])
```
okurz, line 229: ?

pytest.ini.ign
```diff
@@ -0,0 +1,8 @@
+[pytest]
```
okurz, line 1: unrelated. please remove. Consider to put it into a separate PR.

tests/__init__.py
okurz, line 1: why?

tests/test_trigger_bisect_jobs.py
```diff
@@ -203,6 +204,19 @@ def test_triggers():
...
+    args.url = "https://openqa.opensuse.org/tests/7848818"
+    openqa.fetch_url = MagicMock(side_effect=mocked_fetch_url)
+    openqa.openqa_clone = MagicMock(side_effect=CalledProcessError(returncode=255,
+                                                                   cmd='foo',
+                                                                   stderr="foo failed with return code 255"))
+    with pytest.raises(CalledProcessError) as expected:
+        openqa.main(args)
+    assert expected.value.returncode == 255
+    assert expected.value.cmd == 'foo'
+    assert "Command 'foo' returned non-zero exit status 255." in str(expected.value)
```
okurz, line 219: This part I don't understand. It looks like you assert what should happen *before* your change but you want to avoid the stack trace so you should also assert that there is *no* traceback showing up, right?

okurz (changes requested):

openqa-trigger-bisect-jobs
b10n1k: dont know. I reverted

tests/test_trigger_bisect_jobs.py
b10n1k: Sure I could. When I submitted, I didnt mean the end of the commit. I had to push a draft as you asked to provide one at the end of the day. I will work on it

pytest.ini.ign
b10n1k: I already did. they belong to separate commit or PR. submitted by accident btw

tests/__init__.py
b10n1k: first of all, I have already reverted it. the reason I had added this was because I was encountering issues with some experiments on pytest and I was wondering if this was needed for those changes. ignore

openqa-trigger-bisect-jobs
```diff
@@ -109,11 +109,14 @@ client_args = [
...
-        (["echo", "Simulating: "] if dry_run else []) + cmds
-    ).decode("utf-8")
-
+    try:
+        log.debug("call: %s" % cmds)
+        return subprocess.check_output(
+            (["echo", "Simulating: "] if dry_run else []) + cmds
+        ).decode("utf-8")
+    except subprocess.CalledProcessError as e:
+        log.error("%s failed with return code %s", cmds[0], e.returncode)
```
Martchus, line 118: `e.returncode` might actually be a negative signal number. Then this error message might be confusing. Not a big deal but this could be taken into account.

There should probably be also a more generic `except` handler after this (maybe one that catches all kinds of exceptions) to handle other error cases as well, e.g. when the executable wasn't found or the execution wasn't permitted.

Martchus (changes requested):

openqa-trigger-bisect-jobs
```diff
@@ -109,11 +109,14 @@ client_args = [
 
 
 def call(cmds, dry_run=False):
-    log.debug("call: %s" % cmds)
-    return subprocess.check_output(
-        (["echo", "Simulating: "] if dry_run else []) + cmds
-    ).decode("utf-8")
-
+    try:
```
perlpunk, line 112: This is now making every system call non-fatal.
We still want unsuccessful system calls aborting the script, just without the long stack trace.
Also returning the stderr output might now trigger other exceptions for code that expects the system call to return JSON.

okurz (changes requested):
I am not sure about what you intend with your later comments. I think the most important point to cover is the comment from perlpunk in https://github.com/os-autoinst/scripts/pull/375/files#r1949474145

openqa-trigger-bisect-jobs
okurz: I think all this goes way too far. For now we should focus on only the original problem about "Current job … will fail, because the repositories for the below updates are unavailable" because there we know what we should do

PR #376, opened 2025-02-12
Prevent stack trace from subprocess.check_output

b10n1k:
This commit makes sure that the python stack trace is handled appropiately in main function, and logs the error neverthless, for easy tracking. It uses SubprocessError is order to get a general variaty of exceptions from subprocess.

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -111,7 +111,8 @@ client_args = [
 def call(cmds, dry_run=False):
     log.debug("call: %s" % cmds)
     return subprocess.check_output(
-        (["echo", "Simulating: "] if dry_run else []) + cmds
+        (["echo", "Simulating: "] if dry_run else []) + cmds,
+        stderr=subprocess.STDOUT
     ).decode("utf-8")
 
 
@@ -299,10 +300,19 @@ def main(args):
             ]
         )
 
-        out = openqa_clone(
-            params,
-            args.dry_run,
-        )
+        try:
+            out = openqa_clone(
+                params,
+                args.dry_run,
+            )
+        except subprocess.SubprocessError as err:
+            process_output = err.output if err.output else ""
+            log.error(f"{type(err).__name__}: {process_output} ")
+            if 'the repositories for the below updates are unavailable' not in str(process_output):
+                raise
+            else:
+                sys.exit(0)
+
         created_job_ids = []
         try:
             created_job_ids = json.loads(out).values()
--- a/tests/test_trigger_bisect_jobs.py
+++ b/tests/test_trigger_bisect_jobs.py
@@ -7,9 +7,11 @@ import importlib.machinery
 import importlib.util
 import json
 import os.path
-from unittest.mock import MagicMock, call
+import sys
+from unittest.mock import MagicMock, call, patch
 from urllib.parse import urlparse
 
+import pytest
 import requests
 
 rootpath = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
@@ -66,6 +68,31 @@ cmds = [
     "OPENQA_INVESTIGATE_ORIGIN=https://openqa.opensuse.org/tests/7848818",
 ]
 
+def test_catch_CalledProcessError(caplog):
+    import subprocess
+    args = args_factory()
+    args.url = "https://openqa.opensuse.org/tests/7848818"
+    openqa.fetch_url = MagicMock(side_effect=mocked_fetch_url)
+    cmd_args = ["openqa-clone-job"]
+    exp_err = "returned non-zero exit status 255."
+    with patch("subprocess.check_output",
+        side_effect=subprocess.CalledProcessError(returncode=255,
+                                                  cmd=cmd_args,
+                                                  output=exp_err)):
+        with pytest.raises(subprocess.CalledProcessError) as e:
+            openqa.main(args)
+    assert(e.value.returncode == 255)        
+    assert f"CalledProcessError: {exp_err}" in caplog.text
+
+    exp_err = "Current job 7848818 will fail, because the repositories for the below updates are unavailable"
+    with patch("subprocess.check_output",
+        side_effect=subprocess.CalledProcessError(returncode=255,
+                                                  cmd=cmd_args,
+                                                  output=exp_err)):
+        with pytest.raises(SystemExit) as e:
+            openqa.main(args)
+    assert(e.value.code == 0)        
+    assert f"CalledProcessError: {exp_err}" in caplog.text
 
 def test_clone():
     openqa.call = MagicMock(side_effect=mocked_call)


openqa-trigger-bisect-jobs
```diff
@@ -110,9 +110,14 @@ client_args = [
...
-        (["echo", "Simulating: "] if dry_run else []) + cmds
-    ).decode("utf-8")
+    try:
+        return subprocess.check_output(
+            (["echo", "Simulating: "] if dry_run else []) + cmds,
+            stderr=subprocess.STDOUT, shell=True
+        ).decode("utf-8")
+    except subprocess.SubprocessError as err:
+        p_stderr = err.stderr if err.stderr else "No error putput"
+        raise RuntimeError(f"Command failed: {cmds[0]} ({p_stderr})") from err
```
okurz, line 120: As already explained today please remove this. Only catch exceptions within `main`

openqa-trigger-bisect-jobs
```diff
@@ -146,8 +151,10 @@ def openqa_clone(
     default_opts=["--skip-chained-deps", "--json-output", "--within-instance"],
     default_cmds=["_GROUP=0"],
 ):
-    return call(["openqa-clone-job"] + default_opts + cmds + default_cmds, dry_run)
-
+    try:
+        return call(["openqa-clone-job"] + default_opts + cmds + default_cmds, dry_run)
+    except RuntimeError as err:
+        raise
```
okurz, line 157: As already explained today please remove this. Only catch exceptions within `main`

openqa-trigger-bisect-jobs
```diff
@@ -298,11 +305,20 @@ def main(args):
...
             ]
         )
+        import pdb
+        #pdb.set_trace()
+        try:
+            out = openqa_clone(
+                params,
+                args.dry_run,
+            )
+        except RuntimeError as err:
```
okurz, line 315: I thought it was subprocess.SubprocessError?

okurz (changes requested):

openqa-trigger-bisect-jobs
```diff
@@ -298,11 +305,20 @@ def main(args):
...
+            err_comment = str(err)
+            log.error(err_comment)
+            #openqa_comment(base_url, job_id, err_comment, args.dry_run)
+            #print(f"get here")
+            sys.exit()
```
perlpunk, line 320: If the `openqa_clone` fails for a different reason it should still exit with the original return code. We only want to handle the mentioned case.

openqa-trigger-bisect-jobs
b10n1k: I made the change u asked. But I have to say I not sure about that. in theory you can call the openqa_clone or call in other places and then it wont catch the exception. I can think one or two other reasons but I think is minor

openqa-trigger-bisect-jobs
b10n1k: lets close this as we handle this on https://github.com/os-autoinst/scripts/pull/376#discussion_r1952767628

openqa-trigger-bisect-jobs
b10n1k: `mentioned case`?
So you want to pass the `returncode`?

openqa-trigger-bisect-jobs
```diff
@@ -299,10 +300,17 @@ def main(args):
...
-        )
+        try:
+            out = openqa_clone(
+                params,
+                args.dry_run,
+            )
+        except subprocess.SubprocessError as err:
+            p_stderr = err.stderr if err.stderr else "No error putput"
+            log.error(f"{type(err).__name__}: {p_stderr} ")
+            #openqa_comment(base_url, job_id, err_comment, args.dry_run)
```
perlpunk, line 311: We said that we only want AC1 for the first PR, so you can leave that out.
But it's good to keep that in mind.

openqa-trigger-bisect-jobs
perlpunk: `openqa-clone-job` can fail for several reasons. Only if the error says "Current job ... will fail, because the repositories for the below updates are unavailable" we want to treat it as non-fatal.
In all other cases it should exit with something other than zero.

openqa-trigger-bisect-jobs
```diff
@@ -111,7 +111,8 @@ client_args = [
 def call(cmds, dry_run=False):
     log.debug("call: %s" % cmds)
     return subprocess.check_output(
-        (["echo", "Simulating: "] if dry_run else []) + cmds
+        (["echo", "Simulating: "] if dry_run else []) + cmds,
+        stderr=subprocess.STDOUT, shell=True
```
perlpunk, line 115: Why do we need that?

b10n1k: I read that this ensures capturing the error and display it, especially when you redirect the the stderr. the problem I had, from the old PR was how to catch some messages on the test.

openqa-trigger-bisect-jobs
b10n1k: Can I leave it or add a todo or...?

perlpunk: Remove it

openqa-trigger-bisect-jobs
b10n1k: that will take some time to figure it out. I see openqa_clone_job dies. I will take a look

perlpunk: Maybe you can just re-raise the original exception in those other cases.

openqa-trigger-bisect-jobs
b10n1k: as I understand it, yes. but it might not be mandatory. the diff is that it involves `sh` rather than direct execution and it looks useful when you run shell commands. I dont know the exact impact though.

perlpunk: We shouldn't use something just because it looks useful, but we don't know what it does :)

https://docs.python.org/3/library/subprocess.html#frequently-used-arguments

> If passing a single string, either `shell` must be `[True]`(https://docs.python.org/3/library/constants.html#True) (see below) or else the string must simply name the program to be executed without specifying any arguments.

We are not executing any shell commands in this script, we are only doing system calls. We are passing the program and its arguments as an array, not a single string.

The shell is needed for example when we do something like `command | grep`.

b10n1k: I remove it. I have some questions but we can discuss it out of here for less noise. thanks

openqa-trigger-bisect-jobs
b10n1k: I submitted my approach but it is untested and I am not sure if it works actually in a real scenario. I assume that the error will be raised by `SubprocessError`

openqa-trigger-bisect-jobs
```diff
@@ -111,7 +111,8 @@ client_args = [
 def call(cmds, dry_run=False):
     log.debug("call: %s" % cmds)
     return subprocess.check_output(
-        (["echo", "Simulating: "] if dry_run else []) + cmds
+        (["echo", "Simulating: "] if dry_run else []) + cmds,
+        stderr=subprocess.STDOUT
```
okurz, line 115: Why change this? I don't seev why that should be needed from the git commit messages

openqa-trigger-bisect-jobs
```diff
@@ -146,8 +147,8 @@ def openqa_clone(
     default_opts=["--skip-chained-deps", "--json-output", "--within-instance"],
     default_cmds=["_GROUP=0"],
 ):
-    return call(["openqa-clone-job"] + default_opts + cmds + default_cmds, dry_run)
 
+    return call(["openqa-clone-job"] + default_opts + cmds + default_cmds, dry_run)
```
okurz, line 151: Why move this?

openqa-trigger-bisect-jobs
```diff
@@ -299,10 +300,19 @@ def main(args):
...
+                params,
+                args.dry_run,
+            )
+        except subprocess.SubprocessError as err:
+            p_stderr = err.stderr if err.stderr else "No error putput"
+            log.error(f"{type(err).__name__}: {p_stderr} ")
+            if 'the repositories for the below updates are unavailable' not in str(err.output):
+                sys.exit(err.returncode)
+            else:
+                log.debug(f"Ignore {type(err).__name__}")
```
okurz, line 314: What do you mean with "ignore" here?

openqa-trigger-bisect-jobs
b10n1k: u are right. I should be able to write a comment about it but I dont have a good reasoning. During testing this and that I found that this would ensure the output in the console. It worked, it stayed. Later on, ofc things changed.

openqa-trigger-bisect-jobs
```diff
@@ -299,10 +300,19 @@ def main(args):
...
-            params,
-            args.dry_run,
-        )
+        try:
+            out = openqa_clone(
+                params,
+                args.dry_run,
+            )
+        except subprocess.SubprocessError as err:
+            p_stderr = err.stderr if err.stderr else "No error putput"
```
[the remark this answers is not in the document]
b10n1k: s/No error putput/No error message/?
and I have a small correction, but I would like to check first the doc

openqa-trigger-bisect-jobs
b10n1k: Ignore <whatever_raised_Exception>.

perlpunk, line 314: You still need to exit here. Otherwise it would continue trying to clone the same job with different parameters.

I ran this with a manipulated openqa-clone-job (always dying) and early returns in `openqa_comment` and `openqa_set_job_prio`. This is also how you can test it manually. It took me only a minute.

This is what I got:
```
% ./openqa-trigger-bisect-jobs --url https://openqa.opensuse.org/tests/4847227
ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: No error putput 
Traceback (most recent call last):
  File "./openqa-trigger-bisect-jobs", line 320, in main
    created_job_ids = json.loads(out).values()
UnboundLocalError: local variable 'out' referenced before assignment
```
That's because it reaches line 320, when it should actually have exited.

So, I added a `sys.exit()` here.

Now I got:
```
ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: No error putput
```
There is no error output, because you redirect it to stdout with `stderr=subprocess.STDOUT`. Why not simply use `err.output`?
I did that. Now I get
```
ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: b'Current job 4847227 will fail, because the repositories for the below updates are unavailable\n'
```
The `b'...' is a bit annoying, but ok.

Now for a different error cause I get this:
```
ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: b'lala\n'
```
That isn't enough information IMHO. If we see this in the log, we might not immediately know where this failed, and which command it was.

So maybe just re-raise the original exception in this case, like I suggested earlier.

How to unittest this, I don't know right now. I'm still not that familiar with the pytest.patch stuff.

perlpunk:
Here is a suggestion for testing the known error message case:
```python
def test_catch_CalledProcessError2(caplog):
    import subprocess
    args = args_factory()
    args.url = "https://openqa.opensuse.org/tests/7848818"
    openqa.fetch_url = MagicMock(side_effect=mocked_fetch_url)
    exp_err = "the repositories for the below updates are unavailable"

    with patch("subprocess.check_output",
        side_effect=subprocess.CalledProcessError(returncode=255,
                                                  cmd=["openqa-clone-job", "_GROUP=0"],
                                                  output=exp_err)):
        with pytest.raises(SystemExit) as e:
            openqa.main(args)

    assert(e.value.code == 0)
    assert exp_err in caplog.text


def test_clone():
    openqa.call = MagicMock(side_effect=mocked_call)
    openqa.openqa_clone(cmds, dry_run=False)
```

openqa-trigger-bisect-jobs
okurz: So you are instructing the reader to ignore the exception? How about "Terminated with ..."

b10n1k: Lets follow up on https://github.com/os-autoinst/scripts/pull/376#discussion_r1953508580

b10n1k: > You still need to exit here. Otherwise it would continue trying to clone the same job with different parameters.
> 

I thought that "not fatal" (https://github.com/os-autoinst/scripts/pull/376#discussion_r1952832770) means `let it continue. So it should exit anyway but return `err.returncode` if `the repositories for the below updates are unavailable` and `0` in all the others?!

> I ran this with a manipulated openqa-clone-job (always dying) and early returns in `openqa_comment` and `openqa_set_job_prio`. This is also how you can test it manually. It took me only a minute.
> 
> This is what I got:
> 
> ```
> % ./openqa-trigger-bisect-jobs --url https://openqa.opensuse.org/tests/4847227
> ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: No error putput 
> Traceback (most recent call last):
>   File "./openqa-trigger-bisect-jobs", line 320, in main
>     created_job_ids = json.loads(out).values()
> UnboundLocalError: local variable 'out' referenced before assignment
> ```
> 
> That's because it reaches line 320, when it should actually have exited.
> 
> So, I added a `sys.exit()` here.
> 
> Now I got:
> 
> ```
> ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: No error putput
> ```
> 
> There is no error output, because you redirect it to stdout with `stderr=subprocess.STDOUT`. Why not simply use `err.output`? I did that. Now I get
> 
> ```
> ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: b'Current job 4847227 will fail, because the repositories for the below updates are unavailable\n'
> ```
> 

`stderr=subprocess.STDOUT` redirects the the stderr to the stdout.
from the doc: `
If you ran the process with stderr=subprocess.STDOUT, stdout and stderr will be combined in this attribute, and stderr will be None.`
u are right. it needs `output` or `stdout`. thanks

> The `b'...' is a bit annoying, but ok.
> 
> Now for a different error cause I get this:
> 
> ```
> ERROR:./openqa-trigger-bisect-jobs:CalledProcessError: b'lala\n'
> ```
> 
> That isn't enough information IMHO. If we see this in the log, we might not immediately know where this failed, and which command it was.
> 
> So maybe just re-raise the original exception in this case, like I suggested earlier.
> 

the whole point is to not have the stack trace, right? I dont know how you mean `for a different error`. Even so, it is not part of the requirements, having the exception in the `call` could give you those info. let me do the other changes and we can talk about that later.
 
> How to unittest this, I don't know right now. I'm still not that familiar with the pytest.patch stuff.

perlpunk: > the whole point is to not have the stack trace, right? I dont know how you mean `for a different error`.

The whole point is to not have a stack trace for a **known error**.
openqa-clone-job can "fail" in the mentioned case. I would even say it shouldn't be a failure but rather it should exit with zero and return some info in the JSON, but currently it's implemented as a failure.
And it's happening quite often, but it's a failure that we have to accept and simply not trigger any bisect jobs. For a known failure we don't want lengthy stack traces in our logs.

But openqa-clone-job can still fail for other reasons. Of course I can't tell you the reasons now. But if it fails for a reason that is not the above mentioned reason, then it's ok to have a stack trace. We can then investigate why it happened and fix the root cause.

perlpunk: > I thought that "not fatal" ([#376 (comment)](https://github.com/os-autoinst/scripts/pull/376#discussion_r1952832770)) means `let it continue. So it should exit anyway but return `err.returncode`if`the repositories for the below updates are unavailable`and`0` in all the others?!

There is a loop that runs `openqa-clone-job` on the same original job several times, just with different parameters, to bisect which issue is the culprit.
Now the first invocation of `openqa-clone-job` fails with the mentioned error, basically telling us that cloning the job doesn't make sense because the repos are gone, and the clone would always fail.

We have to accept that we can't do any bisection for this job then, and also skip creating the other clones, because we will get that error message every time, because it will run `openqa-clone-job` on the same job. So it should just exit or at least end the loop.

And in the next PR for AC2 it should also create a comment on the job and exit after that.

tests/test_trigger_bisect_jobs.py
```diff
@@ -66,6 +68,38 @@ def mocked_call(cmds, dry_run=False):
...
+    args.url = "https://openqa.opensuse.org/tests/7848818"
+    openqa.fetch_url = MagicMock(side_effect=mocked_fetch_url)
+    cmd_args = ["openqa-clone-job",
+                "--skip-chained-deps",
+                "--json-output",
+                "--within-instance",
+                "https://openqa.opensuse.org/tests/7848818",
+                "OS_TEST_ISSUES=21770,21926,21954,22030,22077,22085,22192",
+                "TEST=foo:investigate:bisect_without_21637",
+                "OPENQA_INVESTIGATE_ORIGIN=https://openqa.opensuse.org/tests/7848818","_GROUP=0"]
```
perlpunk, line 83: do we really need all of those parameters? When I tested it, it didn't make any difference.

perlpunk:
I just edited AC! to say "No Python stacktraces in openqa-gru service logs for known errors"

tests/test_trigger_bisect_jobs.py
b10n1k: likely not

openqa-trigger-bisect-jobs
```diff
@@ -299,10 +299,19 @@ def main(args):
...
-        )
+        try:
+            out = openqa_clone(
+                params,
+                args.dry_run,
+            )
+        except subprocess.SubprocessError as err:
+            p_stderr = err.output if err.output else ""
+            log.error(f"{type(err).__name__}: {p_stderr} ")
+            if 'the repositories for the below updates are unavailable' not in str(err.output):
```
Martchus, line 310: It would make sense to use the variable declared above here as well.

tests/test_trigger_bisect_jobs.py
```diff
@@ -66,6 +68,31 @@ def mocked_call(cmds, dry_run=False):
     "OPENQA_INVESTIGATE_ORIGIN=https://openqa.opensuse.org/tests/7848818",
 ]
 
+def test_catch_CalledProcessError(caplog):
+    import subprocess
+    args = args_factory()
+    args.url = "https://openqa.opensuse.org/tests/7848818"
+    openqa.fetch_url = MagicMock(side_effect=mocked_fetch_url)
+    cmd_args = ["openqa-clone-job",]
```
Martchus, line 76:
```suggestion
    cmd_args = ["openqa-clone-job"]
```

Martchus (changes requested):

openqa-trigger-bisect-jobs
```diff
@@ -299,10 +299,19 @@ def main(args):
...
+                sys.exit(err.returncode)
```
perlpunk, line 311:
```suggestion
                raise err
```

b10n1k: this prints again a stack trace. why?

b10n1k: But i know all that. I am just focused to the fact that the ticket was about no stack trace. And what we have done is to capture the SubprocessError and log it. if we need to track other problem, it would make sense to create custom Exceptions. But I avoided to go this route because I knew i will have other problems with that.

openqa-trigger-bisect-jobs
```diff
@@ -299,10 +299,22 @@ def main(args):
...
+        except subprocess.SubprocessError as err:
+            process_output = err.output if err.output else ""
+            log.error(f"{type(err).__name__}: {process_output} ")
+            if 'the repositories for the below updates are unavailable' not in str(process_output):
+                raise
+            else:
+                sys.exit(0)
+        except Exception as err:
+            err.add_note("openqa_clone failed on unknown cause.")
+            raise
```
okurz, line 316: why is this necessary. Isn't any not catched exception raised further anyway?

b10n1k:
> So can you revert this?

But the code have adjusted based on others' with this "accepted". it is not a problem. maybe just a different approach

> why is this necessary. Isn't any not catched exception raised further anyway?

I dont know if I got confused but we handle subprocess errors. Comments insists that there might be other unknown errors. Although....

openqa-trigger-bisect-jobs
b10n1k: But the code have adjusted based on others' with this "accepted". it is not a problem. maybe just a different approach

openqa-trigger-bisect-jobs
Martchus: I think @perlpunk comment was only about the `if` case in the `SubprocessError` handler. We don't need to add a generic handler just to log an additional error message.

Martchus (changes requested):

okurz:
https://github.com/os-autoinst/scripts/pull/377 was merged with your changes included.

PR #377, opened 2025-02-13
Catch known error in openqa-trigger-bisect-jobs

perlpunk:
Added some commits to #376, see commit messages

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -110,10 +110,14 @@ client_args = [
 
 def call(cmds, dry_run=False):
     log.debug("call: %s" % cmds)
-    return subprocess.check_output(
+    res = subprocess.run(
         (["echo", "Simulating: "] if dry_run else []) + cmds,
-        stderr=subprocess.STDOUT
-    ).decode("utf-8")
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
+    if len(res.stderr):
+        log.warning(f"call() {cmds[0]} stderr: {res.stderr}")
+    res.check_returncode()
+    return res.stdout.decode("utf-8");
 
 
 def openqa_comment(job, host, comment, dry_run):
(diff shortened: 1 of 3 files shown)


openqa-trigger-bisect-jobs
```diff
@@ -299,10 +304,17 @@ def main(args):
...
-            args.dry_run,
-        )
+        try:
+            out = openqa_clone(
+                params,
+                args.dry_run,
+            )
+        except subprocess.SubprocessError as err:
+            if 'the repositories for the below updates are unavailable' in str(err.stderr):
+                sys.exit(0)
```
b10n1k, line 314: I am not sure about that but as I start getting a better view, I am wonder why we need to exit and not just log and continue. no need to answer. just wonder

openqa-trigger-bisect-jobs
```diff
@@ -110,9 +110,14 @@ client_args = [
...
-    return subprocess.check_output(
-        (["echo", "Simulating: "] if dry_run else []) + cmds
-    ).decode("utf-8")
+    res = subprocess.run(
+        (["echo", "Simulating: "] if dry_run else []) + cmds,
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
+    if len(res.stderr):
+        log.warning(f"call() {cmds[0]} stderr: {res.stderr}")
+    res.check_returncode()
```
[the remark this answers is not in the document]
perlpunk: Yes, but this way I have more control over stdout and stderr. I don't need to redirect stderr, and I can log it before calling check_returncode which raises the exception. See commit message

openqa-trigger-bisect-jobs
perlpunk: because we will get the same error message for every item in the loop. there is no sense in continuing if we don't schedule any bisecting jobs

Martchus (approved):

PR #378, opened 2025-02-17
Comment job on bisect openqa-clone-job error

b10n1k:
Simply invoke the openqa-comment only when the SubprocessError raised with the specific error logged.

### DIFF
--- a/openqa-trigger-bisect-jobs
+++ b/openqa-trigger-bisect-jobs
@@ -311,6 +311,8 @@ def main(args):
             )
         except subprocess.SubprocessError as err:
             if 'the repositories for the below updates are unavailable' in str(err.stderr):
+                extra_comment = "Not triggering any bisect jobs because: "
+                openqa_comment(job_id, base_url, f"{extra_comment}{err.stderr}", args.dry_run)
                 sys.exit(0)
             else:
                 raise
--- a/tests/test_trigger_bisect_jobs.py
+++ b/tests/test_trigger_bisect_jobs.py
@@ -8,6 +8,7 @@ import importlib.util
 import json
 import os.path
 import sys
+import re
 from unittest.mock import MagicMock, call, patch
 from urllib.parse import urlparse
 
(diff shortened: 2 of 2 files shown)


openqa-trigger-bisect-jobs
```diff
@@ -311,6 +311,7 @@ def main(args):
             )
         except subprocess.SubprocessError as err:
             if 'the repositories for the below updates are unavailable' in str(err.stderr):
+                openqa_comment(job_id, base_url, err.stderr, args.dry_run)
```
perlpunk, line 314: I think users would appreciate a bit more explanation here, e.g. "Not triggering any bisect jobs because: {err.stderr}"

Martchus: Yes, just stderr without context is not helpful.

Martchus (changes requested):

tests/test_trigger_bisect_jobs.py
```diff
@@ -89,10 +89,23 @@ def test_catch_CalledProcessError(caplog):
...
-
+        mocked.assert_has_calls([
+            call([
+            'openqa-cli', 'api', '--header',
+            'User-Agent: openqa-trigger-bisect-jobs (https://github.com/os-autoinst/scripts)',
+            '--host', 'https://openqa.opensuse.org', '-X', 'POST',
+            'jobs/7848818/comments',
+            'text=Current job 7848818 will fail, because the repositories for the below updates are unavailable'
+            ], stdout=-1, stderr=-1)
+        ])
```
okurz, line 108: I recommend to strip this down to only the relevant bits that are necessary for the test

openqa-trigger-bisect-jobs
okurz: Also this would add another comment but what we should do is only save some extra text that goes into the final comment that is written anyway in line [L330](https://github.com/os-autoinst/scripts/pull/378/files#diff-cd273835e66c0f36337b4df0b3a844b8d5486cf47b5ec0733c4479906c6871a9L330)

okurz (changes requested):

openqa-trigger-bisect-jobs
okurz: I see. In that case disregard my former comment.

tests/test_trigger_bisect_jobs.py
b10n1k: I cant. `mocked.assert_has_calls` doesnt pass if the call doesnt match as a whole. And this is good thing

openqa-trigger-bisect-jobs
```diff
@@ -311,6 +311,8 @@ def main(args):
             )
         except subprocess.SubprocessError as err:
             if 'the repositories for the below updates are unavailable' in str(err.stderr):
+                extra_comment = "Bisects exits due to: "
```
okurz, line 314: I don't understand the language here. What do you mean with "Bisects exits"?

okurz (changes requested):

openqa-trigger-bisect-jobs
b10n1k: I wanted to say the no more bisects jobs. but I guess is better for you `openqa-clone-job` exits... right?

okurz: Sorry, I also don't understand that language.

perlpunk (approved):

Martchus (approved):

PR #381, opened 2025-03-20
Add investigation group ID based on the host in clone settings

b10n1k:
Adjust the passing GROUP_ID in the clone. Keep the default groupless in case host is not o3/osd

https://progress.opensuse.org/issues/164988

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -12,6 +12,7 @@ set -euo pipefail
 host="${host:-"openqa.opensuse.org"}"
 scheme="${scheme:-"https"}"
 host_url="$scheme://$host"
+investigation_gid=${investigation_gid:-0}
 dry_run="${dry_run:-"0"}"
 verbose="${verbose:-"false"}"
 prio_add="${prio_add:-"100"}"
@@ -50,7 +51,7 @@ clone() {
     base_name="$(echo "$clone_job_data" | runjq -r '.job.test')" || return $?
     name="$base_name:investigate$name_suffix"
     base_prio=$(echo "$clone_job_data" | runjq -r '.job.priority') || return $?
-    clone_settings=('_TRIGGER_JOB_DONE_HOOK=1' '_GROUP_ID=0' 'BUILD=')
+    clone_settings=('_TRIGGER_JOB_DONE_HOOK=1' "_GROUP_ID=$investigation_gid" 'BUILD=')
     if [[ $refspec ]]; then
         vars_json=$(fetch-vars-json "$origin")
         testgiturl=$(echo "$vars_json" | runjq -r '.TEST_GIT_URL')
--- a/test/02-investigate.t
+++ b/test/02-investigate.t
@@ -3,7 +3,7 @@
 source test/init
 bpan:source bashplus +err +fs +sym
 
-plan tests 88
+plan tests 90
 
 host=localhost
 url=https://localhost
@@ -87,6 +87,9 @@ client-get-job() {
         30032)
             echo '{"job": { "test": "vim:investigate:retry", "result": "passed", "settings": {"OPENQA_INVESTIGATE_ORIGIN": "3003"} } }'
             ;;
+        10035)
+            echo '{"job": { "test": "vim", "result": "failed", "parent_group": "foo", "group": "bar", "settings": {"OPENQA_INVESTIGATE_ORIGIN": "10035"} } }'
+            ;;
         404)
             echo '404 Not Found'
             ;;
@@ -132,6 +135,9 @@ client-get-job-comments() {
         10032)
             echo '[{"id": 1236, "text":"Starting investigation for job 10032"},{"id": 1237, "text":"Starting investigation for job 10032"}]'
             ;;
+        10035)
+            echo '[{"id": 1236, "text":"Starting investigation for job 10035"}]'
+            ;;
         3000)
             echo '[{"id": 1236, "text":"Automatic investigation jobs for job\n*a:investigate:retry*: t#30001\n*a:investigate:last_good_tests:coffee*: t#30002\n*a:investigate:last_good_build:2001*: t#30003\n*a:investigate:last_good_tests_and_build:coffee+2001*: t#30004"}]'
             ;;
@@ -158,6 +164,9 @@ client-post-job-comment() {
             ;;
         10032)
             echo '{"id": 1237}'
+            ;;
+        10035)
+            echo '{"id": 1236}'
             ;;
         3000)
             warn "Commenting 3000 ($@)"
@@ -210,6 +219,9 @@ get-dependencies-ajax() {
             # job with cancelled job in the cluster (should be treated like a done job)
             echo '{"cluster":{"cluster_foo":[10028,10031],"cluster_bar":[29]}, "edges":[], "nodes":[{"id":10028,"state":"cancelled","result":"none"},{"id":10031,"state":"done","result":"failed"}]}'
             ;;
+        10035)
+            echo '{"cluster":{},"edges":[], "nodes":[{"id":10035,"state":"cancelled","result":"none"}]}'
+            ;;
         *)
             echo '{"debug": "get-dependencies-ajax '"$tid"'"}'
             ;;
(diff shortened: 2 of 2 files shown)


openqa-investigate
```diff
@@ -27,6 +27,18 @@ jq_output_limit="${jq_output_limit:-15}"
...
+case "$host" in
+    openqa.opensuse.org)
+        investigation_gid="132"
+        ;;
+    openqa.suse.de)
+        investigation_gid="637"
+        ;;
+    *)
+        investigation_gid="0"
+        ;;
```
okurz, line 39: Wow, that's … not pretty and fragile. Can we not look up job groups by name?

okurz (changes requested):

openqa-investigate
perlpunk: And I would make it configurable (e.g. turn the feature on/off completely).
Maybe configuring the group id here would also be easier

okurz: Yes, you can use the default-fallback assignment but that does not help the problem that you hardcode instance specific group-ids here. Unless you find a way to use a job group name by convention I think you should go back to consider "plan A" from the ticket instead

perlpunk: My suggestion was to remove the instance specific assignments and only do `investigation_gid=${investigation_gid:-0}` at the top of the script. then it can be configured in openqa.ini.
A very simple change that doesn't cost us anything.

If it turns out that we want to use it, we can always think of finding the group by name later.

test/02-investigate.t
```diff
@@ -3,10 +3,7 @@
 source test/init
 bpan:source bashplus +err +fs +sym
 
-plan tests 88
-
-host=localhost
```
perlpunk, line 8: This is not necessary anymore, and you rather want test `investigation_gid=...`

perlpunk (changes requested):

test/02-investigate.t
b10n1k: Do u mean to revert the host?

perlpunk: yes, and the whole change. why would this be needed?
Instead test for setting `investigation_gid=...`

b10n1k: I put back the host+url. However I dont know where the url used. seems unused. and I am not sure why host is needed

test/02-investigate.t
```diff
@@ -87,6 +87,9 @@ client-get-job() {
         30032)
             echo '{"job": { "test": "vim:investigate:retry", "result": "passed", "settings": {"OPENQA_INVESTIGATE_ORIGIN": "3003"} } }'
             ;;
+	10035)
```
perlpunk, line 90: Please don't use tabs.

I think we have to change our checkstyle to also check `test/*.t`. However, shfmt doesn't seem to check for tabs...

perlpunk (changes requested):

test/02-investigate.t
perlpunk: Actually, shfmt does reformat tabs. but it doesn't see `test/*.t` as shell files

okurz (approved):

test/02-investigate.t
```diff
@@ -279,6 +291,9 @@ try force=true investigate 10031
 is "$rc" 0 'success when job is skipped (because of exclude_no_group and job w/o group)'
 has "$got" 'Job w/o job group, $exclude_no_group is set, skipping investigation'
 
+try investigation_gid=888 investigate 10035
+has "$got" "_GROUP_ID=888" "clone with correct group_id"
+
```
perlpunk, line 296: If I add
```
diag "$got"
```
here, I get
```
# https://localhost/tests/10035 _TRIGGER_JOB_DONE_HOOK=1 _GROUP_ID=888 BUILD= TEST+=:investigate:retry OPENQA_INVESTIGATE_ORIGIN=https://localhost/t10035
# curl (109 openqa-investigate): Error fetching (-L --user-agent openqa-investigate -sS https://localhost/tests/10035/investigation_ajax): curl: (7) Failed to connect to localhost port 443 after 0 ms: Couldn't connect to server
```
We should not try to make real requests in tests, even if it's localhost.
If `investigation_ajax` is not mocked yet, it should be.

perlpunk (changes requested):

test/02-investigate.t
okurz: @b10n1k one hint is to call applications with `unshare -n …` which takes away completel network access. If that behaves different then you know your application is relying on the network when it shouldn't

test/02-investigate.t
perlpunk: Fixed by #382

test/02-investigate.t
perlpunk: One problem here is that the request fails, but the test is fine with that. `unshare` will probably not change that.
When I call unsharei, I get an error, btw: (in a toolbox container)
```
% unshare -n prove test/02-investigate.t -v
unshare: unshare failed: Operation not permitted
```

perlpunk: > One problem here is that the request fails, but the test is fine with that

I think that part can actually be fixed. There is one place where we are missing a return value check... Will make a PR

perlpunk: So if you rebase after #383 and then also check `$rc` as part of your test, I assume it would fail

b10n1k: I mocked the curl for this particular test case. I dont see anymore `curl (109 openqa-investigate): Error fetching`

test/02-investigate.t
```diff
@@ -255,6 +267,31 @@ has "$got" 'WORKER_CLASS:vim=foo,duh,bar' "job assigns the WORKER_CLASS only to
...
+                "needles_log": "",
+                "test_diff_stat": "",
+                "test_log": "",
+                "testgiturl": "https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/"}'
+    echo -n $'\n'
+    echo -n 200
+}
+
+try investigation_gid=888 investigate 10035
+is "$rc" 0 'return code passed when clone with non-default _GROUP_ID'
```
perlpunk, line 291: A return code can't pass
```suggestion
is "$rc" 0 'success when clone with non-default _GROUP_ID'
```

test/02-investigate.t
```diff
@@ -255,6 +267,31 @@ has "$got" 'WORKER_CLASS:vim=foo,duh,bar' "job assigns the WORKER_CLASS only to
 try clone 10021 10030 last_good_tests_and_build:123 refspec
 has "$got" 'unidentified worker class in vars.json' "info shown in the name is WORKER_CLASS is empty"
 
+curl() {
```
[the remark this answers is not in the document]
b10n1k: Not sure how I should apply 72df157 here. 
I updated the mock tho. I think in general it should prevent others to run an actual curl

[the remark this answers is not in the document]
perlpunk: To spell it out:
In 72df157 you created `fetch-vars-json()` in `_common` to wrap the curl call to fetch `vars.json` . In this case it would be:
```
fetch-investigation-ajax() {
    local id=$1
    # put the curl here
}
```
Pretty much the same as in the other commit, right?
And then mock that function.

I think it's just cleaner and more maintainable.

And curl could just be mocked something like
```
curl() {
  exit 1
}
```
in the test, just to make sure it won't be called in any case.

But that's just how I would do it.

test/02-investigate.t
```diff
@@ -255,6 +267,34 @@ has "$got" 'WORKER_CLASS:vim=foo,duh,bar' "job assigns the WORKER_CLASS only to
...
+    if [[ "$@" =~ investigation_ajax ]]; then
+        echo -n '{"diff_packages_to_last_good": "Diff of packages not available",
+              "diff_sut_packages_to_last_good": "Diff of packages not available",
```
perlpunk, line 273: Are all these lines necessary?

b10n1k: no all of them. But I want to represent the $investigation as it is expected.

test/02-investigate.t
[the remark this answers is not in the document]
b10n1k: I never know how far I should go with the refactoring. if i am not wrong the `fetch-investigation-ajax` is not implemented. So I would have to have additional changes. and lets be honest. the most changes I do the most it fires back to me. and I dont have the confidence for doing things as expected. So I never think that far. Unless I know I want to change something is not happening if I am not sure I need it. So it is not how it should or anyone would do it. I avoid to make changes on the code which already provided and instead I try to just provide the minimum on the subject. I am saying this because I spend a lot of time understanding the code and the implications of each change (which might sound simple sometimes but still never know... ) I feel there are expectations which makes the review long. 
Said that not because I am against the proposal, but I feel I had to clarify this as usually I am expecting to provide code as others would do.

test/02-investigate.t
b10n1k: @perlpunk can we close this?

okurz: No, please only simulate the parts necessary for the specific tests to make tests robust and easier to understand. There is no good benefit to simulate everything in every test

okurz (changes requested):

okurz (approved):

test/02-investigate.t
perlpunk: "success when clon*ing*" but ok

perlpunk (approved):
apparently noone else except @okurz and me wanted to review, so I'll just approve

test/02-investigate.t
perlpunk: When the question is "is this necessary?" and the answer is no, then it's likely that it should be removed :)

PR #384, opened 2025-03-24
Create simultaneous submission of the packages to submit_targets

b10n1k:
- Refactor factory_request to take extra variables and make the query more abstract and modular. rename it to `get_obs_sr_id` to represent clear its purpose
- Keep the submit_target (but rename it `factory_target`) open for expansion but likely this is not need it and it can be a constant. The variable is used in all the phases before the update_package. But its logic is changed a bit as there is a new variable `targets` representing the final submission targets.
- Encapsulate logic in a main and break it into smaller logical functions
- Capture and report failed packages

The automatic submission is designed to sumbmit to the Factory using staging repo. However after that there is no need of this. The packages can be submitted to any other project from factory once they are accepted in there. Once the `update_package` is called which used to perform the submit request to factory, now it will loop through a list of `$targets` and create a request (aka osc sr).

https://progress.opensuse.org/issues/127037

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -7,7 +7,7 @@ src_project=${src_project:-devel:openQA}
 dst_project=${dst_project:-${src_project}:tested}
 staging_project=${staging_project:-${src_project}:testing}
 submit_target="${submit_target:-"openSUSE:Factory"}"
-submit_target_escaped=$(echo -n "$submit_target" | sed -e 's|\:|_|g') # avoid, e.g. "repoid 'openSUSE:Factory' is illegal"
+submit_target_extra="${submit_target_extra:-}"
 dry_run="${dry_run:-"0"}"
 osc_poll_interval="${osc_poll_interval:-2}"
 osc_build_start_poll_tries="${osc_build_start_poll_tries:-30}"
@@ -18,11 +18,23 @@ XMLSTARLET=$(command -v xmlstarlet || true)
 # shellcheck source=/dev/null
 . "$(dirname "$0")"/_common
 
-factory_request() {
-    local package=$1
+IFS=',' read -ra targets <<< "$submit_target,$submit_target_extra"
+failed_packages=()
+
+encode_variable() {
+    #  https://stackoverflow.com/a/298258
+    perl -MURI::Escape -e 'print uri_escape($ARGV[0])' "$1"
+}
+
+get_obs_sr_id() {
+    local target=$1
+    local dst_project=$2
+    target=$(encode_variable "$target")
+    dst_project=$(encode_variable "$dst_project")
+    local package=$3
     # writing xpath in url encoding is not for beginners, so don't stare at it too long :)
     local states='(state%2F%40name%3D%27declined%27%20or%20state%2F%40name%3D%27new%27%20or%20state%2F%40name%3D%27review%27)'
-    local actions="(action%2Ftarget%2F%40project%3D%27openSUSE%3AFactory%27%20and%20action%2Fsource%2F%40project%3D%27devel%3AopenQA%3Atested%27%20and%20action%2Ftarget%2F%40package%3D%27$package%27)"
+    local actions="(action%2Ftarget%2F%40project%3D%27${target}%27%20and%20action%2Fsource%2F%40project%3D%27${dst_project}%27%20and%20action%2Ftarget%2F%40package%3D%27$package%27)"
     $osc api "/search/request/id?match=$states%20and%20$actions" | grep 'request id' | sed -e 's,.*id=.\([0-9]*\)[^[0-9]*$,\1,' | head -n 1
 }
 
@@ -31,7 +43,9 @@ reenable_buildtime_services() {
 }
 
 wait_for_package_build() {
-    local osc_query="https://api.opensuse.org/public/build/$dst_project/_result?repository=$submit_target_escaped&package=$package"
+    local target=$1
+    target_escaped=$(echo -n "$target" | sed -e 's|\:|_|g') # avoid, e.g. "repoid 'openSUSE:Factory' is illegal"
+    local osc_query="https://api.opensuse.org/public/build/$dst_project/_result?repository=$target_escaped&package=$package"
     local wip_states='\(unknown\|blocked\|scheduled\|dispatching\|building\|signing\|finished\)'
     local bad_states='\(failed\|unresolvable\|broken\)'
     local attempts="$osc_build_start_poll_tries"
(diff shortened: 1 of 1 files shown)


os-autoinst-obs-auto-submit
```diff
@@ -6,8 +6,8 @@ export LC_ALL=C
 src_project=${src_project:-devel:openQA}
 dst_project=${dst_project:-${src_project}:tested}
 staging_project=${staging_project:-${src_project}:testing}
-submit_target="${submit_target:-"openSUSE:Factory"}"
-submit_target_escaped=$(echo -n "$submit_target" | sed -e 's|\:|_|g') # avoid, e.g. "repoid 'openSUSE:Factory' is illegal"
+factory_target="${factory_target:-"openSUSE:Factory"}"
+factory_target_escaped=$(echo -n "$factory_target" | sed -e 's|\:|_|g') # avoid, e.g. "repoid 'openSUSE:Factory' is illegal"
```
okurz, line 10: Please keep the original names intact. And then what you add is "secondary" or "additional" targets

okurz (changes requested):

os-autoinst-obs-auto-submit
b10n1k: I will but it might be confusing. it was for me at least to follow the logic.

os-autoinst-obs-auto-submit
```diff
@@ -78,22 +92,25 @@ update_package() {
...
-    wait_for_package_build
-    cmd="$osc sr"
-    req=$(factory_request "$package" || :)
-    # TODO: check if it's a different revision than HEAD
-    if test -n "$req"; then
-        cmd="$cmd -s $req"
-    fi
-    $cmd -m "Update to ${version}" "$submit_target"
+    for target in "${targets[@]}"; do
+        echo -n "##### $package($version): try to submit to $target #####"
```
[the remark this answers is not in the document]
b10n1k: removed

b10n1k:
forked from https://github.com/os-autoinst/scripts/pull/380

os-autoinst-obs-auto-submit
```diff
@@ -7,7 +7,7 @@ src_project=${src_project:-devel:openQA}
 dst_project=${dst_project:-${src_project}:tested}
 staging_project=${staging_project:-${src_project}:testing}
 submit_target="${submit_target:-"openSUSE:Factory"}"
-submit_target_escaped=$(echo -n "$submit_target" | sed -e 's|\:|_|g') # avoid, e.g. "repoid 'openSUSE:Factory' is illegal"
+submit_target_extra="${submit_target_extra:-"$submit_target"}"
```
okurz, line 10: The name implies that specified here would be "extra" to "submit_target" but you rely on the default value being passed as submit_target. How about using a default value of empty string and in https://github.com/os-autoinst/scripts/pull/384/files#diff-e9e4a2f7fd0feb7816f56d2fcd9ad02870fc9344ea52dfc275699c1a7d2cbff8R21 you read from both submit_target and submit_target_extra

b10n1k: sounds doable and efficient too. should I go with another name as well? maybe `submit_target_list`?

b10n1k: Done. and variable renamed to `submit_target_supplemental`

okurz:
s/sumbmit/submit/

os-autoinst-obs-auto-submit
```diff
@@ -7,7 +7,7 @@ src_project=${src_project:-devel:openQA}
 dst_project=${dst_project:-${src_project}:tested}
 staging_project=${staging_project:-${src_project}:testing}
 submit_target="${submit_target:-"openSUSE:Factory"}"
-submit_target_escaped=$(echo -n "$submit_target" | sed -e 's|\:|_|g') # avoid, e.g. "repoid 'openSUSE:Factory' is illegal"
+submit_target_supplemental="${submit_target_list:-}"
```
okurz, line 10: Why not just

```suggestion
submit_target_extra="${submit_target_extra:-}"
```

okurz: The "extra" was fine, it just means "extra" to the base target which is already there

okurz (approved):

os-autoinst-obs-auto-submit
```diff
@@ -173,37 +194,62 @@ handle_auto_submit() {
...
+    has_pending_submission || exit 0
+    (
+        cd "$TMPDIR"
+        rc=0
+        auto_submit_packages=$(get_project_packages)
+        for package in $auto_submit_packages; do
+            handle_auto_submit "$package" || rc=$?
+        done
+        # delete package from staging project
+        [[ $(is_staging_specified) ]] \
```
Martchus, line 234: `is_staging_specified` prints either 0 or 1. Both are considered true by `[[ … ]]`. So this isn't going to work. I suggest you change `is_staging_specified` to `return` a value instead of `echo`ing a value and then just invoke it (like you did it with `handle_pending_submission`).

os-autoinst-obs-auto-submit
```diff
@@ -173,37 +194,62 @@ handle_auto_submit() {
...
+            && $osc rdelete -m "Cleaning up $package from $staging_project for next submission" "$staging_project" "$package"
+        if [[ ${#failed_packages[@]} -gt 0 ]]; then
+            echo "Failed packages:"
+            for item in "${failed_packages[@]}"; do
+                package="${item%:*}"
+                exit_status="${item#*:}"
```
Martchus, line 240: No quotes are required for assignments like these:
```suggestion
                package=${item%:*}
                exit_status=${item#*:}
```
So at least in new code I'd try to avoid needles quotes.

b10n1k:
> The commit message is full of typos and mentions variables under a different name than what's used in the actual change.

Also fixed as possible as it can

os-autoinst-obs-auto-submit
```diff
@@ -173,37 +194,61 @@ handle_auto_submit() {
...
-if [[ -e job_post_skip_submission ]]; then
-    echo "Skipping submission, reason: "
-    cat job_post_skip_submission
-    exit 0
-fi
+is_staging_specified() {
+    [[ -n $staging_project && $staging_project != none && $staging_project != "$src_project" ]] && {
+        src_project="$staging_project"
+        return 0
+    } || return 1
```
Martchus, line 215: Using a plain-old if-else would be much more readable:
```suggestion
    if [[ -n $staging_project && $staging_project != none && $staging_project != "$src_project" ]]; then
        src_project=$staging_project
        return 0
    else
        return 1
    fi
```

Martchus (changes requested):

Martchus (approved):

os-autoinst-obs-auto-submit
```diff
@@ -173,37 +194,63 @@ handle_auto_submit() {
...
+}
 
-if [[ -e job_post_skip_submission ]]; then
-    echo "Skipping submission, reason: "
-    cat job_post_skip_submission
-    exit 0
-fi
+is_staging_specified() {
+    if [[ -n $staging_project && $staging_project != none && $staging_project != "$src_project" ]]; then
+        src_project="$staging_project"
```
baierjan, line 213: I think this is causing a regression, overriding the `src_project` now does happen only for the clean-up job (`osc rdelete`) and not before `handle_auto_submit` anymore so the update is always from the devel project and not from the staging project. Am I correct?

b10n1k: https://github.com/os-autoinst/scripts/pull/403

PR #386, opened 2025-03-28
Remove filter-dehydrated-cron-output

b10n1k:
script introduced in 9674513c004d34ef942fee8f96f3f992224cb5a2 is not anymore of any use. The purpose was to filter error messages which cron will use to send email to o3-admins, preventing spam by any other logs.

the goal on https://progress.opensuse.org/issues/179149 was to handle request errors. As it turned out there is already a systemd with retry. So the crontab job was removed. and the filter-dehydrated-cron-output remained `orphan`. However there should be still a notification about likely errors from dehydrated. Which can be solved on the configuration of the dehydrated script itself and specifically on hook.sh.


perlpunk (approved):
Yes, it was used to not produce any output if dehydrated just checked and did not have to renew anything, so we would only get an email if it renewed the certificate.

okurz (approved):

PR #390, opened 2025-04-08
Use latest project version for submit_target_extra

b10n1k:
Try to find through osc the latest version of a project inn order to determine
the final destination of the submit request for `submit_target_extra`.

related issue: https://progress.opensuse.org/issues/127037

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -7,7 +7,7 @@ src_project=${src_project:-devel:openQA}
 dst_project=${dst_project:-${src_project}:tested}
 staging_project=${staging_project:-${src_project}:testing}
 submit_target="${submit_target:-"openSUSE:Factory"}"
-submit_target_extra="${submit_target_extra:-"openSUSE:Backports:SLE-15-SP6:Update"}"
+submit_target_extra_project="${submit_target_extra_project:-"openSUSE:Backports"}"
 dry_run="${dry_run:-"0"}"
 osc_poll_interval="${osc_poll_interval:-2}"
 osc_build_start_poll_tries="${osc_build_start_poll_tries:-30}"
@@ -16,8 +16,20 @@ XMLSTARLET=$(command -v xmlstarlet || true)
 [[ -n $XMLSTARLET ]] || (echo "Need xmlstarlet" && exit 1)
 
 # shellcheck source=/dev/null
-. "$(dirname "$0")"/_common
+. "$(dirname "${BASH_SOURCE[0]}")"/_common
 
+get_submit_target_extra_latest_version() {
+    # awk iterates through all the fields in the match to avoid fixed position
+    # and it should match with the part which starts with `submit_target_extra_project`+`:`
+    latest_submit_target_extra_version=$(osc dists | grep "$submit_target_extra_project" | awk -v prj="$submit_target_extra_project" '{for (i=1; i<=NF; i++) if ($i ~ "^" prj ":") print $i}' | sort --version-sort | tail -n 1)
+    if [[ -z $latest_submit_target_extra_version ]]; then
+        echo "Project not found." && return 1
+    fi
+    # Notice that the :Update doesn't come from osc output
+    echo "$latest_submit_target_extra_version:Update"
+}
+
+submit_target_extra="{submit_target_extra:-get_submit_target_extra_latest_version}"
 IFS=',' read -ra targets <<< "$submit_target,$submit_target_extra"
 failed_packages=()
 
@@ -96,6 +108,7 @@ update_package() {
     wait_for_package_build "$submit_target"
     for target in "${targets[@]}"; do
         cmd="$osc sr"
+        echo "## Ready to submit $package to $target ##"
         req=$(get_obs_sr_id "$target" "$dst_project" "$package" || :)
         # TODO: check if it's a different revision than HEAD
         if test -n "$req"; then
--- a/test/08-os-autoinst-obs-auto-submit.t
+++ b/test/08-os-autoinst-obs-auto-submit.t
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+source test/init
+bpan:source bashplus +err +fs +sym
+
+plan tests 6
+source os-autoinst-obs-auto-submit
+
+submit_target_extra_project="openSUSE:Backports"
+osc_dists_output='
+openSUSE Backports for SLE 15 SP6 openSUSE:Backports:SLE-15-SP6 15.6 standard
+openSUSE Backports for SLE 15 SP5 openSUSE:Backports:SLE-15-SP5 15.5 standard
+openSUSE Backports for SLE 15 SP6:Update  openSUSE:Backports:SLE-15-SP6:Update:GA 1.0 standard 
+openSUSE Backports for SLE 15 SP4 openSUSE:Backports:SLE-15-SP4 15.4 standard'
+
+osc() {
+    if [[ "$1" == "dists" ]]; then
+        echo "$osc_dists_output"
+        return 0
+    else
+        echo "Not mocked 'osc' call:" "$@" >&2
+        return 1
+    fi
+}
+
+try get_submit_target_extra_latest_version
+is "$rc" 0 "rc should be 0"
+is "$got" "openSUSE:Backports:SLE-15-SP6:Update" "the latest version is used from the osc dists output"
+
+osc_dists_output=''
+try get_submit_target_extra_latest_version
+is "$rc" 1 "rc should be 1"
+is "$got" "Project not found." "Error if project no match any from osc dists output"
+
+osc_dists_output='
+openSUSE Backports for SLE 15 SP6 openSUSE:Backports:SLE-15-SP6 15.6 standard
+openSUSE Backports for SLE 15 SP7 openSUSE:Backports:SLE-17-SP5 17.5 standard
+openSUSE Backports for SLE 15 SP6:Update  openSUSE:Backports:SLE-15-SP6:Update:GA 1.0 standard
+openSUSE Backports for SLE 15 SP4 openSUSE:Backports:SLE-15-SP4 15.4 standard'
+
+try get_submit_target_extra_latest_version
+is "$rc" 0 "rc should be 0"
+is "$got" "openSUSE:Backports:SLE-17-SP5:Update" "the latest version is used from the osc dists output comparing the major version"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
           python --version
           pytest --version
       - run: |
-          sudo apt-get install cpanminus html-xml-utils
+          sudo apt-get install cpanminus html-xml-utils xmlstarlet
           sudo cpanm -n -M https://cpan.metacpan.org --installdeps .
       - name: unit- and integration tests
         run: |


r-richardson:
Do we know the version of any project is always in the major.minor format ? If not maybe it would be good to loop over the individual version components for comparison

b10n1k:
> Do we know the version of any project is always in the major.minor format ? If not maybe it would be good to loop over the individual version components for comparison

no this is not the case. its not all in this format. I rely on the fact that Leap is shown as that. I am not sure if it is worth one-for-all comparison. but it depends from the requirements i guess

os-autoinst-obs-auto-submit
```diff
@@ -194,6 +195,45 @@ handle_auto_submit() {
...
+            latest_version="${current_prj}:${version}"
+            latest_major_minor_versions=("${major_minor_versions[@]}")
+        elif [[ "${major_minor_versions[0]}" -eq "${latest_major_minor_versions[0]}" ]]; then
+            # Compare minor vers if majors are equal
+            if [[ "${major_minor_versions[1]}" -gt "${latest_major_minor_versions[1]}" ]]; then
+                latest_version="${current_prj}:${version}"
+                latest_major_minor_versions=("${major_minor_versions[@]}")
+            fi
+        fi
+    done
```
Martchus, line 231: Use `sort --version-sort` instead of this.

If you must compare individual versions it is probably better to use rpm (like this `rpm --eval "%{lua:print(rpm.vercmp('1.2-1', '2.0-1'))}"`) but I guess `sort --version-sort` is good enough anyway.

Martchus (changes requested):

okurz:
Can you provide examples of what this will yield currently? I assume what we need for https://progress.opensuse.org/issues/127037 would be currently https://build.opensuse.org/project/show/openSUSE:Backports:SLE-15-SP6:Update ?

os-autoinst-obs-auto-submit
b10n1k: Nice. actually super nice. But I will like to try. I didnt know about `sort --version-sort` for example.

b10n1k:
> Can you provide examples of what this will yield currently? I assume what we need for https://progress.opensuse.org/issues/127037 would be currently https://build.opensuse.org/project/show/openSUSE:Backports:SLE-15-SP6:Update ?

I will provide a very simplified answer and we can discuss "in person" afterwards...
the default value of submit_target_extra is "openSUSE:Leap:15.6"
And currently this is what it will return `openSUSE:Leap:15.6`.
it will return *<project>:<version>*

b10n1k:
> https://progress.opensuse.org/issues/127037

I think I see what do you mean. and I need to fix it because it wont work as it is. I will try to use the suggestion from Marius and fix it in the next commit.

b10n1k:
> ok, so what happens if you try to create a SR to openSUSE:Leap:15.6 then?

I do actually tried to run the script without the dry_run. 
i found that I dont have permission to submit in openQA:devel, so I get many errors. But I let it keep going. I am not sure why but I see 

```
Waiting while openQA is in progress
  <summary>unknown repository 'openSUSE_Leap_15.6'</summary>
  <details>404 unknown repository 'openSUSE_Leap_15.6'</details>
```

os-autoinst-obs-auto-submit
```diff
@@ -18,6 +18,16 @@ XMLSTARLET=$(command -v xmlstarlet || true)
...
 
+# extract :Update
+update_prj="${submit_target_extra##*:}"
+# extract openSUSE:Backports:SLE-15-SP6
+current_prj="${submit_target_extra%:*}"
+# extact the actual version to compare
+current_prj="${current_prj%:*}"
+# awk iterates through all the fields in the match to avoid fixed position
+latest_submit_target_extra_version=$(osc dists | grep "$current_prj" | awk -v prj="$current_prj" '{for (i=1; i<=NF; i++) if ($i ~ "^" prj ":") print $i}' | sort --version-sort | tail -n 1)
+submit_target_extra="$latest_submit_target_extra_version:$update_prj"
```
kalikiana, line 29: This is not very straight-forward to read and feels like it will break easily. Any chance we can unit test this with a couple of examples?

b10n1k: This can break in many ways and it depends on the $submit_target_extra
I tried to generalize it but at the end the main goal is to have an additional SR to the Backports. There could be some checks or some mapping (some things which I considered when I wrote the code)... lets make it work and then see where it can improve?!

b10n1k: note that this still doesnt work

kalikiana: As you seem to agree with me, are you planning to add a test for this? 🤔

b10n1k: Honestly I havent thought about that. There is no test in the repo for it and I am not sure how easy will be to test against OBS. 
When I said `There could be some checks` I meant some conditions in the script to catch some cases.

b10n1k: But if you insists I will try to add one or two assertions for the variable. sounds good? (what I dont know is how urgent is this, and whether can be another PR, in order to move on with this one. This is already some time we try to complete the task and I would like to see some progress)

kalikiana: The fact that we have no tests right now is not a very helpful argument to my mind. Even writing this code myself I would prepare tests first to confirm what it does.

I'm not talking about testing against production infra btw. Something like a few examples of possible `osc dists` with one or two examples of a project config we use.

b10n1k: @kalikiana I am sure this will start a new iteration of changes. But I did add a test but... 
not sure you will agree with the approach and I dont know if those lines should be separate function in the script. And Furthermore I dont think it covers all you expect to see. I have done assumptions for now. and I couldnt think how the script would look like in the near future. so it is kinda poor approach🤞🏾

b10n1k:
💡Add a check in the case where `osc dists` doesnt found the product
💡Write some tests

test/08-os-autoinst-obs-auto-submit.t
```diff
@@ -0,0 +1,43 @@
...
+openSUSE Backports for SLE 15 SP6 openSUSE:Backports:SLE-15-SP6 15.6 standard
+openSUSE Backports for SLE 15 SP5 openSUSE:Backports:SLE-15-SP5 15.5 standard
+openSUSE Backports for SLE 15 SP6:Update  openSUSE:Backports:SLE-15-SP6:Update:GA 1.0 standard 
+openSUSE Backports for SLE 15 SP4 openSUSE:Backports:SLE-15-SP4 15.4 standard'
+
+check_submit_target_extra_expansion() {
+    update_prj="${submit_target_extra##*:}"
+    current_prj="${submit_target_extra%:*}"
+    current_prj="${current_prj%:*}"
+    latest_submit_target_extra_version=$(echo "$osc_dists_output" | grep "$current_prj" | awk -v prj="$current_prj" '{for (i=1; i<=NF; i++) if ($i ~ "^" prj ":") print $i}' | sort --version-sort | tail -n 1)
```
Martchus, line 20: It seems terribly wrong that this test duplicates this rather complicated line from the code that's to be tested (even though this is stated in the commit message).

Martchus (dismissed):
I guess at this point I miss the bigger picture. I'm wondering how this helps towards the goal of submitting packages to package hub and/or Leap. Aren't there other problems to tackle first? (The version could actually just be hard-coded for the time being. Having to increment it manually on a new Leap release seems not like a big deal.)

b10n1k:
> I guess at this point I miss the bigger picture. I'm wondering how this helps towards the goal of submitting packages to package hub and/or Leap. Aren't there other problems to tackle first? (The version could actually just be hard-coded for the time being. Having to increment it manually on a new Leap release seems not like a big deal.)

https://progress.opensuse.org/issues/127037?issue_count=4&issue_position=2&next_issue_id=108980&prev_issue_id=179296#note-46 is the reference to the request for automated submission to the latest version.

One problem with the big picture is that this solution is not the favorite solution. As it is, it creates a parallel SR. one to factory one to Backport. The other way, is create SR to factory and then from there to Backport. But this, I think, will require more significant changes to the script. Especially if the idea is to not mess with jenkins. My goal is to see packages published in Backport for now, see it working and then reconsider what needs to change.

b10n1k:
💡Use a function to get the latest obs project version
💡Update the test to use the underline function

However the CI fails to run the test. I tried local and each one passes.!!

os-autoinst-obs-auto-submit
```diff
@@ -16,7 +16,24 @@ XMLSTARLET=$(command -v xmlstarlet || true)
...
+    current_prj="${current_prj%:*}"
+    # awk iterates through all the fields in the match to avoid fixed position
+    latest_submit_target_extra_version=$(osc dists | grep "$current_prj" | awk -v prj="$current_prj" '{for (i=1; i<=NF; i++) if ($i ~ "^" prj ":") print $i}' | sort --version-sort | tail -n 1)
+    if [[ -z $latest_submit_target_extra_version ]]; then
+        echo "Project not found." && return 1
+    fi
+    echo "$latest_submit_target_extra_version:$update_prj"
+}
+
+submit_target_extra=check_or_update_submit_target_extra_version
```
okurz, line 36: In line 10 a default is set allowing callers to set a custom value but here you overwrite it again. That's not intuitive. I think you should define a project root as default like `submit_target_extra_root=${submit_target_extra_root:-"openSUSE:Backports:"}"`, use that and then only as default set `submit_target_extra` to the output of check_or_update_submit_target_extra_version

okurz (changes requested):

b10n1k:
💡add xmlstarlet missing dependency to make the new test run (thanks @perlpunk )
💡Replace the `submit_target_extra` to show only the project to submit to. 
💡no change on the test other than the adjustments to reflect the function name and the expected variable.

b10n1k:
💡just updated the commit message

os-autoinst-obs-auto-submit
okurz: Please read my suggestion again. You now still overwrite any potentially user specific config settings

b10n1k: I would say that it updates the user input to actually use the latest version. it can be either ways. you can check again

b10n1k:
💡Make the submit_target_extra more intuitive, separating the variable which is the lookup project (submit_target_extra_project) and the actual destination target for the Backport submission (submit_target_extra).

os-autoinst-obs-auto-submit
okurz: It does not make sense to make submit_target_extra a user specifiable variable but then overwrite it

okurz (changes requested):

os-autoinst-obs-auto-submit
kalikiana: I agree user-specified variables shouldn't be re-used. Even if it works it's confusing to follow.

How about `submit_target_extra_latest_version`

b10n1k: are you talking about `submit_target_extra` or `submit_target_extra_project`?

kalikiana: Well, this was about `submit_target_extra=check_or_update_submit_target_extra_version`.

I guess now that you are renaming the variable and only using it internally that no longer applies.


Please update the commit/PR description to clarify, though, since that still talks about the old name 🤔 Something like **Pull in latest version of projects with submit_target_extra_project** if that is what it does?

okurz: I guess you still didn't read or understand my suggestion from https://github.com/os-autoinst/scripts/pull/390#discussion_r2073407164

Please do

```bash
submit_target_extra="${submit_target_extra:-check_or_update_submit_target_extra_version}"
```

b10n1k:
💡as discussed: latest commit provides env variable `submit_target_extra` which defaults to the output of the `get_submit_target_extra_latest_version`. which consecutively depends on the `submit_target_extra_project`, defaults to `openSUSE:Backports`.
That means that the default is the latest `openSUSE:Backports:$VERSION:Update` and alternatively the full destination via `submit_target_extra`

kalikiana (approved):

os-autoinst-obs-auto-submit
```diff
@@ -16,8 +16,20 @@ XMLSTARLET=$(command -v xmlstarlet || true)
...
+    # and it should match with the part which starts with `submit_target_extra_project`+`:`
+    latest_submit_target_extra_version=$(osc dists | grep "$submit_target_extra_project" | awk -v prj="$submit_target_extra_project" '{for (i=1; i<=NF; i++) if ($i ~ "^" prj ":") print $i}' | sort --version-sort | tail -n 1)
+    if [[ -z $latest_submit_target_extra_version ]]; then
+        echo "Project not found." && return 1
+    fi
+    # Notice that the :Update doesn't come from osc output
+    echo "$latest_submit_target_extra_version:Update"
+}
+
+submit_target_extra="{submit_target_extra:-get_submit_target_extra_latest_version}"
```
perlpunk, line 32: If I execute this line in bash, I get
```
% submit_target_extra="{submit_target_extra:-get_submit_target_extra_latest_version}"
% echo $submit_target_extra 
{submit_target_extra:-get_submit_target_extra_latest_version}
```
see also the latest jenkins pipeline http://jenkins.qa.suse.de/job/submit-openQA-TW-to-oS_Fctry/4002/console

PR #392, opened 2025-04-23
Drop extra submission target

kalikiana:
The repo doesn't exist and it's not clear if we want this.

See: https://progress.opensuse.org/issues/180866
Also: https://progress.opensuse.org/issues/127037


kalikiana:
Effectively going to the state from before #388 so we can discuss the plan without on-going failing pipelines.

perlpunk (approved):

okurz:
I don't understand why you removed that. Submissions have been recently created and they look fine so just removing that doesn't sound correct

okurz:
> > I don't understand why you removed that. Submissions have been recently created and they look fine so just removing that doesn't sound correct
> 
> Because it doesn't work:
> 
> ```
> Waiting while os-autoinst is in progress
>   <summary>unknown repository 'openSUSE_Backports_SLE-15-SP6_Update'</summary>
>   <details>404 unknown repository 'openSUSE_Backports_SLE-15-SP6_Update'</details>
> ```

That is just some log output and not a proof that "it doesn't work". How was https://build.opensuse.org/requests/1271581 created then? I did not personally create that SR

PR #393, opened 2025-04-28
Add back default Backports target for submit_target_extra

b10n1k:
Essentially revert the revert by
https://github.com/os-autoinst/scripts/pull/392/files and apply the fix for the jenkins[0].
The problem had been spotted and the fix was already applied in https://github.com/os-autoinst/scripts/pull/390 but it didnt make it to the production before the revert.

[0] https://progress.opensuse.org/issues/180866

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -7,7 +7,7 @@ src_project=${src_project:-devel:openQA}
 dst_project=${dst_project:-${src_project}:tested}
 staging_project=${staging_project:-${src_project}:testing}
 submit_target="${submit_target:-"openSUSE:Factory"}"
-submit_target_extra="${submit_target_extra:-}"
+submit_target_extra="${submit_target_extra:-"openSUSE:Backports:SLE-15-SP6:Update"}"
 dry_run="${dry_run:-"0"}"
 osc_poll_interval="${osc_poll_interval:-2}"
 osc_build_start_poll_tries="${osc_build_start_poll_tries:-30}"
@@ -93,8 +93,8 @@ update_package() {
     sed -i -e 's/^  \* $/  * Update to latest openQA version/' ./*.changes
     $osc ci -m "Offline generation of ${version}" $ci_args
     rc=0
+    wait_for_package_build "$submit_target"
     for target in "${targets[@]}"; do
-        wait_for_package_build "$target"
         cmd="$osc sr"
         req=$(get_obs_sr_id "$target" "$dst_project" "$package" || :)
         # TODO: check if it's a different revision than HEAD


os-autoinst-obs-auto-submit
```diff
@@ -93,9 +93,9 @@ update_package() {
     sed -i -e 's/^  \* $/  * Update to latest openQA version/' ./*.changes
     $osc ci -m "Offline generation of ${version}" $ci_args
     rc=0
+    wait_for_package_build "$target"
```
perlpunk, line 96: How is that supposed to work outside the loop below?

os-autoinst-obs-auto-submit
```diff
@@ -93,9 +93,9 @@ update_package() {
...
+    cmd="$osc sr"
```
perlpunk, line 97: IIUC, in line 102 `cmd="$cmd -s $req"`, so with every loop it gets extended.
That's why this needs to stay in the loop I think.

os-autoinst-obs-auto-submit
b10n1k: ops. I forgot that. because I was testing it in another branch. will fix

os-autoinst-obs-auto-submit
b10n1k: Not sure I follow. whats extended? you mean `cmd` or `$req`. `$cmd` is always `osc sr` and $req indeed is updated with every loop, but it is already in that loop

okurz:
@kalikiana

os-autoinst-obs-auto-submit
```diff
@@ -7,7 +7,7 @@ src_project=${src_project:-devel:openQA}
 dst_project=${dst_project:-${src_project}:tested}
 staging_project=${staging_project:-${src_project}:testing}
 submit_target="${submit_target:-"openSUSE:Factory"}"
-submit_target_extra="${submit_target_extra:-}"
+submit_target_extra="${submit_target_extra:-"openSUSE:Backports:SLE-15-SP6:Update"}"
```
kalikiana, line 10: Did you try this manually?

I'm asking since we've had broken Jenkins jobs for multiple days related to the original change.

kalikiana: Or if it's not trivial to execute manually, I guess we just need to merge within a working day and consider reverting quickly.

b10n1k: I didnt try manual. I can say that I didnt see the underlying problem from jenkins and it was not expected to be seen because the part which causes it, it wont run against the Backports target anymore. ofc you can go also with the `we just need to merge within a working day and consider reverting quickly.` in any case. But I would say I am quite confident.

kalikiana: We know that it was failing repeatedly and there's no unit test coverage. Regardless of how that happened, that's the reason I asked 🙃

Let's merge it early on Friday then.

kalikiana (approved):

PR #396, opened 2025-04-29
openqa-trigger-bisect-jobs: Add help text

okurz:


openqa-trigger-bisect-jobs
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env python3
 
+"""
+Trigger openQA tests bisecting which pending operating system maintenance
+update, called "incident", might be thea reason for an openQA test failure.
```
perlpunk:
```suggestion
update, called "incident", might be the reason for an openQA test failure.
```

okurz: fixed

perlpunk (approved):

Martchus (approved):

PR #398, opened 2025-05-05
Remove investigation comment on job when clone fail

Wabri:
When the clone fails the script update the comment of the investigation with an alert and the motivation of the failure:

image

reference: https://progress.opensuse.org/issues/180218

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -374,7 +374,15 @@ investigate() {
     comment_id=$(sync_via_investigation_comment "$id" "$first_cluster_job_id") || rc=$?
     [[ $rc != 255 ]] && return $rc
 
-    out=$(trigger_jobs "$id" "${@:2}") || return $?
+    out=""
+    if ! out=$(trigger_jobs "$id" "${@:2}" 2>&1); then
+        local comment="Triggering investigation jobs failed: $out"
+        echoerr "Unexpected error encountered when trigger job $id: '$out'"
+        client-put-job-comment "$first_cluster_job_id" "$comment_id" "$comment"
+        return 1
+    fi
+
+    echo "$out"
     $verbose && echo "$0, id: '$id', out: '$out'"
     finalize_investigation_comment "$id" "$first_cluster_job_id" "$comment_id" "$out"
 }


Wabri:
Work locally:

```
host=localhost:1080 scheme=http exclude_no_group=false ./openqa-investigate 2
------- comment_id: 11
openqa-clone-job (84 ./openqa-investigate): (openqa-clone-job --json-output --skip-chained-deps --max-depth 0 --parental-inheritance --within-instance http://localhost:1080/tests/2 _TRIGGER_JOB_DONE_HOOK=1 _GROUP_ID=0 BUILD= TEST+=:investigate:retry PUBLISH_HDD_1=none OPENQA_INVESTIGATE_ORIGIN=http://localhost:1080/t2) stderr: >>>Environment contains multiple differing definitions for 'host'.<<<
openqa-clone-job (84 ./openqa-investigate): (openqa-clone-job --json-output --skip-chained-deps --max-depth 0 --parental-inheritance --within-instance http://localhost:1080/tests/2 _TRIGGER_JOB_DONE_HOOK=1 _GROUP_ID=0 BUILD= TEST+=:investigate:retry PUBLISH_HDD_1=none OPENQA_INVESTIGATE_ORIGIN=http://localhost:1080/t2) rc: 22 >>><<<
------- out:
{"id":11}
Unable to trigger investigation jobs for job 2
```

Not tested yet on osd.

openqa-investigate
```diff
@@ -374,7 +374,17 @@ investigate() {
     comment_id=$(sync_via_investigation_comment "$id" "$first_cluster_job_id") || rc=$?
     [[ $rc != 255 ]] && return $rc
 
-    out=$(trigger_jobs "$id" "${@:2}") || return $?
+    echo "------- comment_id: $comment_id"
+    out=""
+    if ! out=$(trigger_jobs "$id" "${@:2}"); then
+        echo "------- out: $out"
+        client-delete-job-comment "$first_cluster_job_id" "$comment_id"
+        echoerr "Unable to trigger investigation jobs for job $id"
```
okurz, line 382: What should the administrator receiving this message do about the error? I don't think this is actionable

openqa-investigate
```diff
@@ -374,7 +374,17 @@ investigate() {
...
+        return 1
+    fi
+
+    echo "------- out: $out"
```
okurz, line 386: I assume this is manual debugging left-over?

openqa-investigate
```diff
@@ -374,7 +374,17 @@ investigate() {
...
     [[ $rc != 255 ]] && return $rc
 
-    out=$(trigger_jobs "$id" "${@:2}") || return $?
+    echo "------- comment_id: $comment_id"
+    out=""
+    if ! out=$(trigger_jobs "$id" "${@:2}"); then
+        echo "------- out: $out"
+        client-delete-job-comment "$first_cluster_job_id" "$comment_id"
+        echoerr "Unable to trigger investigation jobs for job $id"
+        return 1
```
okurz, line 383: If there is no comment written how does to cover AC3 from https://progress.opensuse.org/issues/180218#Acceptance-criteria "Given openqa-investigate is started on a job When it aborts (e.g. because the clone failed) Then there is exactly one comment informing that it aborted (and possibly why)"?

okurz (changes requested):

openqa-investigate
Wabri: Yep, I forgot to mark the pr as draft, I'll remove those before presenting the final version.

openqa-investigate
Wabri: You are completely right, maybe something like `Unable to complete the investigation on job $id due to clone failure, check for missing repositories or invalid configurations`?

openqa-investigate
Wabri: It could be an improvement to change the comment description, I'm working on it!

Wabri: I saw the suggestion https://progress.opensuse.org/issues/180218#Suggestions but I'm not sure how to incorporate it

openqa-investigate
okurz: Again, what should the *administrator* receiving this message do about the error? You are part of the admin team for o3 and osd so think what *you* would be able to deduce from this message if anything. I think this message is directed more towards test reviewers but they won't see it here. That's why we have AC3

openqa-investigate
```diff
@@ -374,7 +374,13 @@ investigate() {
     comment_id=$(sync_via_investigation_comment "$id" "$first_cluster_job_id") || rc=$?
     [[ $rc != 255 ]] && return $rc
 
-    out=$(trigger_jobs "$id" "${@:2}") || return $?
+    out=""
```
b10n1k, line 377: This could be 
`out=${$(trigger_jobs "$id" "${@:2}"):-""}`

I think it would simplify the logic a bit?

openqa-investigate
Martchus: Not every log line needs to be actionable by the administrator. The one might still be useful when debugging the script itself. 
Of course it would also make sense to write this message as openQA comment. 
I find the dashes at the beginning of the messages a bit excessive.

openqa-investigate
Wabri: I find this less readable, I prefer to split the assignation to be more clear on the intent and the expectation.

openqa-investigate
```diff
@@ -374,7 +374,22 @@ investigate() {
...
+        if [[ $comment_id != "$first_comment_id" ]]; then
+            client-delete-job-comment "$first_cluster_job_id" "$comment_id"
+        fi
+        local comment="Investigation aborted for job $id.
+
+Reasons: clone failure.
+
+Suggestions: check for missing repositories or invalid configurations, investigate the job manually!"
+        client-put-job-comment "$first_cluster_job_id" "$first_comment_id" "$comment"
+        echoerr "Unable to complete the investigation on job $id due to clone failure, check for: missing repositories, invalid configurations or use the env variable 'force=true' to trigger investigation jobs"
```
perlpunk, line 389: Note that the log will already contain an error message, so we probably don't need this additional one.

openqa-investigate
```diff
@@ -374,7 +374,22 @@ investigate() {
     comment_id=$(sync_via_investigation_comment "$id" "$first_cluster_job_id") || rc=$?
     [[ $rc != 255 ]] && return $rc
 
-    out=$(trigger_jobs "$id" "${@:2}") || return $?
+    out=""
+    if ! out=$(trigger_jobs "$id" "${@:2}"); then
+        first_comment_id=$(client-get-job-comments "$first_cluster_job_id" | runjq -r '[.[] | select(.text | contains("Investigation aborted"))] | min_by(.id) | .id')
+        if [[ $comment_id != "$first_comment_id" ]]; then
+            client-delete-job-comment "$first_cluster_job_id" "$comment_id"
```
perlpunk, line 381: I don't really get the logic here.
If `trigger_jobs` fails, you are searching for comments that contain "Investigation aborted", which only happen if `trigger_jobs` failed, so I'm wondering why we would find existing comments like this.
And then you delete this comment, but a few lines later you try to update this comment with `PUT`, but it might have been deleted.

openqa-investigate
```diff
@@ -374,7 +374,22 @@ investigate() {
...
+        fi
+        local comment="Investigation aborted for job $id.
+
+Reasons: clone failure.
+
+Suggestions: check for missing repositories or invalid configurations, investigate the job manually!"
```
perlpunk, line 387: I think we don't need too specific in here. 
```suggestion
        local comment="Triggering investigation jobs failed.
Check journal for reasons."
```

openqa-investigate
Wabri: I've added this after the test I've made in my local environment if the investigation is runned multiple time the abort comment will be published over and over again. With the requested change you made in https://github.com/os-autoinst/scripts/pull/398#discussion_r2075173772 the problem is solved because that comment trigger line 188 and stop the script with "Skipping investigation of job 1: job cluster is already being investigated, see comment on job 1"

If is the case is correct, I'll remove this change!

openqa-investigate
Wabri: Removed!

Wabri: by the way this review is in contrast to what oliver and marius comments above: https://github.com/os-autoinst/scripts/pull/398#discussion_r2073509225 and https://github.com/os-autoinst/scripts/pull/398#discussion_r2073510998

openqa-investigate
```diff
@@ -374,7 +374,14 @@ investigate() {
     comment_id=$(sync_via_investigation_comment "$id" "$first_cluster_job_id") || rc=$?
     [[ $rc != 255 ]] && return $rc
 
-    out=$(trigger_jobs "$id" "${@:2}") || return $?
+    out=""
+    if ! out=$(trigger_jobs "$id" "${@:2}"); then
+        local comment="Triggering investigation jobs failed.
+        Check journal for reasons."
```
okurz, line 380: Doesn't `$out` contain the reasons that the reader might be interested in? So how about something like

```suggestion
        local comment="Triggering investigation jobs failed: $out"
```

okurz (changes requested):

openqa-investigate
Wabri: Oh, I see. Definitely misunderstood!

openqa-investigate
Wabri: $out is empty if it pass the if statement, why print an empty variable then?

okurz: I see. Is there details in the stderr that we don't catch?

okurz (approved):
Let's try without further changes

openqa-investigate
Martchus, line 380:
```suggestion
        local comment="Triggering investigation jobs failed. Check journal for reasons."
```

Wabri: Change with including stderr in the openqa comment!

Wabri: I've included the stderr both on the job comment and as output of the script

Wabri:
I was working on integrate the change with unit tests, but if we want to change as it is, I'm ok with that!

okurz (approved):

Martchus (approved):

PR #399, opened 2025-05-06
Redirect to stderr when xmlstarlet is missing on the environment

b10n1k:
This is the only way to show the error if ever happen again. problem is that
the test can exit silent without the verbose mode in prove.

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -13,7 +13,7 @@ osc_poll_interval="${osc_poll_interval:-2}"
 osc_build_start_poll_tries="${osc_build_start_poll_tries:-30}"
 throttle_days=${throttle_days:-2}
 XMLSTARLET=$(command -v xmlstarlet || true)
-[[ -n $XMLSTARLET ]] || (echo "Need xmlstarlet" && exit 1)
+[[ -n $XMLSTARLET ]] || (echo "Need xmlstarlet" >&2 && exit 1)
 
 # shellcheck source=/dev/null
 . "$(dirname "$0")"/_common


Makefile
```diff
@@ -20,7 +20,7 @@ test: checkstyle test-unit
 test-unit: test-bash test-python
 
 test-bash: $(BPAN)
-	prove -r $(if $v,-v )$(test)
+	prove -v -r $(if $v,-v )$(test)
```
perlpunk, line 23: that doesn't make sense.
there is already a `-v` here in this line, can you see it?
but you have to enable it via a `make ... v=1`

perlpunk (commented):
regarding the original problem: i guess we would have seen the error (`Need xmlstarlet`) without `-v` if it had been printed to stderr instead of stdout.
So that's what I would suggest.

okurz (changes requested):
See comment by @perlpunk

Makefile
b10n1k: not showed it. and I guess you want to keep it like that. so i am closing this

perlpunk: > not showed it.

?

b10n1k:
> printed to stderr

💡The only way I found is to use the stderr in the script. I hope this is what you meant because it is not likely to happen again and I didn find any way to do this within prove.

os-autoinst-obs-auto-submit
```diff
@@ -13,7 +13,7 @@ osc_poll_interval="${osc_poll_interval:-2}"
 osc_build_start_poll_tries="${osc_build_start_poll_tries:-30}"
 throttle_days=${throttle_days:-2}
 XMLSTARLET=$(command -v xmlstarlet || true)
-[[ -n $XMLSTARLET ]] || (echo "Need xmlstarlet" && exit 1)
+[[ -n $XMLSTARLET ]] || (echo "Need xmlstarlet" >&2 && exit 1)
```
perlpunk, line 16: That is what I meant, but the PR description is now wrong.

perlpunk:
There are several issues with your commit message
> "Redirect to stdout when xmlstarlet is nissing on the environment"

You fixed the `nissing` -> `missing`. But you are printing to **stderr** not **stdout**, and the part "on the environment" is wrong and actually superfluous.
Also I would not phrase it as redirecting, although technically I guess it is. But we're printing something with echo and we want to land it on stderr, so why not just say "Print error about missing dependency to stderr"

> This is the only way to show the error if ever happen again. 

That is not a helpful commit message.
This is not about something happened and is now unlikely to happen.
First of all, it can happen for anyone calling the script or the test who has not installed xmlstarlet.

Error messages should go to stderr in general, not just because we had a problem not seeing it with `prove`.

So I thought I would just create #400 because it was quicker.

b10n1k:
> There are several issues with your commit message
> 
> > "Redirect to stdout when xmlstarlet is nissing on the environment"
> 
> You fixed the `nissing` -> `missing`. But you are printing to **stderr** not **stdout**, and the part "on the environment" is wrong and actually superfluous. Also I would not phrase it as redirecting, although technically I guess it is. But we're printing something with echo and we want to land it on stderr, so why not just say "Print error about missing dependency to stderr"
> 
> > This is the only way to show the error if ever happen again.
> 
> That is not a helpful commit message. This is not about something happened and is now unlikely to happen. First of all, it can happen for anyone calling the script or the test who has not installed xmlstarlet.
> 
> Error messages should go to stderr in general, not just because we had a problem not seeing it with `prove`.
> 
> So I thought I would just create #400 because it was quicker.

thanks for telling me. And yes likely it was faster.

PR #400, opened 2025-05-07
Print error about missing dependency to stderr

perlpunk:
Error messages should go to stderr in general


okurz:
> confused. isnt this what I did #399 (I have to fix a typo in the commit I just saw tho)

Yes same but wrong commit message. Hence @perlpunk created a commit with the same content but proper commit message

Martchus (approved):

PR #403, opened 2025-05-15
Fix regression in the os-autoinst-obs-auto-submit

b10n1k:
This is a regression from
https://github.com/os-autoinst/scripts/commit/360f247977c724e7b4aed8fd71b0fcb11e48e2e4 where the `$staging_project` does *not* overide `$src_project` before the `handle_auto_submit` but instead, it makes it take place after the function. That had side effect in the behavior of the script.

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -221,17 +221,12 @@ has_pending_submission() {
     return 0
 }
 
-is_staging_specified() {
-    if [[ -n $staging_project && $staging_project != none && $staging_project != "$src_project" ]]; then
-        src_project="$staging_project"
-        return 0
-    else
-        return 1
-    fi
-}
-
 submit() {
     # submit from staging project if specified
+    if [[ -n $staging_project && $staging_project != none && $staging_project != "$src_project" ]]; then
+        src_project=$staging_project
+        must_cleanup_staging=1
+    fi
     if [[ -e job_post_skip_submission ]]; then
         echo "Skipping submission, reason: "
         cat job_post_skip_submission
@@ -246,7 +241,7 @@ submit() {
             handle_auto_submit "$package" || rc=$?
         done
         # delete package from staging project
-        is_staging_specified && $osc rdelete -m "Cleaning up $package from $staging_project for next submission" "$staging_project" "$package"
+        [[ $must_cleanup_staging ]] && $osc rdelete -m "Cleaning up $package from $staging_project for next submission" "$staging_project" "$package"
         if [[ ${#failed_packages[@]} -gt 0 ]]; then
             echo "Failed packages:"
             for item in "${failed_packages[@]}"; do


os-autoinst-obs-auto-submit
```diff
@@ -232,6 +231,7 @@ is_staging_specified() {
 
 submit() {
     # submit from staging project if specified
+    is_staging_specified && [[ $staging_project != "$src_project" ]] && src_project=$staging_project
```
baierjan, line 234: I am wondering if the check `[[ $staging_project != "$src_project" ]]` is even necessary?

b10n1k:
💡I test it manual and the output looks ok ```./os-autoinst-obs-auto-submit: line 152: devel:openQA:testing/devel:openQA:tested/devel:openQA:tested.changes: No such file or directory
retry -e -- osc ci -m sync changesrevision with openSUSE:Factory devel:openQA:testing/devel:openQA:tested
retry -e -- osc up --server-side-source-service-files devel:openQA:testing/devel:openQA:tested
Latest revision already in openSUSE:Factory
retry -e -- osc rdelete -m Cleaning up devel:openQA:tested from devel:openQA:testing for next submission devel:openQA:testing devel:openQA:tested
```
💡I hope it is also more visible now when it should run the cleanup and what this condition is for

**Review by okurz — APPROVED** [review]

File: os-autoinst-obs-auto-submit [hunk 2]
```diff
@@ -232,6 +232,7 @@ is_staging_specified() {
 
 submit() {
     # submit from staging project if specified
+    is_staging_specified && src_project=$staging_project
```
Martchus, line 235:
```suggestion
    is_staging_specified && src_project=$staging_project must_cleanup_staging=1
```

And remove the assignment of `must_cleanup_staging` in `is_staging_specified` instead. (A function called `is_…` should not have side-effects like changing variables.)

Martchus (changes requested):

Martchus (approved):

os-autoinst-obs-auto-submit
```diff
@@ -232,6 +231,7 @@ is_staging_specified() {
 
 submit() {
     # submit from staging project if specified
+    is_staging_specified && src_project=$staging_project must_cleanup_staging=1
```
baierjan, line 234: Can we just inline the condition as the function is now used only here?

os-autoinst-obs-auto-submit
```diff
@@ -246,7 +246,7 @@ submit() {
             handle_auto_submit "$package" || rc=$?
         done
         # delete package from staging project
-        is_staging_specified && $osc rdelete -m "Cleaning up $package from $staging_project for next submission" "$staging_project" "$package"
+        [[ $must_cleanup_staging ]] && $osc rdelete -m "Cleaning up $package from $staging_project for next submission" "$staging_project" "$package"
```
baierjan, line 249: ...or is there a reason to not just use the `is_staging_specified` function and introduce yet another variable?

os-autoinst-obs-auto-submit
b10n1k: I guess we can.. but from my POV is a little better than have a long line. I think it is more readable with the function

okurz: So that would be

```suggestion
    [[ -n $staging_project && $staging_project != none && $staging_project != "$src_project" ]] || src_project=$staging_project must_cleanup_staging=1
```

right?

okurz (changes requested):

os-autoinst-obs-auto-submit
b10n1k: yea, this is the suggestion. I am just saying that even if the function is ony used here, it is a little better to read it using the function

kalikiana: I would at least keep the `if` and assignments in separate lines

```suggestion
    if [[ -n $staging_project && $staging_project != none && $staging_project != "$src_project" ]]; then
        src_project=$staging_project
        must_cleanup_staging=1
    fi
```

okurz: I prefer the inlined same as Jan

baierjan: In this case, you want to override the `src_project` if staging_project is set (and clean it if different from `src_project`), so `&&` is correct.

kalikiana: My point is the if is more legible 🙃

kalikiana (approved):

PR #406, opened 2025-05-26
Fix expansion for submit_target_extra

d3flex:
First the expansion is missing $ at the beginning and it makes it assign the expression as a string. Additional ensure that the get_submit_target_extra_latest_version is called as code and returns its value.

https://progress.opensuse.org/issues/127037

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -29,7 +29,7 @@ get_submit_target_extra_latest_version() {
     echo "$latest_submit_target_extra_version:Update"
 }
 
-submit_target_extra="{submit_target_extra:-get_submit_target_extra_latest_version}"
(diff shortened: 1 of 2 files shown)


os-autoinst-obs-auto-submit
```diff
@@ -29,7 +29,7 @@ get_submit_target_extra_latest_version() {
     echo "$latest_submit_target_extra_version:Update"
 }
 
-submit_target_extra="{submit_target_extra:-get_submit_target_extra_latest_version}"
+submit_target_extra="${submit_target_extra:-$(get_submit_target_extra_latest_version)}"
```
perlpunk, line 32:
```suggestion
submit_target_extra=${submit_target_extra:-$(get_submit_target_extra_latest_version)}
```

PR #408, opened 2025-06-03
Potentially improve os-autoinst-obs-auto-submit with "osc sr --cleanup"

okurz:
In particular during our work on
https://progress.opensuse.org/issues/127037 we found that over multiple
days no new submissions were created. We should consider to improve
os-autoinst-obs-auto-submit with "osc sr --cleanup" as suggested by
tinita in but consider what jbaier has stated that we need to ensure
that only packages from :tested are removed, not anywhere else like when
called manually on devel:openQA itself.

Related progress issue: https://progress.opensuse.org/issues/183482


os-autoinst-obs-auto-submit
```diff
@@ -114,6 +114,11 @@ update_package() {
         if test -n "$req"; then
             cmd="$cmd -s $req"
         fi
+        # On :tested projects, e.g. devel:openQA:tested, cleanup submissions
+        # after acceptance for next calls
```
Wabri, line 118: I dont get it.

os-autoinst-obs-auto-submit
```diff
@@ -114,6 +114,11 @@ update_package() {
...
+        if osc info | grep -q "Project name:.*:tested"; then
```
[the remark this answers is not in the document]
Martchus: > We want to do a cleanup when submitting from testing to tested, not from tested to TW.

That's also how I understand the problem.

However, we never submit from testing to tested. We only submit from tested to Factory. At least https://build.opensuse.org/projects/devel:openQA:testing/requests?involvement[]=incoming&involvement[]=outgoing&involvement[]=review&states[]=new&states[]=deleted&states[]=declined&states[]=accepted&states[]=review&states[]=revoked&states[]=superseded&project_names[]&package_names[]&creators[]&reviewers[]&staging_projects[]&created_at_from&created_at_to&search only shows some deletion request but no SRs.

That means this approach is not useful for the sake of ensuring `devel:openQA:testing` is cleaned up. We can still cleanup `devel:openQA:tested` although this likely doesn't bring any benefit.

os-autoinst-obs-auto-submit
okurz: Rephrased

PR #414, opened 2025-06-17
Run devel:openQA:testing jobs in separate job group

perlpunk:
...for easier distinction when something fails.

Issue: https://progress.opensuse.org/issues/178105

I already created the group: https://openqa.opensuse.org/group_overview/135


perlpunk:
> Pass the “--group 135” parameter in the OpenQA trigger call for matched jobs

hallucinating

trigger-openqa_in_openqa
```diff
@@ -21,7 +21,8 @@ machine="${machine:-"64bit"}"
 build_tag=${BUILD_TAG:-}
 client_prefix=${client_prefix:-}
 full_run=${FULL:-}
-group_id="${group_id:-"openQA"}"
+group_id=${group_id:-"openQA"}
+staging_group_id=${staging_group_id:-"openQA Staging"}
```
Martchus, line 25: I'm not sure whether I'd call the group "openQA Staging". The purpose of the contained jobs is to make submissions to Factory. So maybe "Factory submissions"? (That an intermediate repo is used is more an implementation detail which of course could be mentioned in the group description.)

perlpunk: Other opinions?

perlpunk: One thing I noticed - we would probably call the group "Factory submissions" without the openQA prefix.
If we pass this as `_GROUP` then it could potentially be ambiguous in he future, if there is ever another group named like that.
Should we have a feature to be able to pass an unambiguous job group by adding the parent as well, like `_GROUP="openQA : Factory submissions"? (and then forbid colons in group names)

I will rename the group now, as there were no other suggestions.

Martchus (approved):

okurz (approved):

PR #416, opened 2025-06-17
Further adjust search target for Maintenance requests

perlpunk:
When we submit to the target `openSUSE:Backports:SLE-15-SP6:Update`, the target project in the resulting request will be `openSUSE:Maintenance`.

Issue: https://progress.opensuse.org/issues/184210

## Summary by Sourcery

Bug Fixes:
- Fix mapping so that submitting to openSUSE:Backports:SLE-15-SP6:Update targets openSUSE:Maintenance as intended

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -32,6 +32,9 @@ get_obs_sr_id() {
     local target=$1
     local dst_project=$2
     local package=$3
+    # The project in the request target will be different than the original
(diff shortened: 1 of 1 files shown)


perlpunk:
Does this make sense? I left it as a draft for now

Martchus (approved):

okurz (commented):
dunno if it's correct but we can try

realcharmer (approved):

PR #418, opened 2025-06-23
Add project cleanup code from Jenkins to git

perlpunk:
Issue: https://progress.opensuse.org/issues/178105

It's hard to find out that this snippet exists, if it's only in the jenkins configuration.


perlpunk:
I set the not-ready label, so I can remove the code from Jenkins before merging.

okurz (approved):

perlpunk:
Ok, I removed the label and removed those lines from the jenkins config

PR #434, opened 2025-07-16
Add pyhton3-pytest devel dependency

perlpunk:


b10n1k (changes requested):
both entries and git title have a typo.

b10n1k (approved):

Martchus (approved):

perlpunk:
> and I wonder if there could be something to do here for such easy mistakes to prevent or get warned

Possibly. Do you have an idea?

b10n1k:
> > and I wonder if there could be something to do here for such easy mistakes to prevent or get warned
> 
> Possibly. Do you have an idea?

Something like https://github.com/os-autoinst/scripts/pull/435?

PR #435, opened 2025-07-17
[CI] run os-autoinst-scripts builds in obs

d3flex:
Making sure that the package builds, especially when there are changes in dependencies.yaml

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -1,0 +1,34 @@
+---
+pr:
+  steps:
+    - branch_package:
+        source_project: devel:openQA
+        source_package: os-autoinst-distri-scripts
+        target_project: devel:openQA:GitHub
+        add_repositories: disabled
+    - configure_repositories:
+        project: devel:openQA:GitHub
+        repositories:
+          - name: openSUSE_Tumbleweed
+            paths:
+              - target_project: openSUSE:Factory
+                target_repository: snapshot
+              - target_project: devel:openQA
+                target_repository: openSUSE_Tumbleweed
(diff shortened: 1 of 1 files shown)


.obs/workflows.yml
```diff
@@ -0,0 +1,34 @@
+---
+pr:
+  steps:
+    - branch_package:
+        source_project: devel:openQA
+        source_package: os-autoinst-distri-opensuse-deps
```
perlpunk, line 6: I think that should be:
```suggestion
        source_package: os-autoinst-scripts
```
https://build.opensuse.org/package/show/devel:openQA/os-autoinst-scripts

b10n1k: @perlpunk I am not sure if this is the only thing I need. before we merge I guess we have to add the tokens. and what else?

.obs/workflows.yml
```diff
@@ -0,0 +1,34 @@
...
+                target_repository: openSUSE_Tumbleweed
+            architectures: [ x86_64 ]
+
+  filters:
+    event: pull_request
+
+# Setup:
+# 1. Put this .obs/workflows.yml in the main branch of openQA
+# 2a. (Someone of our team) Create personal access token on GitHub with scope "repo"
+# 2b. Ensure it is renewed before expiry
```
perlpunk, line 26:
```suggestion
# 2. Copy the token from or `os-autoinst-obs-workflow` user in our password storage
```

b10n1k:
we updated password/-/blob/main/password too. So it should be ready to go

PR #438, opened 2025-07-18
Run unit tests in OBS builds

perlpunk:
This also fixes dependencies.
It adds them to BuildRequires so we see if a required package does not exist. (`Requires` aren't checked)

Issue: https://progress.opensuse.org/issues/185983

Tested in a branch in OBS

### DIFF
--- a/dist/rpm/os-autoinst-scripts-deps.spec
+++ b/dist/rpm/os-autoinst-scripts-deps.spec
@@ -24,13 +24,15 @@ License:        MIT
 Group:          Development/Tools/Other
 BuildArch:      noarch
 Url:            https://github.com/os-autoinst/scripts
+Source0:        scripts-%{version}.tar.xz
 # The following line is generated from dependencies.yaml
 %define main_requires bash coreutils curl grep html-xml-utils jq openQA-client openssh-clients osc perl >= 5.010 perl(Data::Dumper) perl(FindBin) perl(Getopt::Long) perl(Mojo::File) perl(Text::Markdown) perl(YAML::PP) python3-beautifulsoup4 python3-requests retry sed sudo xmlstarlet yq
 # The following line is generated from dependencies.yaml
-%define test_requires perl(Test::MockModule) perl(Test::Most) perl(Test::Output) perl(Test::Warnings)
+%define test_requires perl(Test::MockModule) perl(Test::Most) perl(Test::Output) perl(Test::Warnings) python3-pytest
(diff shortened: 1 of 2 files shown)


dist/rpm/os-autoinst-scripts-deps.spec
```diff
@@ -41,19 +43,23 @@ Suggests:       postgresql
...
 %prep
+%setup -n scripts-%{version}
 
 %build
 
 %install
 
 %check
+# Can't run bash tests because it needs to clone bpan via git
+make test-python
```
b10n1k, line 62: I wonder if just `%pytest` would do here? as described in https://en.opensuse.org/openSUSE:Packaging_Python#Running_tests

dist/rpm/os-autoinst-scripts-deps.spec
```diff
@@ -24,13 +24,15 @@ License:        MIT
 Group:          Development/Tools/Other
 BuildArch:      noarch
 Url:            https://github.com/os-autoinst/scripts
+Source0:        scripts-%{version}.tar.xz
```
b10n1k, line 27: where this come from? I didnt see it in https://build.opensuse.org/package/show/devel:openQA/os-autoinst-scripts which I would expect to find it, or in https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:scripts:PR-437/os-autoinst-scripts(from your test PR)

perlpunk:
> although the PR437 now shows the build results, which that was the objective 👍👍... I looked into the https://build.opensuse.org/build/devel:openQA:GitHub:os-autoinst:scripts:PR-437/openSUSE_Tumbleweed/x86_64/os-autoinst-scripts/_log which I believe is the right build, and I didnt see installation of python3-pytest, neither any pytest running? So I wonder what should I expect exactly?

https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:scripts:PR-438/os-autoinst-scripts/openSUSE_Tumbleweed/x86_64 is the correct build for this PR

dist/rpm/os-autoinst-scripts-deps.spec
perlpunk: That's created by the obs "tar" plugin
https://build.opensuse.org/projects/devel:openQA:GitHub:os-autoinst:scripts:PR-438/packages/os-autoinst-scripts/files/_service?expand=1

PR #439, opened 2025-07-22
Adapt openqa-schedule-mm-ping-test to uefi

kalikiana:
Fixes: https://progress.opensuse.org/issues/186158


kalikiana:
Maybe we actually want https://github.com/os-autoinst/scripts/pull/440 instead

PR #441, opened 2025-07-23
Improve behavior of `openqa-investigate` when jobs have been cancelled

Martchus:
* Avoid endless retries for jobs that are in the state `cancelled`
* Consider the investigation cancelled if any of the investigation jobs reached an inconclusive result like `none` (job was cancelled when it was still scheduled), `skipped` and `user_cancelled`
* See https://progress.opensuse.org/issues/181421

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -93,6 +93,7 @@ clone() {
     echo "* *$name*: t#$clone_id"
 
     # set priority of cloned jobs
+    [[ $base_prio == null ]] && return 0
     # shellcheck disable=SC2207
     clone_ids=($(echo "$out" | runjq -r 'values[]'))
     for id in "${clone_ids[@]}"; do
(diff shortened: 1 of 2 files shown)


openqa-investigate
```diff
@@ -232,32 +233,41 @@ fetch-investigation-results() {
...
+            is-finished "$state" || return 142
 
             result=$(echo "$job" | runjq -r '.result') || return $?
             echo "$investigate_type|$other_id|$result"
         fi
     done
 }
 
+is-finished() {
+    [[ $1 == done || $1 == cancelled ]]
```
okurz, line 245: From the failed style check:
```
 In openqa-investigate line 245:
    [[ $1 == done || $1 == cancelled ]]
             ^-- SC1010 (warning): Use semicolon or linefeed before 'done' (or quote to make it literal).
```

Wabri: Maybe using " to prevent this, also in the other checks

perlpunk (approved):

PR #443, opened 2025-07-29
Provide automatic extraction of host when arguments is a openqa link

Wabri:
Related to: https://progress.opensuse.org/issues/182357

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -8,8 +8,17 @@ set -o pipefail -o errtrace
 # shellcheck source=/dev/null
 . "$(dirname "${BASH_SOURCE[0]}")"/_common
 
-host="${host:-"openqa.opensuse.org"}"
+id=""
+if [[ ${BASH_SOURCE[0]} == "$0" ]]; then
+    id="${1:?"Usage: $0 <id> or https://<host>/tests/<id>"}"
+fi
+if [[ "$id" =~ ^(https|http)?://([^/]+)/tests/([0-9]+)$ ]]; then
+    scheme="${scheme:-${BASH_REMATCH[1]}}"
+    host="${host:-${BASH_REMATCH[2]}}"
+    id=${BASH_REMATCH[3]}
+fi
 scheme="${scheme:-"https"}"
+host="${host:-"openqa.opensuse.org"}"
 host_url="$scheme://$host"
 dry_run="${dry_run:-"0"}"
 min_search_term="${min_search_term:-"16"}"
@@ -130,10 +139,8 @@ label_on_issues_without_tickets() {
 }
 
 investigate_issue() {
-    local testurl=$1
-    local id="${testurl##*/}"
-    local reason
-    local curl_output
+    local testurl="$1"
+    local reason group_id curl_output
     echo "Requesting jobs/${id} via openqa-cli"
     job_data=$(openqa-cli "${client_args[@]}" jobs/"$id")
     state="$(echo "$job_data" | runjq -r '.job.state')"
@@ -188,4 +195,4 @@ label_issue() {
     investigate_issue "$testurl"
 }
 
-caller 0 > /dev/null || label_issue "$@"
+caller 0 > /dev/null || label_issue "$scheme://$host/$id"


openqa-label-known-issues
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash -ex
```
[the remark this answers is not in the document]
Wabri: Yes, I'm removing it :)

openqa-label-known-issues
```diff
@@ -8,6 +8,14 @@ set -o pipefail -o errtrace
 # shellcheck source=/dev/null
 . "$(dirname "${BASH_SOURCE[0]}")"/_common
 
+if [[ $# -lt 1 ]]; then
+    echo "Usage: $0 https://openqahost/tests/<job_id>"
+    exit 1
+fi
+
+if [[ "$1" =~ ^https?://([^/]+)/tests/([0-9]+)$ ]]; then
```
Martchus, line 16: I suggest you move this into `label_issue()` and use the `$testurl` variable initialized there. 
```suggestion
if [[ $testurl =~ ^https?://([^/]+)/tests/([0-9]+)$ ]]; then
```

Maybe you can replace the not so nice `Need 'testurl'` there with the `Usage: …` message you added above.

Martchus (changes requested):

openqa-label-known-issues
Wabri: Without the assignation of `$host` in the top of the script, we also need to move `$host_url` and `client_args`, I dont think is a good idea.

Wabri: But let me try!

openqa-label-known-issues
```diff
@@ -175,7 +176,23 @@ investigate_issue() {
...
+        host="${host:-${BASH_REMATCH[1]}}"
+    else
+        echo "ERROR: Invalid URL format. Expected: https://<host>/tests/<job_id>" >&2
+        exit 1
+    fi
+
+    scheme="${scheme:-"https"}"
+    host_url="$scheme://$host"
+    client_args=(api --header 'User-Agent: openqa-label-known-issues (https://github.com/os-autoinst/scripts)' --host "$host_url" --retries="$retries")
+
```
Wabri, line 195: But now it is working only without the host definition and fail if run with `host=... openqa-label-known-issues test/<number>`

Wabri: I'm changing the script to work in both way...

openqa-label-known-issues
```diff
@@ -8,6 +8,14 @@ set -o pipefail -o errtrace
 # shellcheck source=/dev/null
 . "$(dirname "${BASH_SOURCE[0]}")"/_common
 
+if [[ $# -lt 1 ]]; then
```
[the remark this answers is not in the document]
Wabri: With the previous change I made it was necessary to move the check there in order to get the proper argument in advance, but now that everything is move to `label_issue` I can remove this since is not useful anymore.

openqa-label-known-issues
Wabri: Changed!

openqa-label-known-issues
```diff
@@ -175,7 +171,37 @@ investigate_issue() {
 }
 
 label_issue() {
-    local testurl="${1:?"Need 'testurl'"}"
+    local arg="${1:-}"
+    if [[ -z "$arg" ]]; then
+        echo "Usage: $0 <job_id> or https://<host>/tests/<job_id>" >&2
+        exit 1
```
perlpunk, line 177: How about:
```suggestion
    local arg="${1:?"Usage: $0 <job_id> or https://<host>/tests/<job_id>"}"
```

perlpunk:
This PR is based on a commit from 2025-04-23.
Maybe time to rebase

openqa-label-known-issues
```diff
@@ -175,7 +171,37 @@ investigate_issue() {
...
+            echo "ERROR: host must be specified when passing only a job ID." >&2
+            echo "Example: host=openqa.suse.de $0 18554385" >&2
+            exit 1
+        fi
+        job_id="$arg"
+        testurl="$scheme://$host/tests/$job_id"
+    else
+        echo "ERROR: Unrecognized argument: $arg" >&2
+        exit 1
+    fi
```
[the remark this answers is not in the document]
Martchus: As if a `getopts` loop would be so much less verbose. I'd be ok with keeping it simple for now.

openqa-label-known-issues
```diff
@@ -175,7 +171,37 @@ investigate_issue() {
...
+    if [[ -z "$arg" ]]; then
+        echo "Usage: $0 <job_id> or https://<host>/tests/<job_id>" >&2
+        exit 1
+    fi
+
+    scheme="${scheme:-"https"}"
+    local job_id=""
+    local testurl=""
+
+    if [[ "$arg" =~ ^(https|http)?://([^/]+)/tests/([0-9]+)$ ]]; then
```
Martchus, line 184: The quotes on the left hand side of a `[[` condition don't make a difference:
```suggestion
    if [[ $arg =~ ^(https|http)?://([^/]+)/tests/([0-9]+)$ ]]; then
```
The same counts for the other `[[` conditions.

openqa-label-known-issues
Wabri: I'm not sure about this:

```
$ ./openqa-label-known-issues
./openqa-label-known-issues: line 174: 1: Usage: ./openqa-label-known-issues <job_id> or https://<host>/tests/<job_id>
$ ./openqa-label-known-issues
Usage: ./openqa-label-known-issues <job_id> or https://<host>/tests/<job_id>
```

the first run is with your suggestion and the second is with the current code.

As user I would prefer the second one because if I get the first one I would think about an unhandled error.

openqa-label-known-issues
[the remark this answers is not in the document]
okurz: Better see https://github.com/os-autoinst/scripts/pull/443#discussion_r2242046328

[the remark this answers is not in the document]
Wabri: https://github.com/os-autoinst/scripts/pull/443#discussion_r2244722782

openqa-label-known-issues
```diff
@@ -175,8 +171,34 @@ investigate_issue() {
...
 
 label_issue() {
-    local testurl="${1:?"Need 'testurl'"}"
-    [[ "$dry_run" = "1" ]] && client_prefix="echo"
+    local arg="${1:?"Usage: $0 <job_id> or https://<host>/tests/<job_id>"}"
+
+    scheme=${scheme:-https}
+    local job_id=""
+    local testurl=""
+
```
Martchus, line 179: I guess it would be nicer if this was a little bit more compact:
```suggestion
    scheme=${scheme:-https}
    local arg=${1:?"Usage: $0 <job_id> or https://<host>/tests/<job_id>"}
    local job_id= testurl=
```

As you mentioned in your other comment, the error message produced by the `?` syntax is not nice for user-facing errors. Not sure whether it is worth improving, though.

perlpunk: if `$1` is not supplied, then arg will never get assigned anything, because the script will be aborted.
And this is without `set -u`:
```
% cat empty.sh
#!/bin/bash
arg=${1:?oops}
echo "arg: $arg"
% ./empty.sh
./empty.sh: line 2: 1: oops
% ./empty.sh 23
arg: 23
```

openqa-label-known-issues
```diff
@@ -8,9 +8,20 @@ set -o pipefail -o errtrace
 # shellcheck source=/dev/null
 . "$(dirname "${BASH_SOURCE[0]}")"/_common
 
-host="${host:-"openqa.opensuse.org"}"
+id=""
+if [[ ${BASH_SOURCE[0]} == "$0" ]]; then
+    id="${1:?"Usage: $0 <id> or https://<host>/tests/<id>"}"
+fi
```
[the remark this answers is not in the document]
perlpunk: For that we actually have the `caller` thing in the last line, which will be true when sourced.
So maybe it would be good to avoid code outside of functions. That also makes it unit-testable

openqa-label-known-issues
```diff
@@ -188,4 +195,4 @@ label_issue() {
     investigate_issue "$testurl"
 }
 
-caller 0 > /dev/null || label_issue "$@"
+caller 0 > /dev/null || label_issue "$scheme://$host/$id"
```
perlpunk, line 198: That is actually not a url to an openqa test...

perlpunk (commented):
The problem is that this is bash and we rely on a lot of global variables, where we would use an object in other languages, or at least some data structure.

We want to rewrite this in a better scripting language at some point, but I create #448 now to show how I would do it right now

perlpunk:
> #448 implies a suggestion or alternative, but it is pretty much what this PR also does. Do not take me wrong but this is not how we should approach from others. I think anyone (no matter how difficult) can benefit from the collaboration. One of the PRs needs to close, and focus to one. @Wabri works on that for some time and anyone can write its own version. lets not distract him like this.

The question is, what should I have done alternatively?
Making suggestions to single parts of this PR would not work, the change I had to make is more complex.
Just not commenting at all and let @Wabri work on it much longer?
This PR has now been open for two weeks, and my goal is to have that feature implemented soon (because it is useful), or we just go with the current script.
It is not supposed to "distract" @Wabri . Sometimes a new PR is just the easiest way to get things done, if a PR doesn't get to a mergeable state.
Also my PR makes this function reusable, so maybe we can use it in the other similar scripts as well.

d3flex:
> > #448 implies a suggestion or alternative, but it is pretty much what this PR also does. Do not take me wrong but this is not how we should approach from others. I think anyone (no matter how difficult) can benefit from the collaboration. One of the PRs needs to close, and focus to one. @Wabri works on that for some time and anyone can write its own version. lets not distract him like this.
> 
> The question is, what should I have done alternatively? Making suggestions to single parts of this PR would not work, the change I had to make is more complex. Just not commenting at all and let @Wabri work on it much longer? This PR has now been open for two weeks, and my goal is to have that feature implemented soon (because it is useful), or we just go with the current script. It is not supposed to "distract" @Wabri . Sometimes a new PR is just the easiest way to get things done, if a PR doesn't get to a mergeable state. Also my PR makes this function reusable, so maybe we can use it in the other similar scripts as well.

I was not planning to continue. But I dont want you to take personal. My answer is... what happened to pair programming? 
Every time I see confusion or dissatisfaction from PRs, the easiest way out, in my mind, is XP... or something is not clear during reviews. Lets leave it here.

openqa-label-known-issues
Wabri: Yep, miss the `tests` string, but since this request changes is now takeover by #448 I'm not going to change this pr and just move into the new one!

perlpunk:
Feel free to take my changes.
I will be absent for a few days. Closing.

perlpunk:
I cannot reopen it anymore

PR #445, opened 2025-08-13
Makefile: Fix missing ruff format check

okurz:
ruff format and check are separate commands and need to be called
separatedly to ensure both correct format and style. Also see
https://github.com/astral-sh/ruff/discussions/9048 and
https://github.com/astral-sh/ruff/issues/8232

PR #447, opened 2025-08-13
Add openqa-list-incidents script

ricardobranco777:
This script allows you to list incidents for the openQA job.  For each incident it shows the relevant packages and tickets for bugzilla.suse.com and jira.suse.com

To fetch the titles of the tickets you can use the `-v` option which needs the private tokens for Bugzilla & Jira.
You can either modify this script or export then as environment variables: `BUGZILLA_TOKEN` and `JIRA_TOKEN`.

Imported from https://gitlab.suse.de/qac/openqa-lsinc

Adapted from original script by @grisu48. This version fetches information from SMELT, Bugzilla & Jira with batch processing & multithreading for speed.

### DIFF
--- a/openqa-list-incidents
+++ b/openqa-list-incidents
@@ -1,0 +1,298 @@
+#!/usr/bin/env python3
+"""
+This script allows you to list incidents for the openQA job.
+For each incident it shows the relevant packages and tickets
+for bugzilla.suse.com and jira.suse.com
+
+To fetch the titles of the tickets you can use the -v option
+which needs the private tokens for Bugzilla & Jira.
+You can either modify this script or export then as environment
+variables: BUGZILLA_TOKEN and JIRA_TOKEN
+"""
+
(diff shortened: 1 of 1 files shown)


okurz:
@d3flex @perlpunk here the same problem as in https://github.com/os-autoinst/scripts/pull/445#issuecomment-3183205248 showed up. For a repository called "scripts" I guess it can be expected that people will call their fork differently :) If this can't be fixed in OBS checks maybe consider to remove them again from PR checks. They can still be kept as tests for the target branch

perlpunk:
> If this can't be fixed in OBS checks maybe consider to remove them again from PR checks. They can still be kept as tests for the target branch

I assume this can be fixed somehow, or at least I would hope so, by somehow specifying the name in obs_scm. But I haven't found an option for that so far.
But for now this check is not required for mergify, so it won't block a merge.

PR #449, opened 2025-08-18
Submit packages via Git for the openSUSE:Leap:16.0 target project

Martchus:
The project `openSUSE:Leap:16.0` has been migrated to Git so we need to
submit there by creating Git PRs.

This changes requires setting up the config for `git-obs` in production
and deployment of SSH keys on https://src.opensuse.org/. See
https://opensuse-commander.readthedocs.io/en/latest/git-obs-quickstart.html
for details.

The commands contained by this change are based on the Git packaging
workflow documentation, see
https://en.opensuse.org/openSUSE:Git_Packaging_Workflow#Using_git-obs_and_plain_git_(fork_workflow).

Related ticket: https://progress.opensuse.org/issues/187356

---

I haven't tested this much except for a local run that managed to create
this submit request: https://src.opensuse.org/pool/openQA/pulls/1

So I'm waiting for feedback on that submit request. If it is good I'll undraft
this PR and will also see whether I can create unit tests for this with reasonable
effort.

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -73,9 +73,21 @@ wait_for_package_build() {
     fi
 }
 
+make_obs_submit_request() {
+    local package=$1 target=$2 version=$3 cmd=($osc sr) req='' rc=''
+    log-info "## Ready to submit $package to $target ##"
+    req=$(get_obs_sr_id "$target" "$dst_project" "$package" || :)
+    # TODO: check if it's a different revision than HEAD
+    [[ -n $req ]] && cmd+=(-s "$req")
+    if ! "${cmd[@]}" -m "Update to ${version}" "$target"; then
+        rc=$?
+        failed_packages+=("$package:$rc")
+    fi
+}
+
 update_package() {
     log-info "update_package $*"
-    local package=$1
+    local package=$1 version=
     xmlstarlet ed -L -i "/services/service" --type attr -n mode --value 'disabled' _service
     sed -i -e 's,mode="disabled" mode="disabled",mode="disabled",' _service
     reenable_buildtime_services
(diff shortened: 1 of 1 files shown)


d3flex:
@Martchus do you think it is good idea to start from scratch this script? Here how I see it.. it is a good change to start over with a clean implementation in Python. `osc` git and even the `git-obs` have an implementation in python (and I see that we can write also plugins) and it will make the life easier in the future. 

Do you or anyone else sees any reason to not do this?

Martchus:
I wouldn't make a rewrite of this script right now. We probably need to get back to the drawing board before we can rewrite - or rather redesign - this whole auto-submission approach. And that's definitely out of scope for what I'm trying to achieve here.

I know that some of the tooling exposes Python modules. I used them in qem-bot code. That's of course generally nice. I suppose the only downside is that these modules might not be stable so we'll have to adapt to API changes. Especially with these new workflows being still in the works this might not be ideal.

okurz (approved):

os-autoinst-obs-auto-submit
```diff
@@ -102,16 +150,13 @@ update_package() {
...
-        req=$(get_obs_sr_id "$target" "$dst_project" "$package" || :)
-        # TODO: check if it's a different revision than HEAD
-        if test -n "$req"; then
-            cmd="$cmd -s $req"
-        fi
-        if ! $cmd -m "Update to ${version}" "$target"; then
-            rc=$?
-            failed_packages+=("$package:$rc")
+        local git_branch=${git_branches[$target]}
+        if [[ $git_branch ]]; then
```
perlpunk, line 156: With this change it's a bit obscurednow  where `rc` gets set. Maybe the `make_...request` functions could actually return `$rc` and it can be catched here? Not sure

perlpunk: And actually, you are setting `local rc` in `make_obs_submit_request` (but not in `make_git_submit_request`), so it won't change the outer `rc` here

perlpunk (changes requested):

perlpunk (approved):

PR #452, opened 2025-09-02
monitor-openqa_job: Ensure script finishes regularly

perlpunk:
openqa-cli monitor exits with non zero when a job failed, so the script aborts and can't call `delete_packages_from_obs_project`, so the packages in devel:openQA:testing are leftover and the autosubmit process is stuck

(I thought I had seen this working in the past, not sure what changed...)

### DIFF
--- a/monitor-openqa_job
+++ b/monitor-openqa_job
@@ -31,7 +31,12 @@ declare -A failed_versions
 failed_jobs=()
 for job_id in $(job_ids job_post_response); do
     log-info "Waiting for job $job_id to finish"
-    $openqa_cli monitor --host "$host" --follow --poll-interval "$sleep_time" "$job_id"
+    rc=0
+    $openqa_cli monitor --host "$host" --follow --poll-interval "$sleep_time" "$job_id" || rc=$?
+    if [[ $rc != 0 && $rc != 2 ]]; then
+        log-warn "openqa-cli monitor failed with an unexpected error ($rc)"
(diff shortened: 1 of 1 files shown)


monitor-openqa_job
```diff
@@ -31,7 +31,7 @@ declare -A failed_versions
 failed_jobs=()
 for job_id in $(job_ids job_post_response); do
     log-info "Waiting for job $job_id to finish"
-    $openqa_cli monitor --host "$host" --follow --poll-interval "$sleep_time" "$job_id"
+    $openqa_cli monitor --host "$host" --follow --poll-interval "$sleep_time" "$job_id" || true
```
okurz, line 34: What is commonly done is the shorter

```suggestion
    $openqa_cli monitor --host "$host" --follow --poll-interval "$sleep_time" "$job_id" ||:
```

but I guess it would be nicer if the exit code is used to at least forward any problems encountered with a `log-info` or `log-warn`. How about that?

okurz (changes requested):

d3flex (approved):

monitor-openqa_job
perlpunk: the monitor command is only used to wait until the job is finished. The result of the job is retrieved with the next command.
If monitor fails, its error output will be in the log.
I could check the return code, and if it is 2, then the job has failed, if it is 1, then there was some other problem.
Should I do this?

PR #455, opened 2025-09-02
Fix reading build names

perlpunk:
Without the `-d` `read` will stop after the first line break. So we were only testing the latest build, which does not have a 64bit image.

Issue: https://progress.opensuse.org/issues/188136

I guess were just lucky for the last 3 years 1845c6d93026681eacfd0f02b46188f7d119a5a1 :four_leaf_clover:


kalikiana (approved):
🙀

nicksinger (approved):

PR #457, opened 2025-09-04
Improve error handling when forking fails

Martchus:
Related ticket: https://progress.opensuse.org/issues/187356

### DIFF
--- a/os-autoinst-obs-auto-submit
+++ b/os-autoinst-obs-auto-submit
@@ -88,7 +88,7 @@ make_obs_submit_request() {
 }
 
 make_git_submit_request() {
-    local package=$1 target=$2 version=$3 branch=$4 forked_repo=''
+    local package=$1 target=$2 version=$3 branch=$4 forked_repo='' fork_rc=0
     title="Update to $version"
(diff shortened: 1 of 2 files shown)


okurz (approved):

os-autoinst-obs-auto-submit
```diff
@@ -102,6 +102,12 @@ make_git_submit_request() {
     #       For `git push` to work (done below) you need to deploy an SSH key on the Gitea instance.
     #       Checkout https://opensuse-commander.readthedocs.io/en/latest/git-obs-quickstart.html for details.
     forked_repo=$($git_obs repo fork "pool/$package" 2>&1 > /dev/null | perl -ne 'print for /Fork .*: (.*)/g')
+    if [[ -z $forked_repo ]]; then
```
perlpunk, line 105: It would probably be good to print the actual output of the fork command, which should already explain the problem.
So catch the output and only after that apply your regex to get the repo_name
Otherwise we are blindly guessing about the cause of the error

perlpunk (approved):

PR #499, opened 2025-12-12
PART 2: Rewrite bash scripts to python - After #498+#505

okurz:
After
* #498
* #505

One main open point is where we should place the Python files. It would be good to have them in a sub-directory and a usual Python practice is to use a properly named package derived from the directory name so "src/os_autoinst_scripts/" sounds correct but is also rather long and also means for testing we need to adjust the Python import path. How about just "oas/" for (Os-Autoinst-Scripts) and ditch the "src/"?


Martchus (commented):
Looks generally good so far.

PR #505, opened 2026-01-08
PART 1: Add dependencies for modern Python development

okurz:
* feat(ci): Simplify deps setup with uv
* feat(pyproject.toml): Add config for proper python package paths
* feat(Makefile): Add more Python related targets using "uv"
* feat(pyproject.toml): Add dependencies in preparation for typecheck+maintainability checks
* feat(pyproject.toml): Enable dependencies for modern scripts (typer+httpx)
* fix(test_trigger_bisect_jobs): fix "path" style issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: astral-sh/setup-uv@v7
         with:
           enable-cache: true
-      - name: Setup enviroment
+      - name: Setup environment
         run: uv sync --extra dev
       - run: |
           sudo apt-get install cpanminus html-xml-utils xmlstarlet
@@ -36,8 +36,9 @@ jobs:
       - uses: astral-sh/setup-uv@v7
         with:
           enable-cache: true
+      - name: Setup environment
+        run: uv sync --extra dev
       - name: Static checks
         run: |
           git config --global --add safe.directory .
-          uv sync --extra dev
           uv run make checkstyle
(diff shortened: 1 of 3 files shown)


Martchus (commented):
Are these dependencies already required? If not I wouldn't separate adding them from the actual changes requiring them.

.github/workflows/ci.yml
```diff
@@ -6,24 +6,27 @@ jobs:
...
       - uses: actions/checkout@v4
-      - run: |
-          pip install --upgrade pip
-          pip install pytest mocker pytest-mock
-          python --version
-          pytest --version
+      - uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+      - run: uv sync
```
d3flex: I would expect to find also uv.lock in the repo. otherwise what the us is syncing?

okurz:
> Are these dependencies already required? If not I wouldn't separate adding them from the actual changes requiring them.

the dependencies would only be required in https://github.com/os-autoinst/os-autoinst-scripts/pull/499

Martchus (changes requested):
Then I say we should add the dependencies only as needed. At least for me it makes no sense to approve this independently from the actual code that I haven't looked at.

.github/workflows/ci.yml
Wabri: I think so as well, `uv.lock` is necessary for reproducible Python environments.

PR #534, opened 2026-04-06
Add openqa-scan script for your openQA needs

ricardobranco777:
Add openqa-scan script for your openQA needs:

- Scan traces & coredumps.
- List jobs.
- Cancel, comment, delete, prioritize and restart jobs.
- Accepts almost any kind of openQA URL that shows jobs.

NOTES
- To avoid its usage as a weapon of mass deletion, the script limits itself to 99 jobs.
- Actions that require an API key assume a working `$HOME/.netrc` supported by default by Python Requests without the need for special authentication code. The format is:
 `machine openqa.suse.de login $USER password $APIKEY:$APISECRET`
- The only dependencies are Python 3.11+ & Python Requests.

Example usage:

```
# List Tumbleweed x86_64 jobs
$ openqa-scan -b latest https://openqa.opensuse.org/group_overview/1

# Scan Tumbleweed x86_64 for coredumps & traces:
$ openqa-scan -x all https://openqa.opensuse.org/group_overview/1

# Scan SLE aggregates
$ openqa-scan -x all -d sle -v 15-SP6 -v 15-SP7 -b yesterday https://openqa.suse.de
$ openqa-scan -x all -d sle -v 15-SP4 -v 15-SP5 -b yesterday https://openqa.suse.de
```

```
usage: openqa-scan [-h] [-A {cancel,delete,restart}] [-a {aarch64,i586,ppc64le,riscv64,s390x,x86_64}] [-b BUILD] [-c COMMENT] [-d DISTRI] [-f FLAVOR] [-g GROUPID] [-p PRIORITY]
                   [-r {aborted,complete,failed,incomplete,none,not_complete,not_ok,obsoleted,ok,parallel_failed,parallel_restarted,passed,skipped,softfailed,timeout_exceeded,user_cancelled,user_restarted}]
                   [-s {assigned,cancelled,done,execution,final,pre_execution,running,scheduled,setup,uploading}] [-v VERSION] [-V] [-x {all,coredumps,traces}]
                   url [url ...]

positional arguments:
  url                   openQA url

options:
  -h, --help            show this help message and exit
  -A, --action {cancel,delete,restart}
  -a, --arch {aarch64,i586,ppc64le,riscv64,s390x,x86_64}
  -b, --build BUILD     build or YYYYMMDD date, today, yesterday or latest
  -c, --comment COMMENT
                        add comment
  -d, --distri DISTRI
  -f, --flavor FLAVOR
  -g, --groupid GROUPID
  -p, --priority PRIORITY
                        set priority
  -r, --result {aborted,complete,failed,incomplete,none,not_complete,not_ok,obsoleted,ok,parallel_failed,parallel_restarted,passed,skipped,softfailed,timeout_exceeded,user_cancelled,user_restarted}
  -s, --state {assigned,cancelled,done,execution,final,pre_execution,running,scheduled,setup,uploading}
  -v, --version VERSION
                        product version
  -V, --verbose         show comments and first line of traces
  -x, --extract {all,coredumps,traces}
```

### DIFF
--- a/openqa-scan
+++ b/openqa-scan
@@ -1,0 +1,668 @@
+#!/usr/bin/env python3
+
+import argparse
+import logging
+import os
+import re
+import resource
+import sys
+import textwrap
+from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import field, dataclass
+from datetime import datetime, timedelta
+from operator import itemgetter
+from urllib.parse import parse_qs, urlencode, urljoin, urlparse
+
+import requests
+from requests.adapters import HTTPAdapter
+from requests.exceptions import RequestException
+from requests.utils import parse_header_links
+
+# Ignore jobs with these strings in their name
+IGNORE_JOBS = (":investigate:",)
+
+# Ignore traces created by sysrq
+IGNORE_SYSRQ = (
+    "sysrq: Show backtrace",
+    "sysrq: Show Blocked State",
+    "sysrq: Show State",
+    "sysrq: Trigger a crash",
+)
+
+# Ignore traces that contain these strings:
+IGNORE_TRACES = (
+    # Ignore traces on xfstests jobs
+    {"trace": "btrfs", "job": "xfstests_"},
+    {"trace": "xfs_", "job": "xfstests_"},
+    # See https://bugzilla.suse.com/show_bug.cgi?id=1255220#c18
+    {"trace": "drm_wait_one_vblank"},
+    # Ignore OOM which is common on container tests
+    {"trace": "out_of_memory"},
+)
+
+TIMEOUT = 300
+
+session = requests.Session()
+
+
+@dataclass(frozen=True, kw_only=True)
+class Job:
+    """
+    Job class to hold metadata and optional logs & comments
+    """
+
+    url: str
+    info: dict
+    comments: list[dict] = field(default_factory=list)
+    logs: dict[str, dict[str, str]] = field(default_factory=dict)
+
+
+def get_file(url: str) -> str | None:
+    """
+    Download a text file and return its contents
+    """
+    try:
+        got = session.get(url, timeout=TIMEOUT)
+        got.raise_for_status()
+    except RequestException as error:
+        logging.error("%s: %s", url, error)
+        return None
+    return got.text
+
+
+def get_json(
+    url: str | None, method: str = "GET", key: str | None = None, **kwargs
+) -> dict | list[dict] | None:
+    """
+    Get JSON following pagination via Link header if present
+    """
+    results = []
+    while url:
+        try:
+            got = session.request(method, url, timeout=TIMEOUT, **kwargs)
+            got.raise_for_status()
+            data = got.json()
+        except RequestException as error:
+            logging.error("%s: %s", url, error)
+            return None
+
+        page = data[key] if key is not None else data
+        if method != "GET" or not isinstance(page, list):
+            if "errors" in page:
+                logging.error("%s: %s", url, ",".join(page["errors"]))
+            return page
+        results.extend(page)
+
+        if "Link" not in got.headers:
+            break
+        links = parse_header_links(got.headers["Link"])
+        url = next((x["url"] for x in links if x.get("rel") == "next"), None)
+
+    return results
+
+
+def get_job(
+    url: str,
+    include_comments: bool = False,
+    include_logs: list[str] | None = None,
+) -> Job | None:
+    """
+    Get a job, optionally including comments and logs
+    """
+    urlx = urlparse(url)
+    job_id = int(os.path.basename(urlx.path))
+    api_url = f"{urlx.scheme}://{urlx.netloc}/api/v1/jobs/{job_id}"
+    if include_logs:
+        api_url += "/details"
+    info = get_json(api_url, key="job")
+    if info is None:
+        return None
+    assert isinstance(info, dict)
+
+    url = f"{urlx.scheme}://{urlx.netloc}/tests/{job_id}"
+
+    comments: list[dict] = []
+    if include_comments:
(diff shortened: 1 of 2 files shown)


openqa-scan
```diff
@@ -0,0 +1,666 @@
...
+        "assigned",
+        "cancelled",
+        "done",
+        "running",
+        "scheduled",
+        "setup",
+        "uploading",
+    ]
+    # Meta states
+    states += ["execution", "final", "pre_execution"]
```
Martchus, line 589: This could use constants from https://github.com/os-autoinst/openQA-python-client/blob/main/src/openqa_client/const.py. Otherwise, don't expect use to keep this in-sync with openQA. Considering the code seems to use this only for `--help` it is probably not that bad, though.

openqa-scan
```diff
@@ -0,0 +1,666 @@
...
+        "parallel_restarted",
+        "passed",
+        "skipped",
+        "softfailed",
+        "timeout_exceeded",
+        "user_cancelled",
+        "user_restarted",
+    ]
+    # Meta results
+    results += ["aborted", "complete", "not_complete", "not_ok", "ok"]
```
Martchus, line 577: Mixing results and meta-results seems conceptually wrong. The same counts for states and meta-states below.

Martchus (commented):
Not sure whether it makes sense to merge this without even some basic automated tests.

openqa-scan
ricardobranco777: > Mixing results and meta-results seems conceptually wrong. The same counts for states and meta-stated below.

Why? These are accepted by the API.

openqa-scan
ricardobranco777: It's not synced and missing some constants.

ricardobranco777:
> Not sure whether it makes sense to merge this without even some basic automated tests.

I added some tests generated by Claude Sonnet 4.6.

openqa-scan
Martchus: I don't know an API route out of my head that takes e.g. a result and a meta-result via the same parameter. So what route do you mean?

Note that most routes probably accept anything as the specified result or state is not validated. So what is accepted is probably not very meaningful.

Martchus:
> I added some tests generated by Claude Sonnet 4.6.

Those tests look interesting. Not sure whether they test the right thing but I guess it is better than nothing. (I won't review these almost 1000 lines of generated code in detail.)

openqa-scan
ricardobranco777: > I don't know an API route out of my head that takes e.g. a result and a meta-result via the same parameter. So what route do you mean?

It's not a route.

Right now, `-r not_ok` on https://openqa.opensuse.org/group_overview/1 gives me only failed & skipped jobs while `-r ok` gives me only passed & softfailed jobs, which makes sense.

Martchus: I didn't know that we support that on the group overview. I guess than it is ok to mix those.

openqa-scan
```diff
@@ -0,0 +1,666 @@
...
+        text = get_file(log)
+        if text is not None:
+            logs[os.path.basename(log)] = {"text": text, "url": log}
+
+    return Job(url=url, info=info, comments=comments, logs=logs)
+
+
+def get_jobs(url: str, ids: list[str] | None = None) -> list[Job]:
+    """
+    Get jobs from overview route or a list of job ID's
```
perlpunk:
```suggestion
    Get jobs from overview route or a list of job IDs
```

openqa-scan
```diff
@@ -0,0 +1,666 @@
...
+    if builds is None:
+        return None
+    # We use sets to deduplicate and later we convert to list on return
+    info: dict[str, set] = defaultdict(set)
+    if urlx.netloc == "openqa.opensuse.org":
+        # Legit builds on o3 are YYYYMMDD
+        for build in builds:
+            if not build["build"].isdigit():
+                continue
+            # Set only distri's that are non-zero from
```
perlpunk:
```suggestion
            # Set only distris that are non-zero from
```

okurz (changes requested):
The test code shows significant duplication. Please refactor to improve this

Martchus (approved):

ricardobranco777:
> The test code shows significant duplication. Please refactor to improve this

Where do you see duplication?  They're testing each function.  

The signal to noise ratio is due to type hints (which were required by ruff in your CI).  The original tests didn't have it and I think type hints in tests is overkill.

okurz:
> > The test code shows significant duplication. Please refactor to improve this
> 
> Where do you see duplication? They're testing each function.
> 
> The signal to noise ratio is due to type hints (which were required by ruff in your CI). The original tests didn't have it and I think type hints in tests is overkill.

I see the following repeated 11 times

```
            patch.object(oqs, "get_all_jobs", return_value=jobs),
            patch.object(oqs, "get_urls", return_value=[""]),
            patch.object(oqs, "post_route") as mock_pr,
```

ricardobranco777:
> I see the following repeated 11 times
> 
> ```
>             patch.object(oqs, "get_all_jobs", return_value=jobs),
>             patch.object(oqs, "get_urls", return_value=[""]),
>             patch.object(oqs, "post_route") as mock_pr,
> ```

DRY is not really worth it in this case and with unit tests in general.  A refactor here would decrease readability.  Here you want to know directly what you're mocking and patching.

okurz:
> > I see the following repeated 11 times
> > ```
> >             patch.object(oqs, "get_all_jobs", return_value=jobs),
> >             patch.object(oqs, "get_urls", return_value=[""]),
> >             patch.object(oqs, "post_route") as mock_pr,
> > ```
> 
> DRY is not really worth it in this case and with unit tests in general. A refactor here would decrease readability. Here you want to know directly what you're mocking and patching.

I as repo owner disagree

ricardobranco777:
Script moved to new repo here where I can add experimental stuff:

https://github.com/ricardobranco777/openqa-scan

PR #537, opened 2026-04-11
feat(openqa-llm-investigation): improvements

okurz:


okurz:
@Vogtinator this is what a gemma4-E4B model concluded with the above changes for the "isosize" failure in https://openqa.opensuse.org/tests/5879250

> **LLM Investigation summary:**
> 1. This is an existing issue, as the test has failed consecutively three times, and it is not a widespread failure since no other jobs are failing in the same module (`isosize`) in this build.
> 2. The failure is isolated to this specific test (`rescue`) and is not present in other tests in this build.
> 3. BISECT: YES, because the failure is isolated to this single test and is a recurring failure (3 consecutive failures), strongly indicating a new regression that needs investigation.

Vogtinator:
That looks better, basically what I'd expect if the used model is equivalent to "three if's in a suit" at least. It still doesn't grasp that the .iso file is too fat and investigation won't help there, but it's progress

openqa-llm-investigate
```diff
@@ -108,35 +111,110 @@ def investigate(
...
+            sum(
+                1
+                for b_job in build_failed_jobs
+                if str(b_job.get("id")) != str(job_id)
+                and failed_modules_set.intersection(
+                    m.get("name") for m in b_job.get("modules", []) if m.get("result") == "failed"
+                )
+            )
+            if failed_modules
+            else 0
```
Martchus: This expression has gotten too complicated in my opinion. Why do you use `sum(1 for …)` here? Maybe I don't fully understand this code but wouldn't `length(…)` make more sense here?

Martchus (changes requested):

openqa-llm-investigate
okurz: In Python, you cannot use `len()` directly on a generator expression. In my understanding the pattern `sum(1 for x in iterable if condition)` is actually a standard and idiomatic Python way. It counts the elements that match the condition in a lazy, memory-efficient way without building an intermediate list. I have now applied multiple refactorings:
1. Extracted the complete outer indented block into its own function
2. Extracted a helper function for the inner check called `is_similar`

This brings this statement to a single line:

```
same_module_failures = sum(1 for j in build_failed_jobs if is_similar(j)) if failed_modules else 0
```

Martchus (approved):

perlpunk (approved):

PR #542, opened 2026-04-16
Increase number of attemps to check build on OBS

kalikiana:
30 is not enough and leads to alerts that are not actionable because OBS will eventually provide the builds.


os-autoinst-obs-auto-submit
```diff
@@ -12,7 +12,7 @@ submit_target="${submit_target:-"openSUSE:Factory"}"
 submit_target_extra_project="${submit_target_extra_project:-"openSUSE:Backports"}"
 dry_run="${dry_run:-"0"}"
 osc_poll_interval="${osc_poll_interval:-2}"
-osc_build_start_poll_tries="${osc_build_start_poll_tries:-30}"
+osc_build_start_poll_tries="${osc_build_start_poll_tries:-90}"
```
okurz, line 15: 90 means multiplied with "osc_poll_intervall" that we *only* sleep 180s so the change goes in the right direction. Do we need some exponential back-off here as well?

okurz (approved):

gauravpathak (approved):

PR #543, opened 2026-04-19
openqa-investigate: Overwrite status of original job when needed

ricardobranco777:
The bare investigative retry job is a full clone of the failed job and should overwrite the status of the original job with its own status if it passed or softfailed to avoid manual intervention.

Introduce the `_FORCE_STATUS_ON_RETRY` variable to make this feature opt-in.

Related ticket: https://progress.opensuse.org/issues/186708

### DIFF
--- a/openqa-investigate
+++ b/openqa-investigate
@@ -308,6 +308,10 @@ post-investigate() {
     if is-ok "$retry_result"; then
         comment+=" $retry_result."$'\n\n'
         comment+="📋 **Likely a sporadic failure**."
+        force_status="$(echo "job_data" | runjq -r .job.settings._FORCE_STATUS_ON_RETRY)"
+        if [ "$force_status" != "null" ] && [ "$force_status" -gt 0 ]; then
+            comment+=$'\n'"label:force_result:$retry_result:retry_job_$retry_result"
+        fi
     else
         local product_issue=false
         local test_issue=false


okurz (changes requested):
In the commit message /which its own/with its own/. 

The idea is nice but I am sure there are people that don't want the original test results to be overriden automatically. How about a separate hook script where we can define it's own group exclusion/inclusion rules? That would also make a more explicit distinction between "read-only" investigation and an automated decision process.

ricardobranco777:
> In the commit message /which its own/with its own/.

Fixed.

> The idea is nice but I am sure there are people that don't want the original test results to be overriden automatically. How about a separate hook script where we can define it's own group exclusion/inclusion rules? That would also make a more explicit distinction between "read-only" investigation and an automated decision process.

Let's ask other teams as that would duplicate the logic in other script.  I think in the worst case scenario we could implement this exclusion in a separate file sourced by this script.

I want to know why is the status of the last job important to some.

d3flex (commented):
that needs some careful thoughts. I dont know if it is a good idea because it has an implication on the db and any results are alternated. it wont make easier the review - investigation meant to just help with this IMO -. How can you say how spoadic is a failure? What if the so-called sporadic failure fails 10 in 10 times? and it might be for different reasons. I can think more reasons to avoid this edit in the `openqa-investigate`, but I leave it to others to have a say on it

ricardobranco777:
> that needs some careful thoughts. I dont know if it is a good idea because it has an implication on the db and any results are alternated. it wont make easier the review - investigation meant to just help with this IMO -. How can you say how spoadic is a failure? What if the so-called sporadic failure fails 10 in 10 times? and it might be for different reasons. I can think more reasons to avoid this edit in the `openqa-investigate`, but I leave it to others to have a say on it

What matters in this case is that the retried job passed.  It would avoid yet another restart from reviewers who mass restart all failed jobs in a jobgroup or who miss or don't bother to see the comments from the investigative retry jobs.

Martchus:
That seems useful but should probably be somehow configurable, e.g. by group ID/name. Not sure whether that should be a separate hook script (as @okurz suggested), though (as it would probably make things more complicated).

ricardobranco777:
> That seems useful but should probably be somehow configurable, e.g. by group ID/name. Not sure whether that should be a separate hook script (as @okurz suggested), though (as it would probably make things more complicated).

Or we could start by enabling this only for jobs that have a `RETRY` setting that is greater than zero.

okurz:
> > That seems useful but should probably be somehow configurable, e.g. by group ID/name. Not sure whether that should be a separate hook script (as @okurz suggested), though (as it would probably make things more complicated).

Well, certainly it's easier to just introduce the exclusion/inclusion lists or feature switch in the script where called. You can start with that

> Or we could start by enabling this only for jobs that have a `RETRY` setting that is greater than zero.

Do you mean that if a job with RETRY fails the first time we wouldn't retry but just overwrite the result? Or do you mean that if e.g. a scenario with RETRY=1 fails once, is retried, fails again, then triggers openqa-investigate, retry passes, 2nd original job is force-result'd? Don't quite see the relation. Also I am concerned that this will even more reduce the motivation for test owners to stabilize their tests and instead complain that openQA takes too long to execute jobs as one would often a "still running test" which is actually only the n-th retry of unstable, repeatedly failing scenarios

EDIT: I just found a good example https://openqa.suse.de/tests/21945920#comment-2487473

All openqa-investigate jobs are green:

Screenshot_20260420_190844_investigate_all_green

All recent previous jobs are green:

Screenshot_20260420_190903_previous_all_green

so this PR would force-result the failed job automatically. I am sure that no reviewer would notice. So no one will try to improve the test stability or look if this is maybe still a hidden product regression. I don't see us in the position just yet to make such automated decision at least not by default. How about a separate test setting `_AUTOMATIC_FORCE_PASSED_ON_RETRY_PASS=1` that test owners can add as "opt-in"?

ricardobranco777:
> Do you mean that if a job with RETRY fails the first time we wouldn't retry but just overwrite the result?

No.  This whole review process kicks in after the expected retries.

> Or do you mean that if e.g. a scenario with RETRY=1 fails once, is retried, fails again, then triggers openqa-investigate, retry passes, 2nd original job is force-result'd?

Yes.

> Also I am concerned that this will even more reduce the motivation for test owners to stabilize their tests and instead complain that openQA takes too long to execute jobs as one would often a "still running test" which is actually only the n-th retry of unstable, repeatedly failing scenarios

This is not a statement of fact and it relies on flawed psychological assumptions.  Trying to "motivate" people like this usually has the most unexpected outcomes, like the priority recalculation has shown us.

This has nothing to do with motivating or demotivating anyone.  It's about automating stuff that we do manually and improving the review experience.

And which tests are you referring to?  The public cloud tests have no RETRY setting.  Only a subset of the container tests do, like the upstream container tests, and these tests are only fixed upstream.  They are the most stable & reliable tests we have, but they can fail due to network issues, and that's why I introduced [openqa-bats-review](https://github.com/os-autoinst/os-autoinst-scripts/blob/master/openqa-bats-review) because relying on the status of the last job is just wrong when we should use the set intersection of all failures instead.  Sadly this script doesn't seem to be active in osd, but this PR could have helped here:

https://openqa.suse.de/tests/21931140#comments

ricardobranco777:
> EDIT: I just found a good example https://openqa.suse.de/tests/21945920#comment-2487473

The above job doesn't have a RETRY setting.

okurz:
> > EDIT: I just found a good example https://openqa.suse.de/tests/21945920#comment-2487473
> 
> The above job doesn't have a RETRY setting.

That's not relevant for my argument

ricardobranco777:
> > > EDIT: I just found a good example https://openqa.suse.de/tests/21945920#comment-2487473
> > 
> > 
> > The above job doesn't have a RETRY setting.
> 
> That's not relevant for my argument

It's relevant for my last position, but anyway I see nothing wrong with overriding the result in that job.

okurz:
I see two alternative ways going foward to make this mergeable:
1. Implement the feature switch as mentioned in https://github.com/os-autoinst/os-autoinst-scripts/pull/543#issuecomment-4282545245 or a group-inclusion/exclusion
2. Find a representative set of test owners supporting the current implementation as-is

ricardobranco777:
> I see two alternative ways going foward to make this mergeable:
> 
> 1. Implement the feature switch as mentioned in [openqa-investigate: Overwrite status of original job when needed #543 (comment)](https://github.com/os-autoinst/os-autoinst-scripts/pull/543#issuecomment-4282545245) or a group-inclusion/exclusion

Can we implement this via some setting like `RETRY_OVERRIDE` ?

> 2. Find a representative set of test owners supporting the current implementation as-is

With the above variable this feature would be opt-in.

okurz:
> > I see two alternative ways going foward to make this mergeable:
> > 
> > 1. Implement the feature switch as mentioned in [openqa-investigate: Overwrite status of original job when needed #543 (comment)](https://github.com/os-autoinst/os-autoinst-scripts/pull/543#issuecomment-4282545245) or a group-inclusion/exclusion
> 
> Can we implement this via some setting like `RETRY_OVERRIDE` ?

Well, in https://github.com/os-autoinst/os-autoinst-scripts/pull/543#issuecomment-4282545245 I suggested `_AUTOMATIC_FORCE_PASSED_ON_RETRY_PASS=1` but, yes, that would work. You just set the variable in test and read it out like you do with the `RETRY` variable already.

Also, complementary one could consider a job group based inclusion/exclusion, e.g. take a look how the existing variable https://github.com/os-autoinst/os-autoinst-scripts/pull/543/changes#diff-f73cf39a07f6cf8cdb453862496919d06df16d07e58b274e68ea148dd1f7dae5R19
is used.

ricardobranco777:
> Well, in [#543 (comment)](https://github.com/os-autoinst/os-autoinst-scripts/pull/543#issuecomment-4282545245) I suggested `_AUTOMATIC_FORCE_PASSED_ON_RETRY_PASS=1` but, yes, that would work. You just set the variable in test and read it out like you do with the `RETRY` variable already.

Opted for the shorter `_FORCE_STATUS_ON_RETRY` given that the status may be `passed` or `softfailed`.

okurz (approved):

Martchus (approved):

PR #546, opened 2026-04-21
openqa-advanced-retrigger-jobs: Allow to customize SQL command

okurz:
I used this now to retrigger all jobs in a certain failed module of the
past day like this:

```
sql_command="select j.id from jobs j join job_modules jm on jm.job_id = j.id where j.result='failed' and clone_id is null and t_finished >= '2025-08-25' and jm.name='my_module' and jm.result='failed'" openqa-advanced-retrigger-jobs
```

### DIFF
--- a/openqa-advanced-retrigger-jobs
+++ b/openqa-advanced-retrigger-jobs
@@ -18,13 +18,25 @@ Retrigger openQA jobs based on database queries.
 By default retriggers openQA jobs with '$result' since '$failed_since' on
 '$host'.
 
-Can be restricted to jobs that ran on worker by setting the variable 'WORKER'
-and optionally 'INSTANCE'.
-
-Needs SSH access to the target openQA host '$host'.
+Needs SSH access to the target openQA host '$host' to query the database
+unless 'JOB_IDS' is provided.
 
 Options:
- -h, --help         display this help
+ -h, --help           display this help
+
+Environment variables:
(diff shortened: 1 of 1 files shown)


aylen384:
Could you add this env var (and ideally also the rest) to the usage.

Martchus (approved):

okurz:
> Could you add this env var (and ideally also the rest) to the usage.

done

openqa-advanced-retrigger-jobs
```diff
@@ -51,7 +63,7 @@ main() {
     done
 
     [ "$dry_run" = "1" ] && client_prefix="echo"
-    sql_command="select id from jobs where (${worker_string}${result} and clone_id is null and t_finished >= '$failed_since'$additional_filters);"
+    sql_command="${sql_command:-"select id from jobs where (${worker_string}${result} and clone_id is null and t_finished >= '$failed_since'$additional_filters);"}"
```
Martchus: This has enough quotes so I'd at least avoid the outer quotes (which are not making a difference anyway):
```suggestion
    sql_command=${sql_command:-"select id from jobs where (${worker_string}${result} and clone_id is null and t_finished >= '$failed_since'$additional_filters);"}
```

okurz: done

nicksinger (approved):

PR #547, opened 2026-04-21
trigger-openqa_in_openqa: Allow to customize full_run_args

okurz:


trigger-openqa_in_openqa
```diff
@@ -108,7 +109,7 @@ trigger() {
     # where o3 is not reachable over https
     declare -a ARGS
     if [ "$target_host" = "openqa.opensuse.org" ]; then
-        ARGS=("OPENQA_HOST=http://openqa.opensuse.org")
+        ARGS=("OPENQA_HOST=http://openqa.opensuse.org" "$full_run_args")
     fi
```
Martchus: Why is this customization within `if [ "$target_host" = "openqa.opensuse.org" ]; then`?

This code will also add an empty array element by default.

```suggestion
    [[ $target_host = "openqa.opensuse.org" ]] && ARGS+=("OPENQA_HOST=http://openqa.opensuse.org")
    [[ $full_run_args ]] && ARGS+=("$full_run_args")
```

okurz: done

trigger-openqa_in_openqa
```diff
@@ -126,9 +127,8 @@ trigger() {
     # prevent host access problem when running within o3 infrastructure
     # where o3 is not reachable over https
     declare -a ARGS
-    if [ "$target_host" = "openqa.opensuse.org" ]; then
-        ARGS=("OPENQA_HOST=http://openqa.opensuse.org")
-    fi
+    [[ $target_host = "openqa.opensuse.org" ]] && ARGS+=("OPENQA_HOST=http://openqa.opensuse.org")
+    [[ $full_run_args ]] && ARGS+=("$full_run_args")
```
Martchus: Those args will not be split which I would expect here (even though it means one cannot pass spaces):

```suggestion
    [[ $full_run_args ]] && ARGS+=($full_run_args)
```

You'll probably have to tell shellcheck to be ok with that.

okurz: done

perlpunk (approved):

Martchus (approved):

PR #561, opened 2026-04-29
feat(reboot-stability-check): Add usage+help+importability

okurz:
* reboot-stability-check: Add usage+help

### DIFF
--- a/reboot-stability-check
+++ b/reboot-stability-check
@@ -26,4 +26,31 @@ main() {
     done
 }
 
+usage() {
+    cat << EOF
+Usage: $0 [OPTIONS] HOSTS...
+
+Conducts reboot stability checks against a space-separated list of hosts.
+
+Can be configured additionally with environment variables.
+
+Options:
+ -h, --help         display this help
+EOF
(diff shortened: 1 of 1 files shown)


reboot-stability-check
```diff
@@ -26,4 +26,31 @@ main() {
...
+
+Can be configured additionally with environment variables.
+
+Options:
+ -h, --help         display this help
+EOF
+    exit "$1"
+}
+
+opts=$(getopt -o h -l help -n "$0" -- "$@") || usage 1
```
d3flex, line 43: it's totally fine but i would like your opinion.. if you grep the repo for `getopt` you find only `./openqa-search-maintenance-core-jobs:187:while getopts :a:d: name; do`
The question is wherether we should stay consistent?

okurz: No, getopt is used in multiple other scripts. getopt was chosen to support long options like --help, which getopts does not natively support

d3flex (approved):

Martchus (approved):

PR #566, opened 2026-05-05
Improved Python style checks for other usual projects we maintain 

d3flex:
`ty` was initially included to match qem-bot's setup, but the minimum Python version it supports as a check target is 3.7.  production systems seems still on Leap 15.6 default to Python 3.6.  `ty` has been removed from this PR and left for a separate one where the scope and approach can be properly defined.

Initial I tried to copy exactly what is in qem-bot so all the ruff rules should apply the same. 
minor changes in Makefile.
Added uv.lock

https://progress.opensuse.org/issues/191575

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,17 +8,19 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v6
-      - run: |
-          pip install --upgrade pip
-          pip install .[dev]
-          python --version
-          pytest --version
+      - uses: actions/setup-python@v6
+        with:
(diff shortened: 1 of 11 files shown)


uv.lock
okurz, line 1: All good in this PR with exception of this file. This file shouldn't be commited unless we have an automatic way to update it. Also see https://github.com/openSUSE/qem-bot/issues/323 . Your PR will likely have everything that is necessary if you just leave this file out

okurz (changes requested):

Martchus (approved):

okurz (approved):

PR #569, opened 2026-05-05
feat(openqa-investigate-multi): Add usage+help+importability

okurz:
* openqa-investigate-multi: Allow to import for testability and extendability
* openqa-investigate-multi: Add usage+help

### DIFF
--- a/openqa-investigate-multi
+++ b/openqa-investigate-multi
@@ -7,6 +7,19 @@ set -euo pipefail
 . "$(dirname "${BASH_SOURCE[0]}")"/_common
 
 main() {
+    opts=$(getopt -o h -l help -n "$0" -- "$@") || usage 1
+    eval set -- "$opts"
+    while true; do
+        case "$1" in
+            -h | --help) usage 0 ;;
+            --)
+                shift
+                break
+                ;;
+            *) break ;;
+        esac
+    done
+
     local rc=0
     # shellcheck disable=SC2013
     for i in $(cat - | sed 's/ .*$//'); do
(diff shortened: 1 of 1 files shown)


openqa-investigate-multi
```diff
@@ -18,4 +28,16 @@ main() {
     exit $rc
 }
 
+usage() {
+    cat << EOF
+Usage: echo jobnumber | $0 [OPTIONS]
```
perlpunk: Actually this is taking multiple job numbers, one per line, right?
Maybe the usage should reflect that.

Martchus (approved):

openqa-investigate-multi
okurz: done

perlpunk (commented):
Somehow github doesn't take my comment, let's see what happens if I click on submit review

perlpunk:
```
Usage: echo "JOB_ID1 [JOB_ID2 ...]" | $0 [OPTIONS]
```
Is it working for multiple ids on one line?

(I can't post an inline comment right now, github internal server error)

d3flex (changes requested):
comment from @perlpunk is valid. look at the test
```bash
❯ echo "123 456" | sed 's/ .*$//' 
123
```

okurz:
> ```
> Usage: echo "JOB_ID1 [JOB_ID2 ...]" | $0 [OPTIONS]
> ```
> 
> Is it working for multiple ids on one line?

No, doesn't. I use the same usage string as in openqa-label-known-issues-multi now

perlpunk (approved):

d3flex (approved):

PR #588, opened 2026-05-21
feat: add zombie-reaper.py to handle zombie qemu left-overs

okurz:
Related issue: https://progress.opensuse.org/issues/201144

### DIFF
--- a/s390x-qemu-zombie-reaper.py
+++ b/s390x-qemu-zombie-reaper.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python3
+# Copyright SUSE LLC
+# ruff: noqa: T201, S404
+
+"""Automated openQA Zombie Reaper for s390x.
+
+Checks s390x hypervisors for persistent zombie qemu processes and reboots the host
+if any are found, while also retriggering the affected openQA jobs.
+"""
+
+from __future__ import annotations
+
+import json
+import shlex
+import subprocess
+import time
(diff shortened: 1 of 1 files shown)


Martchus (commented):
The script name should probably reflect that this is about s390x and qemu processes specifically.

I think it is part of the normal operation that qemu processes are in the zombie state for a very short time (until the parent process reads the exit status). It is probably unlikely that the script will take action in those cases but it isn't ideal.

okurz:
> The script name should probably reflect that this is about s390x and qemu processes specifically.

done

> I think it is part of the normal operation that qemu processes are in the zombie state for a very short time (until the parent process reads the exit status). It is probably unlikely that the script will take action in those cases but it isn't ideal.

hm, not sure if that is realistic but ok :) Made the script reasonably more complicated by doing the following
* look for zombies
* if any found, sleep 10s
* look for same zombie PIDs again, if found, trigger reboot + job restart

perlpunk:
> * look for same zombie PIDs again, if found, trigger reboot + job restart

and then, to be safe, check if it's not a meanwhile newly started process with the same PID ;-)

okurz:
> > * look for same zombie PIDs again, if found, trigger reboot + job restart
> 
> and then, to be safe, check if it's not a meanwhile newly started process with the same PID ;-)

done

Martchus (approved):

perlpunk (approved):

PR #589, opened 2026-06-01
feat: add --clear option to remove due dates from Resolved tickets

asmorodskyi:
Allow backlog-set-due-date to also clear due dates for tickets
  in Resolved status via the -c/--clear flag.

### DIFF
--- a/backlog-set-due-date
+++ b/backlog-set-due-date
@@ -15,6 +15,9 @@ Usage: $0 [OPTIONS]
 This script sets the due date on tickets in Redmine based on
 specified conditions.
 
+By default it sets due dates on In Progress tickets. With --clear
+it removes due dates from Resolved tickets instead.
+
 If you want to run this for testing there are two options:
 1. Run the script without updating the tickets
 dry_run=1 ./backlog-set-due-date
@@ -24,16 +27,29 @@ dry_run=1 ./backlog-set-due-date <issues-json-file>
 
 Options:
  -h, --help         display this help
+ -c, --clear        remove due dates from Resolved tickets
 EOF
     exit "$1"
 }
 
+update-due-date() {
+    local due_date=$1 notes=$2 id=$3
+    $prefix curl -v -H "X-Redmine-API-Key: $redmine_api_key" -H 'Content-Type: application/json' -X PUT \
+        -d "{\"issue\": {\"due_date\": \"$due_date\", \"notes\": \"$notes\"}}" \
+        "$host/issues/$id.json"
+}
+
 main() {
-    opts=$(getopt -o h -l help -n "$0" -- "$@") || usage 1
+    local clear=0
+    opts=$(getopt -o hc -l help,clear -n "$0" -- "$@") || usage 1
     eval set -- "$opts"
     while true; do
         case "$1" in
             -h | --help) usage 0 ;;
+            -c | --clear)
+                clear=1
+                shift
+                ;;
             --)
                 shift
                 break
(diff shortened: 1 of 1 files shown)


asmorodskyi:
I haven't tested it because I don't have API key

backlog-set-due-date
```diff
@@ -54,13 +68,22 @@ main() {
...
+                -d '{"issue": {"due_date": null, "notes": "Removing due date because ticket is Resolved"}}' \
+                "$host/issues/$id.json"
+        done
+    else
+        for id in $(jq .id "$issues"); do
+            due_date=$(date -d"+$duration" +%Y-%m-%d)
+            echo "Updating ticket $id, new due date setup to $due_date"
+            $prefix curl -v -H "X-Redmine-API-Key: $redmine_api_key" -H 'Content-Type: application/json' -X PUT \
+                -d "{\"issue\": {\"due_date\": \"$due_date\", \"notes\": \"Setting due date based on mean cycle time of SUSE QE Tools\"}}" \
+                "$host/issues/$id.json"
```
okurz: Quite some duplication but should be easy to refactor with an extracted helper function and then call it like


```suggestion
            set-due-date "$due_date" "Setting due date based on mean cycle time of SUSE QE Tools" "$id"
```

and correspondingly above with a literal "null" instead of computed due-date

okurz (changes requested):

asmorodskyi:
also I think we should rename the file because current name not accurate anymore but I would like to do in separate commit so the diff of actual changes would be still readable . Does it make sense ?

backlog-set-due-date
```diff
@@ -15,6 +15,9 @@ Usage: $0 [OPTIONS]
 This script sets the due date on tickets in Redmine based on
 specified conditions.
 
+By default it sets due dates on In Progress tickets. With --clear
```
Martchus, line 18:
```suggestion
By default, it sets due dates on "In Progress" tickets. With --clear
```

backlog-set-due-date
```diff
@@ -15,6 +15,9 @@ Usage: $0 [OPTIONS]
...
+it removes due dates from Resolved tickets instead.
```
Martchus, line 19:
```suggestion
it removes due dates from "Resolved" tickets instead.
```

backlog-set-due-date
```diff
@@ -24,16 +27,29 @@ dry_run=1 ./backlog-set-due-date <issues-json-file>
 
 Options:
  -h, --help         display this help
+ -c, --clear        remove due dates from Resolved tickets
```
Martchus, line 30:
```suggestion
 -c, --clear        remove due dates from "Resolved" tickets
```

asmorodskyi:
so verification runs showing that it does not work out of the box . and one hour time invested didn't help to find why . moving to draft and postponing work on it for now

backlog-set-due-date
```diff
@@ -24,16 +27,29 @@ dry_run=1 ./backlog-set-due-date <issues-json-file>
...
 EOF
     exit "$1"
 }
 
+update-due-date() {
+    local due_date=$1 notes=$2 id=$3
+    $prefix curl -v -H "X-Redmine-API-Key: $redmine_api_key" -H 'Content-Type: application/json' -X PUT \
+        -d "{\"issue\": {\"due_date\": \"$due_date\", \"notes\": \"$notes\"}}" \
```
perlpunk, line 38: If `$due_date` is `null`, then you're passing the string `null`, which is probably invalid (as it is not a date). So in case of `null` the quotes should be removed.
You wrote that it failed, and I don't know what exactly failed, but this could be the reason.

PR #593, opened 2026-06-10
feat: Fail submission if packages have not been cleaned up for days

kalikiana:
Currently there's only INFO messages and even though this will break things Jenkins pipelines are successful when there are old packages.

### DIFF
--- a/trigger-openqa_in_openqa
+++ b/trigger-openqa_in_openqa
@@ -122,6 +122,11 @@ create_devel_openqa_snapshot() {
     if [[ $staged_packages ]]; then
         echo "+++ NOTE+++ Only triggering tests from $src_project (not overriding $staging_project and doing a submission) because $staging_project still contains packages. You might need to do a manual cleanup if no other pipeline is running. Packages: $staged_packages" \
             | tee job_post_skip_submission
+        local staging_project_url="https://build.opensuse.org/project/show/$staging_project"
+        if curl -sL "$staging_project_url" | grep -q Days; then
+            echo "Stale packages have not been cleaned up in 1+ days. Check $staging_project_url"
+            return 1
+        fi
         staging_project=$src_project
         return
     elif [[ -e job_post_skip_submission ]]; then


trigger-openqa_in_openqa
```diff
@@ -122,6 +122,11 @@ create_devel_openqa_snapshot() {
     if [[ $staged_packages ]]; then
         echo "+++ NOTE+++ Only triggering tests from $src_project (not overriding $staging_project and doing a submission) because $staging_project still contains packages. You might need to do a manual cleanup if no other pipeline is running. Packages: $staged_packages" \
             | tee job_post_skip_submission
+        local staging_project_url="https://build.opensuse.org/project/show/$staging_project"
+        if curl -sL "$staging_project_url" | grep -q -v Days; then
```
Martchus: How does this condition check for "in 6+ hours"?

okurz (changes requested):
https://github.com/os-autoinst/os-autoinst-scripts/pull/593/changes#r3389054502

trigger-openqa_in_openqa
perlpunk: Also, `Days` will never be in the curl output. 1. It's lowercase "days" 2. The package list is generated by javascript, so the HTML will never contain any package names or dates.
I would try using `osc api`. It doesn't provide package timestamps for the whole project, but if there are any packages (`osc ls devel:openQA:testing`), one can check at least one of them with `obs api /source/devel:openQA:testing/os-autoinst/_history` to get the timestamp.

kalikiana: Fixed.  Thanks for noticing!

trigger-openqa_in_openqa
```diff
@@ -122,6 +122,11 @@ create_devel_openqa_snapshot() {
     if [[ $staged_packages ]]; then
         echo "+++ NOTE+++ Only triggering tests from $src_project (not overriding $staging_project and doing a submission) because $staging_project still contains packages. You might need to do a manual cleanup if no other pipeline is running. Packages: $staged_packages" \
             | tee job_post_skip_submission
+        local staging_project_url="https://build.opensuse.org/project/show/$staging_project"
+        if curl -sL "$staging_project_url" | grep -q Days; then
+            echo "Stale packages have not been cleaned up in 1+ days. Check $staging_project_url"
+            return 1
+        fi
```
okurz, line 129: The above comment says "You might need" so it's conditional. Also the message states "if no other pipeline is running" which isn't checked here. Are we sure this is the right condition to consistently fail on? What other situations can arise where this failure would just annoy as because we already know that we have a situation where there are still packages? What if package submissions are just pending acceptance for valid reasons?

okurz (changes requested):

trigger-openqa_in_openqa
perlpunk: did you completely miss my comment? this can't work

trigger-openqa_in_openqa
perlpunk: > Also the message states "if no other pipeline is running" which isn't checked here.

We don't run pipelines in parallel, and for this case it wouldn't be a problem anyway.

> Are we sure this is the right condition to consistently fail on? What other situations can arise where this failure would just annoy as because we already know that we have a situation where there are still packages? What if package submissions are just pending acceptance for valid reasons?

the testing project should always be cleaned after openQA tests have finished.
if there are still packages for longer, something went wrong. we can try to improve that situation by cleaning the project in more cases, but right now we have the situation that the whole submission process is blocked by this and noone notices if that happens.

but the implementation of the check won't work like that, as i wrote in my other comment

trigger-openqa_in_openqa
Martchus: > this can't work

I also don't see how this can work.

Martchus (changes requested):

PR #594, opened 2026-06-11
chore: Ignore temporary dirs in git

okurz:
* Also ignore temporary bpan dir in git
* Ignore dynamically downloaded test-more-bash dir in git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /__pycache__/
 .*.swp
 /uv.lock
+test-more-bash/
+
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@
 .*.swp
 /uv.lock
 test-more-bash/
+bpan/
 


Martchus (approved):

Wabri (changes requested):
chore(blocking): Commits doesn't follows the conventional commits, I would prefer to have a org wide rules about this

okurz:
> chore(blocking): Commits doesn't follows the conventional commits, I would prefer to have a org wide rules about this

fixed. And created another PR to introduce gitlint:
* #599

Wabri (approved):

.gitignore
```diff
@@ -3,3 +3,6 @@
 /__pycache__/
 .*.swp
 /uv.lock
+test-more-bash/
+bpan/
```
perlpunk, line 7: When are those directories created? I don't see them locally

okurz: hm, maybe those are actually some old left over directories as https://github.com/os-autoinst/os-autoinst-scripts/blob/master/Makefile#L106 would create a bpan dir but it would be called .bpan

perlpunk: I suggest to revert it in #656

PR #595, opened 2026-06-11
feat: rewrite monitor-openqa_job in Python

okurz:
## Motivation
Replace the bash script with a Python-based implementation to enhance
maintainability, logging, and error handling.

## Design Choices
Use typer for CLI arguments, httpx for API calls, and subprocess.run for
osc commands. Implement polling, retries, and standard logging in Python.

## Verification
Added unit test with complete coverage. Tested manually by downloading
the latest `job_post_response` from
http://jenkins.qa.suse.de/job/monitor-openQA_in_openQA-TW/
and executing `monitor-openqa_job` which shows

```
INFO: Waiting for job 6010424 to finish
…
INFO: Result of job 6010429: passed
```

## Hints to reviewers
`monitor-openqa_job` is likely only used within http://jenkins.qa.suse.de/job/monitor-openQA_in_openQA-TW/ and single-purpose. I see low risk of this rewrite PR for a drop-in replacement. If problems are encountered then we will likely soon observe them leading to pipeline failures.

## Benefits
Provides reliable, testable, and maintainable monitoring behavior with complete
unit test coverage.

### DIFF
--- a/monitor-openqa_job
+++ b/monitor-openqa_job
@@ -1,97 +1,275 @@
-#!/bin/bash -ex
+#!/usr/bin/env python3
+# Copyright SUSE LLC
+# ruff: noqa: D103, PLR0913, PLR0917, S404, S603, T201, TRY400
+"""Monitor an openQA job by polling the status of a job over the API."""
 
-set -euo pipefail
+import json
+import logging
+import os
+import pathlib
+import re
+import shlex
+import subprocess
+import sys
+import time
+from typing import Annotated, Any
 
-# configuration variables with defaults.
-color=${color:-auto}
-sleep_time="${sleep_time:-"10"}"
-openqa_cli="${openqa_cli:-"openqa-cli"}"
-host="${host:-"https://openqa.opensuse.org"}"
-openqa_groupid="${openqa_groupid:-"24"}"
-obs_component="${obs_component:-"package"}"
-obs_package_name="${1:-""}"
-staging_project=${staging_project:-devel:openQA:testing}
-comment_on_obs=${comment_on_obs:-}
-OPENQA_CLI_RETRIES="${openqa_cli_retries:-7}"
+import httpx
+import tenacity
+import typer
 
-export OPENQA_CLI_RETRIES
+log = logging.getLogger(__name__)
 
-usage() {
-    cat << EOF
-Usage: $0 [OPTIONS]
 
+def load_job_ids(path: str = "job_post_response") -> list[int]:
+    try:
+        with pathlib.Path(path).open("r", encoding="utf-8") as f:
+            data = json.load(f)
+        ids = data.get("ids", [])
+        return [int(jid) for jid in ids]
+    except FileNotFoundError as e:
+        print(f"Need job response status file '{path}'", file=sys.stderr)
+        raise typer.Exit(code=2) from e
+    except Exception as e:
+        log.error("Failed to read or parse '%s': %s", path, e)
+        raise typer.Exit(code=2) from e
+
+
+def fetch_job_api(client: httpx.Client, url: str, params: dict[str, Any], retries: int) -> dict[str, Any]:
+    for attempt in tenacity.Retrying(
+        stop=tenacity.stop_after_attempt(retries),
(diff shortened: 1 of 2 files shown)


monitor-openqa_job
```diff
@@ -1,97 +1,252 @@
...
+                    failed_versions[version] = 1
+
+    if not failed_jobs:
+        raise typer.Exit(code=0)
+
+    log.warning("%d jobs did not pass:", len(failed_jobs))
+    for fid in failed_jobs:
+        log.warning("%s/t%d", host_val, fid)
+
+    delete_packages_from_obs_project(staging_project_val, osc_cmd_str)
```
Martchus, line 260: Please retain explanatory comments. This line had previously the comment `# delete packages from staging project in error case as we will not continue submitting those packages` which explained *why* we do the deletion here. Maybe there are also other lines which have lost useful comments in this rewrite.

Martchus (changes requested):

monitor-openqa_job
okurz: you are right about this comment. there was one other comment stating "comment on failed jobs" which is only explaining the *what* and I changed that to be an explicit self-explanatory function call.

Done

tests/test_monitor_openqa_job.py
```diff
@@ -0,0 +1,195 @@
...
+import sys
+from typing import Any
+from unittest.mock import MagicMock, Mock
+
+import httpx
+import pytest
+import typer
+from pytest_mock import MockerFixture
+
+# Load the script dynamically as a module
```
Wabri, line 23: nitpick(if-minor): I think the line below are self-explanatory and this comment don't add anything more

tests/test_monitor_openqa_job.py
```diff
@@ -0,0 +1,195 @@
...
+rootpath = pathlib.Path(__file__).parent.parent.resolve()
+loader = importlib.machinery.SourceFileLoader("monitor_job", f"{rootpath}/monitor-openqa_job")
+spec = importlib.util.spec_from_loader(loader.name, loader)
+assert spec is not None
+monitor_job = importlib.util.module_from_spec(spec)
+sys.modules[loader.name] = monitor_job
+loader.exec_module(monitor_job)
```
Wabri: suggestion(non-blocking):
```suggestion
path = rootpath / "monitor-openqa_job"
spec = importlib.util.spec_from_file_location(
    "monitor_job",
    path,
    loader=importlib.machinery.SourceFileLoader("monitor_job", str(path)),
)
assert spec is not None
monitor_job = importlib.util.module_from_spec(spec)
sys.modules["monitor_job"] = monitor_job
spec.loader.exec_module(monitor_job)
```
I'm not sure what are the version I like the most, but this is more minimal at least.

tests/test_monitor_openqa_job.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def test_logging(caplog: pytest.LogCaptureFixture) -> None:
+    with caplog.at_level(logging.DEBUG):
+        monitor_job.log.info("info")
+        monitor_job.log.warning("warn")
+        monitor_job.log.error("err")
+        monitor_job.log.debug("dbg")
+    assert all(x in caplog.text for x in ["info", "warn", "err", "dbg"])
```
Wabri: question(non-blocking): do we mind if this line fail we will not know what of those log are not what we expect?

tests/test_monitor_openqa_job.py
okurz: I included the comment as we have the application as standalone script, not a Python module so it's not quite that obvious IMHO

Martchus (approved):

tests/test_monitor_openqa_job.py
okurz: replaced by loop so we would have explicit failures showing which failed

tests/test_monitor_openqa_job.py
okurz: done

Wabri (approved):

PR #602, opened 2026-06-16
chore(Makefile): add help target and documentation

okurz:
Motivation:
Make it easier for developers to discover and understand the available build
and test targets in the Makefile.

Design Choices:
Same as in other projects implemented a help target that parses special
comments ('##') from the Makefile. Appended explanatory comments to all
user-facing targets.

Benefits:
Improves developer experience by providing self-documenting make targets via
the default help command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,11 @@ BPAN := .bpan
 #------------------------------------------------------------------------------
 # User targets
 #------------------------------------------------------------------------------
-default:
+.PHONY: help
+help: ## Show this help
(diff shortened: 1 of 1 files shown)


Martchus (approved):

Martchus:
"Fix missing" sounds like "Add" and with conventional commits it would probably be "feat: Add …".

perlpunk (commented):
This commit rather looks like adding comments to every target, and no actual change for the `clean` target itself. What am I missing?

okurz:
> This commit rather looks like adding comments to every target, and no actual change for the `clean` target itself. What am I missing?

ok, that was very much a different commit when I initially created it based on a different branch, sorry. Updated the commit message. The content is unchanged.

perlpunk (approved):

PR #603, opened 2026-06-16
Add script, service and timer to update certificates from OBS

Martchus:
These certificates are not published somewhere so we need to use `osc` to download them.

This change adds a systemd unit and timer that are supposed to be used in user mode. It also contains an install target for the systemd units. It is not suited for packaging and mainly to make it clear where the unit should be placed to use them locally. It is also useful for local testing:
```
$ make install-systemd-local DESTDIR=/tmp/install
$ cp -vR /tmp/install/etc/systemd/user/ ~/.config/systemd/
$ systemctl --user daemon-reload
```

Then the timer unit can be tested via:
```
$ systemctl --user start os-autoinst-scripts-update-factory-staging-cert.timer
```

The service units can be tested via:
```
$ systemctl --user start os-autoinst-scripts-update-cert-from-obs@factory-staging.service
$ journalctl --user -u os-autoinst-scripts-update-cert-from-obs@factory-staging.service
Starting Update a certificate from OBS...
certificate written to /home/martchus/dev/openqa/openqa/share/factory/other/fixed/openSUSE-Factory-Staging.crt
Finished Update a certificate from OBS.
```

```
$ systemctl --user start os-autoinst-scripts-update-cert-from-obs@suse-unsupported.service
$ journalctl --user -u os-autoinst-scripts-update-cert-from-obs@suse-unsupported.service
Starting Update a certificate from OBS...
certificate written to /home/martchus/dev/openqa/openqa/share/factory/other/fixed/suse-unsupported.crt
Finished Update a certificate from OBS.
```

This works on my system. On the production hosts we should probably create a dedicated user account with the required osc config and write permissions in the assets directory.

Related ticket: https://progress.opensuse.org/issues/200075

### DIFF
--- a/config/obs-certs/factory-staging.conf
+++ b/config/obs-certs/factory-staging.conf
@@ -1,0 +1,2 @@
+OBS_PROJECT=openSUSE:Factory:Staging
+CRT_NAME=openSUSE-Factory-Staging.crt
--- a/openqa-update-cert-from-obs
+++ b/openqa-update-cert-from-obs
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+
+config=${1:?Path to config file needs to be specified as first argument}
+# shellcheck source=config/obs-certs/factory-staging.conf
+source "$config"
+
+src=${OBS_PROJECT:?OBS_PROJECT needs to be set in $config}
+dst=${OPENQA_BASEDIR:-/var/lib}/openqa/share/factory/other/fixed
+dst_file=$dst/${CRT_NAME:?CRT_NAME needs to be set in $config}
+dst_file_tmp=$dst_file.tmp
+
+mkdir -p "$dst"
+osc --apiurl "${OBS_API_URL:-https://api.opensuse.org}" signkey --sslcert "$src" > "$dst_file_tmp"
+mv "$dst_file_tmp" "$dst_file"
+echo "certificate from '$src' written to '$dst_file'"
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,17 @@ clean: ## Clean up generated files
 	$(RM) -r .pytest_cache/
 	find . -name __pycache__ | xargs -r $(RM) -r
 
+install-systemd-local: ## Install contained systemd units for local use (not meant for packaging)
+	install -d -m 755 "$(DESTDIR)"/etc/systemd/system
+	for i in systemd/*.{service,timer}; do \
+		install -m 644 $$i "$(DESTDIR)"/etc/systemd/system ;\
+	done
+	install -d -m 755 "$(DESTDIR)"/etc/systemd/user
+	for i in systemd/user/*.{service,timer}; do \
+		install -m 644 $$i "$(DESTDIR)"/etc/systemd/user ;\
+	done
+	find "$(DESTDIR)"/etc/systemd -name '*.service' -exec sed -i -e "s|/opt/os-autoinst-scripts|$(PWD)|g" {} \+
+
 #------------------------------------------------------------------------------
 # Internal targets
 #------------------------------------------------------------------------------
--- a/README.md
+++ b/README.md
@@ -196,6 +196,39 @@ environment variables:
 * `from_email` - The From address for notification emails
 * `force_result` - If set to `1` tickets in the [tracker openqa-force-result](https://progress.opensuse.org/projects/openqav3/issues?query_id=700) can override job results
 
+## Update certificates from OBS
+The script `openqa-update-cert-from-obs` allows downloading the certificate of
+an OBS project. These certificates are so far used to run SecureBoot-enabled
+tests with images that are signed with them. The directory `config/obs-certs`
+contains configurations for this script.
+
+There are also systemd units/timers provided which can be installed locally. Run
+these commands with the user that should execute the updates regularly:
+
+```
+$ make install-systemd-local DESTDIR=/tmp/install
+$ cp -vR /tmp/install/etc/systemd/user/ ~/.config/systemd/
+$ systemctl --user daemon-reload
+```
+
+The user needs an osc configuration and write permissions under the openQA assets
+directory.
+
+The timer units can be enabled like that:
+```
+$ loginctl enable-linger
+$ systemctl --user enable --now os-autoinst-scripts-update-factory-staging-cert.timer
+```
+
+The service units can be started out of schedule like that:
+```
+$ systemctl --user start os-autoinst-scripts-update-cert-from-obs@factory-staging.service
+$ journalctl --user -u os-autoinst-scripts-update-cert-from-obs@factory-staging.service
+Starting Update a certificate from OBS...
+certificate from '…' written to '…/openqa/share/factory/other/fixed/openSUSE-Factory-Staging.crt'
+Finished Update a certificate from OBS.
+```
+
 ## Contribute
 
 This project lives in https://github.com/os-autoinst/os-autoinst-scripts
--- a/config/obs-certs/suse-unsupported.conf
+++ b/config/obs-certs/suse-unsupported.conf
@@ -1,0 +1,3 @@
+OBS_API_URL=https://api.suse.de
+OBS_PROJECT=SUSE:SLE-15-SP7:GA:Staging:A
+CRT_NAME=suse-unsupported.crt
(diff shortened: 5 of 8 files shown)


okurz (commented):
+1, needs CI fixed

okurz (approved):

config/obs-certs/factory-staging.conf
```diff
@@ -0,0 +1,2 @@
+OBS_PROJECT=openSUSE:Factory:Staging
+CRT_NAME=openSUSE-Factory-Staging.crt
```
foursixnine, line 2: @Vogtinator would `osc signkey --sslcert openSUSE:Factory` extract the signing key corresponding with `openSUSE-UEFI-CA-Certificate-2048.crt`?

Subjects are slightly different:

```
foursixnine@deimos:ovmf> openssl x509 -in openSUSE-UEFI-CA-Certificate-2048.crt --text -noout  | grep Subject:
        Subject: CN=openSUSE Secure Boot CA, C=DE, L=Nuremberg, O=openSUSE Project, emailAddress=build@opensuse.org

foursixnine@deimos:ovmf> openssl x509 -in openSUSE-UEFI-CA-Certificate-2048-Factory.crt --text -noout  | grep Subject:
        Subject: CN=openSUSE Secure Boot Signkey, C=DE, L=Nuremberg, O=openSUSE Project, emailAddress=build@opensuse.org
```

openqa-update-cert-from-obs
```diff
@@ -0,0 +1,16 @@
...
+
+src=${OBS_PROJECT:?OBS_PROJECT needs to be set in $config}
+dst=${OPENQA_BASEDIR:-/var/lib}/openqa/share/factory/other/fixed
+dst_file=$dst/${CRT_NAME:?CRT_NAME needs to be set in $config}
+dst_file_tmp=$dst_file.tmp
+
+mkdir -p "$dst"
+osc --apiurl "${OBS_API_URL:-https://api.opensuse.org}" signkey --sslcert "$src" > "$dst_file_tmp"
+mv "$dst_file_tmp" "$dst_file"
+echo "certificate written to $dst_file"
```
foursixnine:
```suggestion
echo "certificate from $src written to $dst_file"
```

config/obs-certs/factory-staging.conf
Martchus: I compared the sha256 sum of the downloaded Factory cert with the one from @Vogtinator's OBS repo and they were identical:

```
sha256sum /home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/openSUSE-Factory-Staging.crt ~/dev/openqa/openqa/share/factory/other/fixed/openSUSE-Factory-Staging.crt 
8f6e4c3c00b5914f1f923e0428e3e85898640763141ba5c6f6fbcafff224ae54  /home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/openSUSE-Factory-Staging.crt
8f6e4c3c00b5914f1f923e0428e3e85898640763141ba5c6f6fbcafff224ae54  /home/martchus/dev/openqa/openqa/share/factory/other/fixed/openSUSE-Factory-Staging.crt
```

I also compared the sha256 sum of the unsupported cert and there I got a mismatch but the cert itself nevertheless seems the same. ~~(Not exactly sure what is different so that the sha256 sum is not identical.)~~ The only difference is the trailing newline:

```
colordiff -Naur /home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/suse-unsupported.crt ~/dev/openqa/openqa/share/factory/other/fixed/suse-unsupported.crt 
--- /home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/suse-unsupported.crt   2025-06-26 14:54:20.000000000 +0200
+++ /home/martchus/dev/openqa/openqa/share/factory/other/fixed/suse-unsupported.crt     2026-06-16 13:10:12.777097033 +0200
@@ -29,3 +29,4 @@
 J2J72BOfBqySVqZT6ItH7dqTLamvXjwSdgq+ssc/n9Phg1DONxsWqiuByf109ZRy
 UpCCbyQ3se2NCGZJjG27MFi3Ir8EXA/LCwUY
 -----END CERTIFICATE-----
+
```

foursixnine: @Martchus there are two extra openSUSE certs:
- `openSUSE-Factory-Staging.crt` from `openSUSE:Factory:Staging` which this file references.
- `openSUSE-UEFI-CA-Certificate-2048.crt` which I'd like to confirm if it would be the same from `openSUSE:Factory` or comes from somewhere else

Vogtinator: Yes, it should. Are the fingerprints ok?

Martchus: `openSUSE-Factory-Staging.crt` and `openSUSE-UEFI-CA-Certificate-2048.crt` look like different certs but we probably only care about the former. The latter is not needed here because it is part of the firmware vars files provided by the regular `qemu-ovmf-x86_64` package anyway (so there is no fork of that package or dynamic enrollment is required to have it).

> @Vogtinator would osc signkey --sslcert openSUSE:Factory extract the signing key corresponding with openSUSE-UEFI-CA-Certificate-2048.crt?

So the answer to that question is: no
It extracts the signing key corresponding to `openSUSE-Factory-Staging.crt` (which is the one we are actually missing in the regular `qemu-ovmf-x86_64` package).

Vogtinator: ... actually, no - the CA should trust the SB signkey but is not used in OBS directly.

In practice, for TW both should work, but it makes more sense to test with the CA.

Vogtinator: > The latter is not needed here because it is part of the firmware vars files provided by the regular qemu-ovmf-x86_64 package anyway (so there is no fork of that package or dynamic enrollment is required to have it).

Correct, unless you try to run tests on a different distro that doesn't have such specific efivar volumes directly.

> It extracts the signing key corresponding to openSUSE-Factory-Staging.crt (which is the one we are actually missing in the regular qemu-ovmf-x86_64 package).

openSUSE:Factory != openSUSE:Factory:Staging

foursixnine: > In practice, for TW both should work, but it makes more sense to test with the CA.

How do we get the one with the CA? (maybe answered V?)

>>    qemu-ovmf-x86_64 package anyway (so there is no fork of that package or dynamic enrollment is required to have it).

> Correct, unless you try to run tests on a different distro that doesn't have such specific efivar volumes directly.

So the CA comes in the ovmf package?

Vogtinator: > How do we get the one with the CA? (maybe answered V?)

Originally? No idea. For the SLES one it's on https://www.suse.com/support/security/keys/, for openSUSE I couldn't find it with a cursory search.

> So the CA comes in the ovmf package?

Yes. Can probably be extracted from some signed files like shim as well which contain the whole trust chain including the issuer certificate. Probably not a good way though.

Martchus: > How do we get the one with the CA?

I don't know. I just extracted it from via `virt-fw-vars --input /usr/share/qemu/ovmf-x86_64-smm-opensuse-vars.bin --extract-certs` from the vars provided by `qemu-ovmf-x86_64`. That's how I know it is present in this vars file anyway and doesn't need to be added dynamically for our purposes.

I ran `osc signkey --sslcert openSUSE:Factory` to see whether it might correspond to that certificate but no; that's a different one (even though it also has `CN=openSUSE Secure Boot CA`). I also ran `osc signkey --sslcert openSUSE:Tumbleweed` but that's yet another cert.

Martchus: But does it really matter? The script in this PR will download the certs that we so far only had via @Vogtinator's custom package and I verified that they are indeed the same.

And btw, the script is extendable. So if we need a third cert in the future from some other OBS/IBS project we can easily add it (as needed, a plain HTTPs download is probably still preferable if possible).

foursixnine: > But does it really matter?

It does, but not for this PR.

> And btw, the script is extendable. So if we need a third cert in the future from some other OBS/IBS project we can easily add it (as needed, a plain HTTPs download is probably still preferable if possible

Yeah, where to get the certs from is what I'm trying to find out :)

foursixnine (approved):

PR #605, opened 2026-06-16
refactor: rewrite cert updater script to Python

okurz:
Motivation:
Provide a robust command-line interface with proper --help documentation for
the certificate updating script.

Design Choices:
Port the bash script to Python using typer for argument parsing. Relocate
installation and usage documentation from README.md to the typer --help
block. Add unit tests with 100% statement and branch coverage.

Benefits:
Easier maintenance, better command-line documentation, and rigorous test
coverage.

After:
* #603

### DIFF
--- a/openqa-update-cert-from-obs
+++ b/openqa-update-cert-from-obs
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+
+config=${1:?Path to config file needs to be specified as first argument}
+# shellcheck source=config/obs-certs/factory-staging.conf
+source "$config"
+
+src=${OBS_PROJECT:?OBS_PROJECT needs to be set in $config}
+dst=${OPENQA_BASEDIR:-/var/lib}/openqa/share/factory/other/fixed
+dst_file=$dst/${CRT_NAME:?CRT_NAME needs to be set in $config}
+dst_file_tmp=$dst_file.tmp
+
+mkdir -p "$dst"
+osc --apiurl "${OBS_API_URL:-https://api.opensuse.org}" signkey --sslcert "$src" > "$dst_file_tmp"
+mv "$dst_file_tmp" "$dst_file"
+echo "certificate written to $dst_file"
--- a/openqa-update-cert-from-obs
+++ b/openqa-update-cert-from-obs
@@ -1,16 +1,152 @@
-#!/bin/bash
-set -e
+#!/usr/bin/env python3
+# Copyright SUSE LLC
+# ruff: noqa: D103, S404, S603, T201
+"""Update certificates from OBS."""
 
-config=${1:?Path to config file needs to be specified as first argument}
-# shellcheck source=config/obs-certs/factory-staging.conf
-source "$config"
+import pathlib
+import shlex
+import subprocess
+import sys
+from typing import Annotated
 
-src=${OBS_PROJECT:?OBS_PROJECT needs to be set in $config}
-dst=${OPENQA_BASEDIR:-/var/lib}/openqa/share/factory/other/fixed
-dst_file=$dst/${CRT_NAME:?CRT_NAME needs to be set in $config}
-dst_file_tmp=$dst_file.tmp
+import typer
 
-mkdir -p "$dst"
-osc --apiurl "${OBS_API_URL:-https://api.opensuse.org}" signkey --sslcert "$src" > "$dst_file_tmp"
-mv "$dst_file_tmp" "$dst_file"
-echo "certificate written to $dst_file"
+app = typer.Typer(add_completion=False)
+
+
+def parse_config(config_path: pathlib.Path) -> dict[str, str]:
+    config_vars = {}
+    try:
+        with config_path.open("r", encoding="utf-8") as f:
+            lines = f.readlines()
+    except Exception as e:
+        print(f"Error reading config file {config_path}: {e}", file=sys.stderr)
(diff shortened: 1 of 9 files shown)


systemd/user/os-autoinst-scripts-update-cert-from-obs@.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update a certificate from OBS
+After=network-online.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/bash /opt/os-autoinst-scripts/openqa-update-staging-cert /opt/os-autoinst-scripts/config/obs-certs/%i.conf
```
Wabri: issue(blocking): 
```suggestion
ExecStart=/usr/bin/bash /opt/os-autoinst-scripts/openqa-update-staging-cert-from-obs /opt/os-autoinst-scripts/config/obs-certs/%i.conf
```

systemd/user/os-autoinst-scripts-update-cert-from-obs@.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update a certificate from OBS
+After=network-online.target
```
Wabri, line 3: issue(blocking): I think `After` won't work without `Wants`

systemd/user/os-autoinst-scripts-update-cert-from-obs@.service
```diff
@@ -0,0 +1,9 @@
...
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/bash /opt/os-autoinst-scripts/openqa-update-staging-cert /opt/os-autoinst-scripts/config/obs-certs/%i.conf
+Restart=on-failure
+RestartSec=1m
```
Wabri: question: Does this means that if the script is failing it will still run every 1min?

Wabri (changes requested):

systemd/user/os-autoinst-scripts-update-cert-from-obs@.service
Martchus: Yes - maybe we should increase that.

Martchus:
I don't think the additional 400 lines and additional dependencies are justified here. Let's merge my PR first and see how it works in production.

systemd/user/os-autoinst-scripts-update-cert-from-obs@.service
Martchus: I can update my PR - which I'd like to merge first - accordingly.

systemd/user/os-autoinst-scripts-update-cert-from-obs@.service
Martchus: That's fixed in my version so this PR is probably not based on the current version of my PR.

Martchus:
@Wabri I implemented your suggestions in https://github.com/os-autoinst/os-autoinst-scripts/pull/603.

systemd/user/os-autoinst-scripts-update-cert-from-obs@.service
okurz: fixed

okurz:
> I don't think the additional 400 lines and additional dependencies are justified here. Let's merge my PR first and see how it works in production.

To be fair it's now +152,-16 for mostly better user experience, help text and CLI parsing. I incorporated your updates to the original script. Please re-review

openqa-update-cert-from-obs
```diff
@@ -1,16 +1,152 @@
...
+import subprocess
+import sys
+from typing import Annotated
+
+import typer
+
+app = typer.Typer(add_completion=False)
+
+
+def parse_config(config_path: pathlib.Path) -> dict[str, str]:
```
asmorodskyi, line 44: I think this function can be fully replaced by `configparser` . WDYT ?

openqa-update-cert-from-obs
```diff
@@ -1,16 +1,152 @@
...
+        "signkey",
+        "--sslcert",
+        obs_project,
+    ]
+
+    try:
+        with dst_file_tmp.open("w", encoding="utf-8") as f_tmp:
+            subprocess.run(cmd, check=True, stdout=f_tmp, stderr=subprocess.PIPE, text=True)
+    except subprocess.CalledProcessError as e:
+        if dst_file_tmp.exists():
```
asmorodskyi: isn't `with` handling all this out of the box ? looks like code which we can drop

asmorodskyi:
> > I don't think the additional 400 lines and additional dependencies are justified here. Let's merge my PR first and see how it works in production.
> 
> To be fair it's now +152,-16 for mostly better user experience, help text and CLI parsing. I incorporated your updates to the original script. Please re-review

and seems like we can make it even smaller :)

openqa-update-cert-from-obs
okurz: Yes, but only if we replace the config file format. I am adressing that in a separate PR
* #664

okurz:
> > I don't think the additional 400 lines and additional dependencies are justified here. Let's merge my PR first and see how it works in production.
> 
> To be fair it's now +152,-16 for mostly better user experience, help text and CLI parsing. I incorporated your updates to the original script. Please re-review

I refactored so the main script is now +121,-16.

PR #608, opened 2026-06-17
feat: Add a python script to rotate Project Access Token in GitLab

gauravpathak:
Ticket: https://progress.opensuse.org/issues/201372

Add a python script to automatically rotate Project Access Token in GitLab and update the CI/CD variable using the token for pushing the changes to repository to ensure automated processes do not break due to expired token.

### DIFF
--- a/rotate-gitlab-token.py
+++ b/rotate-gitlab-token.py
@@ -1,0 +1,125 @@
+#!/usr/bin/python3
+# Copyright SUSE LLC
+"""Automate the rotation of GitLab Project Access Tokens and update CI/CD variables.
+
+This script checks the expiration date of a specified GitLab Project Access Token.
+If the token is scheduled to expire in 1 day or less, the script rotates the token,
+extends its validity by 364 days (with Maintainer access level 40), and automatically
+updates the corresponding CI/CD variable with the newly generated token value.
+
+It is designed to be executed within a GitLab CI/CD pipeline to ensure automated
+processes do not break due to expired credentials.
+
+Environment Variables:
+    CI_SERVER_PROTOCOL : Protocol for the GitLab instance (default: 'https').
+    CI_SERVER_HOST     : Hostname of the GitLab instance (default: 'gitlab.suse.de').
+    CI_PROJECT_ID      : The target GitLab project ID (default: '13758').
+    CI_PUSH_TOKEN      : The default CI variable containing the current active token.
+
+Command-line Arguments:
+    -t, --token-name      : (Required) The name of the Project Access Token to monitor.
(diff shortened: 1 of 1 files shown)


rotate-gitlab-token.py
```diff
@@ -0,0 +1,120 @@
...
+        if token.name == tok_name and token.active:
+            token_id = token.id
+            break
+    return token_id
+
+
+def main() -> None:
+    """Run the main execution of the script."""
+    parser = argparse.ArgumentParser(description="Parse CLI arguments supplies to this script")
+    parser.add_argument("-t", "--tokname", type=str, required=True, default=None,
```
okurz:
```suggestion
    parser.add_argument("-t", "--token-name", type=str, required=True, default=None,
```

rotate-gitlab-token.py
```diff
@@ -0,0 +1,120 @@
...
+                        help="Project Access Token Name (default: None)")
+    parser.add_argument("-c", "--civarname", type=str, default=None,
```
okurz:
```suggestion
    parser.add_argument("-c", "--ci-var-name", type=str, default=None,
```

rotate-gitlab-token.py
```diff
@@ -0,0 +1,120 @@
...
+    for token in access_tokens:
+        if token.name == tok_name and token.active:
+            token_id = token.id
+            break
+    return token_id
+
+
+def main() -> None:
+    """Run the main execution of the script."""
+    parser = argparse.ArgumentParser(description="Parse CLI arguments supplies to this script")
```
okurz: this description should be a description of the application, not (just) for ArgumentParser. You can print the doc-string of the file. See https://github.com/os-autoinst/openqa_review/blob/master/openqa_review/openqa_review.py#L1545

But also I suggest to use `typer`. See https://github.com/os-autoinst/os-autoinst-scripts/blob/master/monitor-openqa_job#L193

okurz (changes requested):

rotate-gitlab-token.py
gauravpathak: I like the first one. It's simple.

rotate-gitlab-token.py
gauravpathak: Tried using this suggestion locally, but it is not working
```
    proj_access_tok_key = args.token-name
                          ^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'token'
```
Underscore is working fine.

okurz: yes, if you define `--token-name` then the variable is "magically" called `args.token_name` :)

okurz (approved):

Wabri (approved):

PR #609, opened 2026-06-17
feat: Add openqa-get-qem-bot-blocking-jobs

okurz:
Motivation:
Provide a CLI utility to find blocking SLE maintenance jobs in openQA. It
parses GitLab CI logs as the source of truth for qem-bot's rejection decisions.

Design Choices:
Implemented using Python and Typer. It fetches and parses raw GitLab CI traces
rather than relying on high-level dashboards which lack rejection details.

Benefits:
Ensures a reliable, detailed, and consistent view of active blocking jobs.

### DIFF
--- a/openqa-get-qem-bot-blocking-jobs
+++ b/openqa-get-qem-bot-blocking-jobs
@@ -1,0 +1,165 @@
+#!/usr/bin/env python3
+# Copyright SUSE LLC
+# ruff: noqa: PLR0913, PLR0917, T201, S107
+"""Find blocking SLE maintenance jobs based on parsing gitlab CI job logs of qem-bot.
+
+This script parses the gitlab CI job logs of qem-bot directly because they serve as
+the absolute source of truth for why qem-bot decides not to approve a submission,
+which cannot be determined by simply querying the high-level dashboard.
+"""
+
+import logging
+import re
+import sys
+from pathlib import Path
+from typing import Annotated, Any
+
+import httpx
+import typer
+
+app = typer.Typer(add_completion=False)
+log = logging.getLogger(__name__)
+
+
+def setup_logging(verbose: int) -> None:
+    """Set up logging level based on verbosity count."""
+    verbose_to_level = {0: logging.WARNING, 1: logging.INFO}
+    level = verbose_to_level.get(verbose, logging.DEBUG)
+    logging.basicConfig(level=level, format="%(levelname)s: %(message)s", stream=sys.stderr, force=True)
+
+
+def get_gitlab_access_token(filename: str) -> str:
+    """Read the gitlab access token from the given filename."""
+    token_path = Path(filename).expanduser()
+    log.debug("Reading GitLab access token from %s", token_path)
+    try:
+        return token_path.read_text(encoding="utf-8").strip()
+    except Exception as e:
+        log.exception("Error reading gitlab access token file '%s'", filename)
+        raise typer.Exit(code=1) from e
+
+
+def get_gitlab_jobs(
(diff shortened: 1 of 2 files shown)


Martchus:
What sets this apart from the dashboard (besides being a CLI utility)?

okurz:
> What sets this apart from the dashboard (besides being a CLI utility)?

`openqa-get-qem-blocking-jobs` lists the jobs where the specified job groups are the primary blockers. This can be helpful for reviewers in testing squads to focus on the submissions where *their* group is blocking.

openqa-get-qem-bot-blocking-jobs
```diff
@@ -0,0 +1,155 @@
...
+    response = client.get(jobs_url, headers=headers)
+    response.raise_for_status()
+    return response.json()
+
+
+def get_gitlab_job_raw(
+    client: httpx.Client, gitlab_url: str, project_id: int, job_id: int, headers: dict[str, str]
+) -> str:
+    """Fetch the raw trace log of a GitLab job via the API."""
+    job_raw_url = f"{gitlab_url}/api/v4/projects/{project_id}/jobs/{job_id}/trace"
```
kalikiana, line 58: I like the idea... but I would strongly prefer to query the dashboard, and make sure we always produce the same results. Otherwise this will lead to confusion about what is "the blockers".

openqa-get-qem-bot-blocking-jobs
```diff
@@ -0,0 +1,155 @@
...
+
+@app.command()
+def main(
+    project_id: Annotated[int, typer.Option(help="GitLab project ID")] = 6096,
+    job_name: Annotated[str, typer.Option(help="Name of the GitLab job to filter")] = "approve submissions",
+    gitlab_url: Annotated[str, typer.Option(help="GitLab API URL")] = "https://gitlab.suse.de",
+    openqa_url: Annotated[str, typer.Option(help="openQA API URL")] = "https://openqa.suse.de",
+    filter_regex: Annotated[
+        str, typer.Option(help="Regex to filter openQA job groups")
+    ] = r"(Container|Cloud|SLEM|Minimal.*VM)",
```
kalikiana, line 136: Also having the package name filter would be nice. Feel free to keep it simple in the first version, tho.

kalikiana (changes requested):

openqa-get-qem-bot-blocking-jobs
okurz: > I like the idea... but I would strongly prefer to query the dashboard, and make sure we always produce the same results. Otherwise this will lead to confusion about what is "the blockers".

The point is to look at the source of truth of why qem-bot does not approve so it can't be the dashboard

openqa-get-qem-bot-blocking-jobs
okurz: what "package name filter"?

openqa-get-qem-bot-blocking-jobs
Martchus: That point should be stated as motivation then.

okurz: alright. Done. I updated the doc string, commit message and PR description

openqa-get-qem-bot-blocking-jobs
okurz: I think you assume that the "package name" would actually be available from logs which is not the case. Nevertheless I created https://github.com/os-autoinst/os-autoinst-scripts/pull/622 to add a "submission filter option"

okurz:
adressed all comments

Martchus (approved):

kalikiana (approved):

PR #610, opened 2026-06-17
fix: Fallback ci_var_name to global default if not supplied via CLI

gauravpathak:


rotate-gitlab-token.py
```diff
@@ -85,6 +85,8 @@ def main() -> None:
     elif args.ci_var_name is not None:
         ci_push_tok = os.getenv(args.ci_var_name)
         ci_var_name = args.ci_var_name
+    else:
+        ci_var_name = CI_VAR_KEY
```
perlpunk: Why not just define the default in the `add_argument` call?
```
parser.add_argument("-c", "--ci-var-name", type=str, default=CI_VAR_KEY,
```

perlpunk (approved):

Martchus (approved):

PR #611, opened 2026-06-18
feat(token-rotation): automate CI scheduling

gauravpathak:
Ticket: https://progress.opensuse.org/issues/201372

- Add `create_or_update_ci_pipeline` to auto-schedule the next token rotation 2 days prior to expiration
- Update existing pipeline schedules or create a new one on the  `main` branch
- Secure GitLab API client instantiations by setting `ssl_verify=True`
- Parse token expiration using `date.fromisoformat` to accurately  calculate the required cron schedule

### DIFF
--- a/rotate-gitlab-token.py
+++ b/rotate-gitlab-token.py
@@ -32,6 +32,8 @@ from gitlab.v4.objects import Project
 GITLAB_URL = os.getenv("CI_SERVER_PROTOCOL", "https") + "://" + os.getenv("CI_SERVER_HOST", "gitlab.suse.de")
 PROJECT_ID = os.getenv("CI_PROJECT_ID", "13758")
 CI_VAR_KEY = "CI_PUSH_TOKEN"
+CI_ROTATE_SCHED_DESC = "Rotate_Proj_Access_Token"
+DAYS_UNTIL_TOKEN_ROTATION = 21
 
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
What's "token-rotationa"? Does it work to say "feat(rotate-gitlab-token): automate CI scheduling"?

perlpunk (approved):

rotate-gitlab-token.py
```diff
@@ -32,6 +32,8 @@
 GITLAB_URL = os.getenv("CI_SERVER_PROTOCOL", "https") + "://" + os.getenv("CI_SERVER_HOST", "gitlab.suse.de")
 PROJECT_ID = os.getenv("CI_PROJECT_ID", "13758")
 CI_VAR_KEY = "CI_PUSH_TOKEN"
+CI_ROTATE_SCHED_DESC = "Rotate_Proj_Access_Token"
+DAYS_UNTIL_TOKEN_ROTATION = 7
```
okurz: I suggested 21. Any good reason to do 7 instead? Why 21? Because then we can run the script just once weekly and if it fails twice then at least we have a third chance to renew before a token expires. Also that means on problems we have more than a week to fix it meaning that we can wait for people to come back from holiday, etc.

okurz (changes requested):

okurz (approved):

PR #612, opened 2026-06-18
feat: rewrite openqa-label-known-issues in Python

okurz:
Motivation:
The original bash script was becoming unmaintainable, hard to extend, and tricky
to test due to complex regex and common dependency sourcing.

Design Choices:
Ported the entire logic to Python using httpx and typer. Added a comprehensive
pytest-based test suite with 100% statement and branch coverage.

Benefits:
The script is now highly maintainable, type-safe, has rich logging support, and
a rigorous test harness preventing regressions.

Related issue: https://progress.opensuse.org/issues/202680


openqa-label-known-issues
```diff
@@ -1,223 +1,622 @@
...
+    body = (
+        f"--{boundary}\n"
+        'Content-Type: text/plain; charset="utf-8"\n'
+        "Content-Transfer-Encoding: quoted-printable\n\n"
+        f"{plain}\n\n"
+        f"--{boundary}\n"
+        'Content-Type: text/html; charset="utf-8"\n'
+        "Content-Transfer-Encoding: quoted-printable\n\n"
+        f"<html><body>\n{md}</body></html>\n\n"
+        f"--{boundary}--"
```
Martchus: Is the original code also making the HTTP body in such a rather hacky way? If not I suggest we use a library.

okurz: yes, comes from https://github.com/okurz/os-autoinst-scripts/blob/feature/openqa_auto_review_python/_common#L247
now using email.message.EmailMessage

Martchus (approved):

nicksinger (approved):
Blindly approving, I assume most of this is written by an LLM (we don't disclose it, so I have to assume) so good luck to all of us. Some functional tests would be helpful here but I guess we have to do a live test in production

PR #613, opened 2026-06-19
feat: add retry limit to auto-review feature

okurz:
Motivation:
Prevent infinite restart loops of failing openQA jobs that are matched by
the auto-review feature but scheduled in a way where they always fail.

Design Choices:
Parses optional custom retry limits per-ticket (e.g. :retry:3) from the
subject. Traverses parent clones backwards to count previous retries, with
a command-line option `--retry-limit` to customize the default of 7.

Benefits:
Stops runaway job retriggering, saves hardware resources, and improves
general openQA test automation stability.

Related issue: https://progress.opensuse.org/issues/202680

After:
* #612

### DIFF
--- a/openqa-label-known-issues
+++ b/openqa-label-known-issues
@@ -189,14 +189,49 @@ def label_on_issue(
     return True
 
 
+def get_default_retry_limit(default_limit: int = 7) -> int:
+    env_limit = os.environ.get("auto_review_retry_limit")
+    if env_limit is not None:
+        try:
+            return int(env_limit)
+        except ValueError:
+            pass
+    return default_limit
+
+
+def count_restarts(job: dict, client_call: list[str]) -> int:
+    restarts = 0
+    current_job = job
+    while restarts < 100:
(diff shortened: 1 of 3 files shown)


openqa-label-known-issues
```diff
@@ -189,14 +189,49 @@ def label_on_issue(
...
+    restarts = 0
+    current_job = job
+    while restarts < 100:
+        cloned_from = current_job.get("cloned_from") or current_job.get("settings", {}).get("cloned_from")
+        if not cloned_from:
+            break
+        restarts += 1
+        current_id = str(cloned_from)
+        cmd_job = (
+            ["openqa-cli"] + [arg for arg in client_call if arg not in {"echo", "openqa-cli"}] + [f"jobs/{current_id}"]
```
perlpunk: That doesn't look very nice.
Now that we have python instead of bash, maybe we can make use of a better data structures or even an object for the client_call / dry_run thing.

okurz: Ok, done with a class "OpenQAClient"

Martchus (approved):

perlpunk (approved):

PR #614, opened 2026-06-22
feat(monitor): Indicate failed jobs via artifact/file

Martchus:
* Avoid duplication of `raise typer.Exit(code=1)`
* Avoid deciding the exit code in `comment_on_failed_jobs` which is not
  intuitive as the function name sounds like the function is only supposed
  to add comments (and only call e.g. `raise typer.Exit(code=1)` if *that*
  fails)
* Avoid that the monitoring pipeline fails (causing alerts) in case the
  openQA jobs being monitored fail (as we receive notifications for those
  anyway)
* Write the file `job_post_skip_submission` which is then picked up by
  Jenkins as artifact and made available in the submission pipeline that
  will then skip the submission

Related ticket: https://progress.opensuse.org/issues/202467

### DIFF
--- a/monitor-openqa_job
+++ b/monitor-openqa_job
@@ -222,9 +222,9 @@ app = typer.Typer(add_completion=False)
 Monitor an openQA job by polling the status of a job over the API.
 
 Continuously polls the openQA API for the status of the specified jobs until
-all jobs finish. After the jobs have finished this program exits with an exit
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

kalikiana (approved):
Neat

PR #621, opened 2026-07-03
style(*): improve python style checks

Wabri:
- add a new make target with `uv sync`
- change `[project.optional-dependencies]` to `[dependency-groups]` for better uv compatibility (see https://peps.python.org/pep-0735/#rationale)
- uv.lock must not be ignored in the repository
- fix ruff docstring warning in `openqa-trigger-bisect-jobs`

Reference: https://progress.opensuse.org/issues/191575

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ help: ## Show this help
 
 default: help
 
+# Override to use uv, e.g. PYTHON_RUN="uv run"
+PYTHON_RUN ?=
+
 .PHONY: test
 ifeq ($(CHECKSTYLE),0)
 checkstyle_tests =
@@ -39,7 +42,7 @@ test-bash: $(BPAN) ## Run bash tests
 	"${PROVE}" -r $(if $v,-v )$(test)
 
 test-python: ## Run python tests
-	py.test tests
+	$(PYTHON_RUN) py.test tests
 
 test-online: ## Run tests that require online connection
 	dry_run=1 bash -x ./openqa-label-known-issues-multi < ./tests/incompletes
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,10 +21,18 @@ BPAN := .bpan
 #------------------------------------------------------------------------------
 .PHONY: help
 help: ## Show this help
-	@grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+	@echo "Call one of the available targets:"
+	@sed -n 's/\(^[^.#[:space:]A-Z]*\):.*$$/\1/p' Makefile | uniq
```
okurz: I don't see the relation to "improve python style checks". At least put those changes into separate commits but likely better to put those into separate PRs

.gitignore
```diff
@@ -2,7 +2,7 @@
 /tests/__pycache__/
 /__pycache__/
 .*.swp
-/uv.lock
 test-more-bash/
 bpan/
+uv.lock
```
okurz: you change the ignore rules but also don't include uv.lock. Without a process for automatic update I suggest to *not* include this file but then you should also not change this

Makefile
```diff
@@ -21,10 +21,18 @@ BPAN := .bpan
...
 
 default: help
 
+.PHONY: init
+init: ## Initialize development environment
+	uv sync
+
```
okurz: See related comment in .gitignore

okurz (changes requested):

okurz (approved):

Martchus (approved):

PR #630, opened 2026-07-16
Update `openqa-search-maintenance-core-jobs` to latest changes in openQA job groups

manfredi:
- removes no longer existing groups
- Improve build_check searching results by grep perl-regexp


okurz (changes requested):
Please be more specific in the commit message. Right now it says "Update the script". The commit message should say which script

okurz (approved):

perlpunk (approved):

PR #632, opened 2026-07-17
fix: Render multi-line skip reasons as an indented block

okurz:
Motivation: A multi-line job_post_skip_submission (e.g. a note followed by a
package list) looked misaligned when appended after "reason: " on one line.

Design Choices: Add _log_skip_reason which keeps single-line reasons inline
and prints multi-line reasons on following lines, each indented by two
spaces.

Benefits: Multi-line skip reasons read as a clear block instead of wrapping
awkwardly after the "reason:" label.

Related issue: https://progress.opensuse.org/issues/201843

### DIFF
--- a/test/04-common.t
+++ b/test/04-common.t
@@ -2,7 +2,7 @@
 
 source test/init
 
-plan tests 21
+plan tests 27
 
 source ./_common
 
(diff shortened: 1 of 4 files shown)


asmorodskyi (approved):

test/04-common.t
```diff
@@ -64,7 +64,6 @@ somecommand() {
 
 try runcli somecommand
 like "$got" "somecommand.*>>>STDERR<<<.*STDOUT" "somecommand stdout and stderr"
-
```
perlpunk, line 67: I think we should keep the blank line between the tests

okurz: done

Wabri (approved):

PR #634, opened 2026-07-17
fix: fail hard early when jq command is missing

okurz:
Motivation:
When the required jq dependency is missing, the scripts continue running
and fail silently or produce confusing follow-up errors.

Design Choices:
Added a command availability check for jq inside _common that exits with 1
upon failure. Sourced _common via local path in tests to avoid PATH shadowing.

Benefits:
Scripts fail hard immediately with a clear error message, speeding up
investigation of missing environment dependencies.

### DIFF
--- a/_common
+++ b/_common
@@ -10,6 +10,13 @@ submit_target=${submit_target:-openSUSE:Factory}
 
 warn() { printf '%s\n' "$@" >&2; }
 
+for cmd in jq retry osc openqa-cli; do
+    if ! command -v "$cmd" > /dev/null 2>&1; then
+        warn "ERROR: '$cmd' command not found. Please install $cmd."
+        exit 1
+    fi
+done
+
 enable_force_result=${enable_force_result:-false}
 
 client_call=()
--- a/test/04-common.t
+++ b/test/04-common.t
@@ -2,9 +2,9 @@
 
 source test/init
 
-plan tests 13
+plan tests 21
 
-source _common
+source ./_common
 
 success() {
     echo "SUCCESS $@"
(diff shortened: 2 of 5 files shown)


_common
```diff
@@ -10,6 +10,13 @@ submit_target=${submit_target:-openSUSE:Factory}
 
 warn() { printf '%s\n' "$@" >&2; }
 
+for cmd in jq retry; do
```
gauravpathak:
```suggestion
for cmd in jq retry osc openqa-cli; do
```

okurz: done

test/init
```diff
@@ -5,3 +5,13 @@ export BPAN_ROOT=$PWD/.bpan
 source "$BPAN_ROOT/test/init"
 
 dir=$PWD/test
+
+# Mock commands that may be missing in testing environments
+command() {
+    if [[ "$1" == "-v" ]]; then
```
perlpunk, line 11: Quotes are unnecessary in `[[ ]]` but as our new rules say, style comments can be ignored

test/04-common.t
```diff
@@ -64,3 +64,19 @@ somecommand() {
 
 try runcli somecommand
 like "$got" "somecommand.*>>>STDERR<<<.*STDOUT" "somecommand stdout and stderr"
+
+try "bash -c 'command() { if [[ \"\$1\" == \"-v\" && \"\$2\" == \"jq\" ]]; then return 1; fi; if [[ \"\$1\" == \"-v\" && ( \"\$2\" == \"retry\" || \"\$2\" == \"osc\" || \"\$2\" == \"openqa-cli\" ) ]]; then return 0; fi; builtin command \"\$@\"; }; export -f command; source ./_common'"
```
perlpunk, line 68:
```suggestion
try "bash -c 'command() { if [[ \$1 == -v && \$2 == jq ]]; then return 1; fi; if [[ \$1 == -v && ( \$2 == retry || \$2 == osc || \$2 == openqa-cli ) ]]; then return 0; fi; builtin command \"\$@\"; }; export -f command; source ./_common'"
```

test/init
```diff
@@ -5,3 +5,13 @@ export BPAN_ROOT=$PWD/.bpan
 source "$BPAN_ROOT/test/init"
 
 dir=$PWD/test
+
+# Mock commands that may be missing in testing environments
```
perlpunk, line 9: I don't completely understand this. We should have all commands available during testing, no?

test/init
okurz: >  but as our new rules say, style comments can be ignored

... and we implement automatic checks to ensure a better style immediately :)

test/init
okurz: Not in CI. Tests passed for me locally but then failed in GHA as those external applications are not installed there

test/init
perlpunk: > ... and we implement automatic checks to ensure a better style immediately :)

There you go: #637 :)

perlpunk (approved):

gauravpathak (approved):

PR #637, opened 2026-07-17
style: Simplify shell scripts by removing superfluous quotes

perlpunk:
Add --simplify to shfmt call.

This will remove superfluous quotes inside `[[ ... ]]` for example.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,11 +53,11 @@ test-online: ## Run tests that require online connection
 checkstyle: test-shellcheck test-yaml checkstyle-python check-code-health test-gitlint ## Run all style checks
 
 shfmt: ## Format shell scripts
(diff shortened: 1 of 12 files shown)


perlpunk:
Ironically, the stylecheck now fails. Maybe there are different versions in CI and in my local Leap 16?

edit: yes, 3.12.0-1.6 in Tumbleweed vs. 3.5.1-bp160.1.13 in Leap

okurz (approved):

Wabri (approved):

perlpunk:
In case my last comment wasn't clear, we can't really use this (yet), as the behaviour between Leap 16 and Tumbleweed is different, and I did not find any option to influence this

PR #642, opened 2026-07-22
refactor: Use openqa-label-known-issues script for single url

perlpunk:
openqa-label-known-issues-hook is used in minion jobs, but it only calls /openqa-label-known-issues-multi for one job, so it makes sense to use the script intended for just one job.

### DIFF
--- a/openqa-label-known-issues-hook
+++ b/openqa-label-known-issues-hook
@@ -35,7 +35,7 @@ done
 
 main() {
     host_url="$scheme://$host"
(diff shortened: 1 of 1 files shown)


openqa-label-known-issues-hook
okurz, line 15: that would need to be changed as well. I have already submitted a commit to https://github.com/os-autoinst/os-autoinst-scripts/pull/641 after your initial comment https://github.com/os-autoinst/os-autoinst-scripts/pull/641#issuecomment-5044250820

okurz (changes requested):

PR #643, opened 2026-07-22
feat: wait for hypervisor SSH before retriggering jobs

okurz:
Motivation:
Retriggered jobs were instantly failing because they were picked up
before the rebooting hypervisor's SSH server was back online.

Design Choices:
Added a wait_for_host polling loop inside s390x-qemu-zombie-reaper.py
to ensure the host is responsive before openqa-cli restart commands
are issued.

Benefits:
Prevents new jobs from crashing immediately against unreachable hosts
and outlines the future architectural fix for the s390x backend.

Related issue: https://progress.opensuse.org/issues/201144

### DIFF
--- a/s390x-qemu-zombie-reaper.py
+++ b/s390x-qemu-zombie-reaper.py
@@ -14,6 +14,7 @@ import json
 import shlex
 import subprocess
 import time
+from dataclasses import dataclass
 from enum import Enum
 from typing import TYPE_CHECKING, Annotated
 
@@ -37,6 +38,19 @@ HYPERVISORS = {
     "s390zl12.oqa.prg2.suse.org": ["worker31", "worker32"],
     "s390zl13.oqa.prg2.suse.org": ["worker32", "worker33"],
 }
+
+
+@dataclass(frozen=True)
+class ReaperConfig:
+    """Configuration for the zombie reaper execution."""
+
+    dry_run: bool = False
+    verbose: bool = False
+    reboot_method: RebootMethod = RebootMethod.SYSRQ
+    max_wait_minutes: int = 30
+    stability_delay_minutes: int = 3
+    ssh_timeout: int = 10
+    ssh_check_interval: int = 15
 
 
 def run_cmd(cmd: str, *, check: bool = True, verbose: bool = False) -> str:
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

s390x-qemu-zombie-reaper.py
```diff
@@ -100,17 +125,26 @@ def trigger_actions(
 
     if dry_run:
         print(f"[DRY-RUN] Would execute: {reboot_cmd}")
+        if jobs:
+            print(f"[DRY-RUN] Would wait for {host} to be responsive over SSH.")
+            print("[DRY-RUN] Would wait an additional 3 minutes.")
```
kalikiana: Can we avoid hard-coding the 3 minutes here, in the log of the real execution as well as in the wait_for_host function?

kalikiana (changes requested):

s390x-qemu-zombie-reaper.py
okurz: yes, you are right. As it's a longer waiting time I wanted to explicitly state the expected time. I could also write "some minutes" :) But I guess we can define constants and re-use those

s390x-qemu-zombie-reaper.py
```diff
@@ -39,6 +40,24 @@ class RebootMethod(str, Enum):
...
+
+
+@dataclass(frozen=True)
+class ReaperConfig:
+    """Configuration for the zombie reaper execution."""
+
+    dry_run: bool = False
+    verbose: bool = False
+    reboot_method: RebootMethod = RebootMethod.SYSRQ
+    max_wait_minutes: int = DEFAULT_MAX_WAIT_MINUTES
```
asmorodskyi: I think we can just define default value directly here and don't need extra variable to store it

s390x-qemu-zombie-reaper.py
```diff
@@ -39,6 +40,24 @@ class RebootMethod(str, Enum):
 }
 
 
+# Default timing behaviors
+DEFAULT_MAX_WAIT_MINUTES = 30
+DEFAULT_STABILITY_DELAY_MINUTES = 3
+SSH_TIMEOUT_SECONDS = 10
```
asmorodskyi: ssh settings are also part of ReaperConfig , IMHO

s390x-qemu-zombie-reaper.py
```diff
@@ -81,64 +100,115 @@ def get_running_jobs(hypervisor_host: str, *, verbose: bool = False) -> list[int
     return list(set(jobs_to_restart))
 
 
+def wait_for_host(
```
asmorodskyi: I think the best logic for wait for ssh is written in public cloud tests https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/instance.pm#L410 . it came from years of contant running and fine tuning to establish best stability .

ideally we should just "still" it as is , but this probably worth separate PR

s390x-qemu-zombie-reaper.py
```diff
@@ -81,64 +100,115 @@ def get_running_jobs(hypervisor_host: str, *, verbose: bool = False) -> list[int
...
+    max_wait_minutes: int = DEFAULT_MAX_WAIT_MINUTES,
+) -> bool:
+    """Wait until the host is responsive over SSH."""
+    print(f"Waiting for {host} to become responsive over SSH...")
+    start_time = time.time()
+    max_wait_seconds = max_wait_minutes * 60
+
+    while time.time() - start_time < max_wait_seconds:
+        # Use short timeout to avoid hanging indefinitely
+        check_cmd = f"ssh -o ConnectTimeout={SSH_TIMEOUT_SECONDS} -o BatchMode=yes {host} true"
```
asmorodskyi: `-o ControlPath=none -o ConnectTimeout=10 -o strictHostKeyChecking=no -o UserKnownHostsFile=/dev/null` this is what public cloud tests are using while wait for host to offer ssh connection

s390x-qemu-zombie-reaper.py
```diff
@@ -81,64 +100,115 @@ def get_running_jobs(hypervisor_host: str, *, verbose: bool = False) -> list[int
...
         action_msg = f"Triggering reboot on {host}..."
 
-    if dry_run:
+    if config.dry_run:
         print(f"[DRY-RUN] Would execute: {reboot_cmd}")
-        for job_id in jobs:
-            retrigger_cmd = f"openqa-cli api --osd -X POST jobs/{job_id}/restart"
-            print(f"[DRY-RUN] Would execute: {retrigger_cmd}")
     else:
         print(action_msg)
```
asmorodskyi: can you please move this outside dry_run switch ?

asmorodskyi (dismissed):

s390x-qemu-zombie-reaper.py
okurz: @asmorodskyi @kalikiana but here we are not waiting for openQA SUTs, we are waiting for production machines so we *can* and *should* check the host key

s390x-qemu-zombie-reaper.py
okurz: ok, done

s390x-qemu-zombie-reaper.py
okurz: yes, done. defined aligned with https://github.com/os-autoinst/os-autoinst-scripts/pull/643#discussion_r3659545317

s390x-qemu-zombie-reaper.py
okurz: done

okurz:
addressed all comments

kalikiana (approved):

s390x-qemu-zombie-reaper.py
kalikiana: for the record, we discussed it briefly and decided options can be tweaked later if needed

PR #650, opened 2026-07-28
fix(query-label): fix variable interpolation

okurz:
Motivation:
Older psql clients do not support variable interpolation when running queries
passed via the --command option, leading to syntax errors on older hosts.

Design Choices:
Pass the query via standard input (-f -) to psql, which supports native
variable interpolation across all client versions.

Benefits:
Restores compatibility with older target databases without sacrificing the use
of native psql variables.

### DIFF
--- a/openqa-query-for-job-label
+++ b/openqa-query-for-job-label
@@ -40,7 +40,7 @@ main() {
 
     [ "$dry_run" = "1" ] && _dry_run="echo"
     for h in $host; do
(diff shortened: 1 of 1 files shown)


aylen384 (changes requested):
Please use short url syntax while you are at it

Martchus (approved):

okurz:
> Please use short url syntax while you are at it

done. added a commit

aylen384 (approved):

PR #651, opened 2026-07-29
feat: extend s390x zombie reaper with libvirt health check

okurz:
Motivation:
Ensure that s390x hypervisors are rebooted when libvirt loses connection to
the qemu:///system daemon (or is otherwise unresponsive), preventing silent
reoccurring openQA job failures on the same hypervisor.

Design Choices:
Added a `check_libvirt_health` helper executing a timed-out `virsh list`
command on the host, gracefully skipping reboots if SSH connection fails.
Extended `handle_host` to trigger reboots and retrigger jobs if libvirt is
unhealthy.

Benefits:
Automates recovery of hypervisors affected by libvirtd and QEMU connection
disconnect issues, matching the existing robust zombie-process healing flow.

Related issue: https://progress.opensuse.org/issues/204384

### DIFF
--- a/s390x-qemu-zombie-reaper.py
+++ b/s390x-qemu-zombie-reaper.py
@@ -39,6 +39,7 @@ HYPERVISORS = {
     "s390zl13.oqa.prg2.suse.org": ["worker32", "worker33"],
 }
 
+RET_SUCCESS = 0
 SSH_ERR_CONNECT = 255
 
 
@@ -116,7 +117,7 @@ def wait_for_host(host: str, config: ReaperConfig) -> bool:
                 check=False,
                 timeout=config.ssh_timeout + 5,
             )
-            if res.returncode == 0:
+            if res.returncode == RET_SUCCESS:
                 print(f"Host {host} is responsive again.")
                 return True
         except (OSError, subprocess.TimeoutExpired) as e:
@@ -180,7 +181,8 @@ def check_libvirt_health(host: str, config: ReaperConfig) -> bool:
     try:
         res = subprocess.run(  # noqa: S603
             shlex.split(check_cmd),
-            capture_output=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
             text=True,
             check=False,
             timeout=15,
@@ -193,12 +195,12 @@ def check_libvirt_health(host: str, config: ReaperConfig) -> bool:
         if config.verbose:
             print(f"Host {host} is unreachable over SSH (exit code 255). Skipping libvirt health check.")
         return True
-    if res.returncode != 0:
+    if res.returncode != RET_SUCCESS:
         print(f"libvirt health check failed on {host}: virsh list returned {res.returncode}")
-        if res.stderr:
-            print(f"stderr: {res.stderr.strip()}")
+        if res.stdout:
+            print(f"output: {res.stdout.strip()}")
         return False
-    if "error:" in res.stderr or "error:" in res.stdout:
+    if "error:" in res.stdout:
         print(f"libvirt health check failed on {host}: error detected in output")
         return False
     return True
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

s390x-qemu-zombie-reaper.py
```diff
@@ -170,6 +172,38 @@ def trigger_actions(
...
+        )
+    except (OSError, subprocess.TimeoutExpired) as e:
+        print(f"libvirt health check failed on {host}: {e}")
+        return False
+
+    if res.returncode == SSH_ERR_CONNECT:
+        if config.verbose:
+            print(f"Host {host} is unreachable over SSH (exit code 255). Skipping libvirt health check.")
+        return True
+    if res.returncode != 0:
```
nicksinger: going with the theme of "no magic numbers" you should add some `RET_SUCCESS = 0` constant too

s390x-qemu-zombie-reaper.py
```diff
@@ -170,6 +172,38 @@ def trigger_actions(
...
+        print(f"libvirt health check failed on {host}: virsh list returned {res.returncode}")
+        if res.stderr:
+            print(f"stderr: {res.stderr.strip()}")
+        return False
+    if "error:" in f"{res.stdout}{res.stderr}":
```
nicksinger: How about using the feature described in https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.stdout and redirect both into the same stream (`stderr=subprocess.STDOUT`)?

s390x-qemu-zombie-reaper.py
```diff
@@ -178,42 +212,44 @@ def handle_host(
...
-    verified_identities = run_cmd(id_cmd, check=False, verbose=config.verbose)
-    stuck_identities = set(initial_identities.splitlines()).intersection(verified_identities.splitlines())
-
-    if not stuck_identities:
-        if config.verbose:
+    if zombie_pids_str:
+        # Snapshot process identities (PID + start time + state) to detect PID reuse
+        pids = ",".join(zombie_pids_str.split())
+        id_cmd = f"ssh {host} ps -o pid,lstart,state -p {pids} --no-headers"
+        initial_identities = run_cmd(id_cmd, check=False, verbose=config.verbose)
```
nicksinger: `candidates` sounds more fitting than `identities` to me, WDYT?

s390x-qemu-zombie-reaper.py
```diff
@@ -178,42 +212,44 @@ def handle_host(
...
+
+        print(f"Detected potential zombies on {host}, waiting 10s to verify persistence...")
+        time.sleep(10)
+
+        # Re-verify identities. Only processes that match exactly are confirmed as persistent zombies.
+        verified_identities = run_cmd(id_cmd, check=False, verbose=config.verbose)
+        stuck_identities = set(initial_identities.splitlines()).intersection(verified_identities.splitlines())
+
+        if stuck_identities:
+            stuck_pids = [line.split()[0] for line in stuck_identities]
```
nicksinger:
```suggestion
            stuck_pids = [line.split(maxsplit=1)[0] for line in stuck_identities]
```

Out of curiosity, I benchmarked my suggestion - it is quite a bit faster (at least if you run it often enough ;) ). Here is my (naive) test-script if you want to see for your own:

```python
import time

# ps auxf > /tmp/test.txt
with open("/tmp/test.txt", "r") as fd:
  test_string = fd.read()

iterations = 10000
i = 0

for maxsplit in [0, 1]:
  start_time = time.time()
  while i <= iterations:
    test_string.split()
    i += 1
  end_time = time.time()
  print(f"{end_time-start_time:.2f}s runtime with {maxsplit} splits")
```

```
» python bench_split.py
2.9045s runtime with 0 splits
0.0000s runtime with 1 splits
```

s390x-qemu-zombie-reaper.py
```diff
@@ -178,42 +212,44 @@ def handle_host(
...
-    else:
-        print(f"No active jobs found using {host}.")
 
-    trigger_actions(host, jobs, config)
+    if not critical_msg and not check_libvirt_health(host, config):
+        critical_msg = f"!!! CRITICAL: libvirt is unhealthy on {host}"
+
+    if critical_msg:
+        print(critical_msg)
+        print(f"Identifying jobs using {host}...")
```
nicksinger:
```suggestion
        print(f"Identifying jobs which ran on {host}...")
```

nicksinger (changes requested):
Nothing too critical, the overall logic seems fine. Still some minor suggestions you can think about :)

s390x-qemu-zombie-reaper.py
okurz: ok, done

s390x-qemu-zombie-reaper.py
okurz: done

s390x-qemu-zombie-reaper.py
okurz: candidates it is then :)

s390x-qemu-zombie-reaper.py
okurz: done

s390x-qemu-zombie-reaper.py
okurz: done

okurz:
all review suggestions incorporated

asmorodskyi:
this file breaking many rules which are hard requirement in other places where we have python code . And I can predict your answers why but nevertheless I would ask why we not applying same standards everywhere ?

okurz:
> this file breaking many rules which are hard requirement in other places where we have python code . And I can predict your answers why but nevertheless I would ask why we not applying same standards everywhere ?

Which do you mean?

asmorodskyi:
> > this file breaking many rules which are hard requirement in other places where we have python code . And I can predict your answers why but nevertheless I would ask why we not applying same standards everywhere ?
> 
> Which do you mean?

ok it was too harsh statement :) after deeper look most of concerns are gone only thing which is left is use of print instead of proper logger but it can not be called "breaking many rules which are hard requirement" so I am taking my words back . Also this certainly not directly related to this PR ...

asmorodskyi (approved):

Martchus (approved):

okurz:
@nicksinger please re-review

nicksinger (approved):

PR #652, opened 2026-08-03
fix: call _get_packages_to_submit with kwarg

okurz:
Motivation:
The function _get_packages_to_submit requires dry_run as a keyword-only
parameter, but it was being called with dry_run passed as a positional
argument, raising a TypeError.

Design Choices:
Explicitly pass dry_run as a keyword argument (dry_run=dry_run) in the
call inside _run_submissions, and add corresponding unit tests in
tests/test_auto_submit.py.

Benefits:
Fixes the TypeError during auto-submissions and ensures the parameter passing
logic is well-tested.

Related issue: https://progress.opensuse.org/issues/201843


perlpunk (approved):

okurz:
@perlpunk I added another smaller fix for a deeper task in a "dry-run" workflow and a commit to force submitting despite packages remaining in devel:openQA:testing when needed

EDIT: Actually I moved those out into a separate PR:
* #653 

so we are back to the state you previously approved already

asmorodskyi (approved):

PR #655, opened 2026-08-05
Fix Agama live boot and prevent GPU call traces on GPU nodes

Julie-CAO:
- dracut requires the root=live:... kernel parameter to tell initrd where to download and mount the squashfs root image

- nouveau results in call trace on machine with a GPU device

verification run: https://openqa.opensuse.org/tests/6144298   //passed Agama installation


asmorodskyi (approved):

okurz (changes requested):
s/ouveau/nouveau

but please rephrase the commit message title to deliver the "So What", not the "What"

Julie-CAO:
> s/ouveau/nouveau
> 
> but please rephrase the commit message title to deliver the "So What", not the "What"

Thanks. Fixed.

okurz (approved):

